PR #3, opened 2020-05-13
Add coloring support for physically placed mem regions

gschwaer:
This PR enables coloring for physically placed mem regions in the configuration by a flag (mem_region.colored). If set to true, a physically placed mem region will be colored according to the vm's colors. If set to false or omitted, it changes nothing.

Use case:
System memories like on chip memory (OCM on Xilinx ZCU102) to which the access should be cache colored.


gschwaer:
Closing because I reorganized my repo branches. Reopened in #6

PR #4, opened 2020-08-14
Allow config relative binary paths and configs placed outside of the …

gschwaer:
This PR enables the specification of a repo external config folder. Also it enables specification of a relative binary path in config.c (relative to the config folder, e.g., ./image.bin => in the specified config folder).


josecm:
I've implemented this in a different way which precludes the need for replicating the linker script in each external config folder (d890ba8).

PR #6, opened 2020-12-10
Fix: Enable coloring for physically placed mem regions

gschwaer:
In default Bao when physically placed memory regions are in a colored VM, they are not colored. This PR enables coloring for physically placed mem regions in the configuration by an additional flag (mem_region.colored). If set to true, a physically placed mem region will be colored according to the vm's colors. If set to false or omitted, the region is not colored.

It might be worth discussing if this should not be the standard behaviour.

Use case:
System memories like on chip memory (OCM on Xilinx ZCU102) to which the access should be cache colored.


josecm:
I've synced this with the dev branch and change the PR target. Also I've changed the name of the new `colored` config varaible to clarify this relates to the physical allocation. @gschwaer  If you are okay with it, I'll merge and close.

gschwaer:
Hi Jose, sure go ahead. Thank you for including this admittedly niche feature.

PR #17, opened 2021-12-12
Resolve macro TCR_T0SZ_MSK  definition errors 

maoyuze:


josecm:
Well spotted! May I ask in what context did you find the error in this macro? I would also ask you to fix the same issue with `VTCR_T0SZ_MSK` also.

PR #42, opened 2022-12-10
add gicv3 feature check before use

JianlongCao:
Why: gic driver requires system register support,
     it would be safe to check before use
How: add feature check before using
Notes: it might be able put feature read together
       incase read sysreg multiple times

Signed-off-by: Cao Jianlong <caojianlong@outlook.com>


josecm (changes requested):
The gicv3 driver is also used in aarch32 based systems, but since the sysreg you are checking does not exist for this architecture compiling for it fails (e.g. compiling with PLATFORM=fvp-r). This needs to be fixed.

josecm:
@JianlongCao thanks for the PR and sorry for the huge delay in looking at this.

The check you are adding makes sense, but I'm curious on what you motivation for it was. Are you running bao on a platform that does not support sysreg access to the GIC interface?

DavidMCerdeira:
> The gicv3 driver is also used in aarch32 based systems, but since the sysreg you are checking does not exist for this architecture compiling for it fails (e.g. compiling with PLATFORM=fvp-r). This needs to be fixed.

How should we go about doing this? 
I see two approaches:
1- Define architecture specific code e.g., `arch_gicc_cpu_has_gicv`. For aarch32 we would simply return `false`.
2- Add pre-processor macros and conditionally compile this check based on the architecture.

JianlongCao:
> @JianlongCao thanks for the PR and sorry for the huge delay in looking at this.
> 
> The check you are adding makes sense, but I'm curious on what you motivation for it was. Are you running bao on a platform that does not support sysreg access to the GIC interface?

Apologies for the delayed response. I am considering adding this feature primarily from a functional safety perspective. In similar scenarios, we have added safety mechanisms in the past, although in reality, I have not encountered such a fault scenario.

josecm:
I think it is ok to merge this. It performs a sanity check that is useful when porting to new platforms, and it is not in any critical path. Agree @DavidMCerdeira?

@JianlongCao I'd just ask you to merge the full PR in a single commit with a message following the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) structure.

JianlongCao:
> I think it is ok to merge this. It performs a sanity check that is useful when porting to new platforms, and it is not in any critical path. Agree @DavidMCerdeira?
> 
> @JianlongCao I'd just ask you to merge the full PR in a single commit with a message following the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) structure.

Done, the PRs are merged

josecm:
@JianlongCao sorry for being pedantic here, but the commit message it is still not in conventional commit form and we do plan to start following this and add CI pipeline checks. May I suggest: "feat(armv8/gicv3): add feature check before use"

DavidMCerdeira:
Looks good to me!

One note regarding your commit message, could you please write this more clearly: "it might be able put feature read together incase read sysreg multiple times". Honestly, I'm not sure I understand what you mean.

JianlongCao:
> Looks good to me!
> 
> One note regarding your commit message, could you please write this more clearly: "it might be able put feature read together incase read sysreg multiple times". Honestly, I'm not sure I understand what you mean.

Explain to the notes: To create a function in init to check all the arm features available and use a feature flag array for furthur component init. 
I have deleted the notes, maybe dont have to put it in commit msg.

DavidMCerdeira (approved):

src/arch/armv8/gicv3.c
```diff
@@ -27,6 +27,11 @@ size_t gich_num_lrs()
 
 static inline void gicc_init()
 {
+    size_t cpu_has_gicv = sysreg_id_aa64pfr0_el1_read() & ID_AA64PFR0_GIC_MSK;
+    if (!cpu_has_gicv) {
```
josecm, line 31: Just catch this. Shouldn't this variable be called `cpu_has_gic_sysreg` or something similar? Just because in the GIC spec terminology, GICV refers to the GIC virtual CPU interface...

josecm:
This is also not compiling for aarch32 platforms.

josecm:
@JianlongCao I'm closing this PR due to inactivity. Please let me know if you want to proceed and I'll re-open it.

PR #44, opened 2023-01-02
Fix the printk with error negative digits display

JianlongCao:
The current printk has an error when processing negative print, displaying a very large value. For example, print(%d, -1) is displayed as 4294967295 because the unsigned data type is used when processing this type.

Signed-off-by: JianlongCao <caojianlong@outlook.com>


josecm:
@JianlongCao sorry for the delay on this. Thanks for the fix!

PR #49, opened 2023-02-03
fix(aarch32): clear TE bit to take exceptions on A32 state

AfonsoSantos96:
This PR allows the exceptions on the EL2 to be taken on the A32 state, as the TE bit is cleared.
The reset value of the TE bit from HSCTLR is implementation defined.

Signed-off-by: Afonso Santos <afomms@gmail.com>

### DIFF
--- a/src/arch/armv8/armv8-r/aarch64/boot.S
+++ b/src/arch/armv8/armv8-r/aarch64/boot.S
@@ -129,10 +129,8 @@ skip_non_loadable:
     isb
 
     /* Enable caches and MPU */
-    mrs x3, sctlr_el2
-    ldr x4, =(SCTLR_C | SCTLR_I | SCTLR_M)
-    orr x3, x3, x4
-    msr sctlr_el2, x3
+    ldr x4, =(SCTLR_RES1 | SCTLR_C | SCTLR_I | SCTLR_M)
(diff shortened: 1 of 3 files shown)


src/arch/armv8/armv8-r/aarch64/boot.S
```diff
@@ -129,10 +129,8 @@ skip_non_loadable:
     isb
 
     /* Enable caches and MPU */
-    mrs x3, sctlr_el2
-    ldr x4, =(SCTLR_C | SCTLR_I | SCTLR_M)
-    orr x3, x3, x4
-    msr sctlr_el2, x3
+    ldr x4, =(SCTLR_RES1_AARCH32 | SCTLR_C | SCTLR_I | SCTLR_M)
```
josecm, line 132: You are using 'SCTLR_RES1_AARCH32' in aarch64 boot. Please fix with 'SCTLR_RES1'

josecm (changes requested):

src/arch/armv8/armv8-r/aarch64/boot.S
AfonsoSantos96: You're right, thanks!

josecm:
Looks good now! Thanks!

PR #51, opened 2023-02-10
SSTC extension support

ninolomata:
Features:

- add henvcfg csr;
- add support for the sstc extension;

### DIFF
--- a/src/arch/riscv/inc/arch/cpu.h
+++ b/src/arch/riscv/inc/arch/cpu.h
@@ -8,6 +8,8 @@
 
 #include <bao.h>
 
+#define CPU_HAS_EXTENSION(EXT) (DEFINED(EXT))
+
 extern cpuid_t CPU_MASTER;
 
 struct cpu_arch {
--- a/src/arch/riscv/sbi.c
+++ b/src/arch/riscv/sbi.c
@@ -230,10 +230,13 @@ struct sbiret sbi_time_handler(unsigned long fid)
     if (fid != SBI_SET_TIMER_FID) return (struct sbiret){SBI_ERR_NOT_SUPPORTED};
 
     uint64_t stime_value = vcpu_readreg(cpu()->vcpu, REG_A0);
-
-    sbi_set_timer(stime_value);  // assumes always success
-    CSRC(CSR_HVIP, HIP_VSTIP);
-    CSRS(sie, SIE_STIE);
+    if(CPU_HAS_EXTENSION(CPU_EXT_SSTC)) {
+        CSRW(CSR_VSTIMECMP, stime_value);
+    } else {
+        sbi_set_timer(stime_value);  // assumes always success
+        CSRC(CSR_HVIP, HIP_VSTIP);
+        CSRS(sie, SIE_STIE);
+    }
 
     return (struct sbiret){SBI_SUCCESS};
 }
--- a/src/arch/riscv/vm.c
+++ b/src/arch/riscv/vm.c
@@ -41,6 +41,8 @@ void vcpu_arch_reset(struct vcpu *vcpu, vaddr_t entry)
     vcpu->regs.a0 = vcpu->arch.hart_id = vcpu->id;
     vcpu->regs.a1 = 0;  // according to sbi it should be the dtb load address
 
+    CSRW(CSR_HENVCFG, 0);
+
     CSRW(CSR_HCOUNTEREN, HCOUNTEREN_TM);
     CSRW(CSR_HTIMEDELTA, 0);
     CSRW(CSR_VSSTATUS, SSTATUS_SD | SSTATUS_FS_DIRTY | SSTATUS_XS_DIRTY);
--- a/src/arch/riscv/vm.c
+++ b/src/arch/riscv/vm.c
@@ -41,8 +41,12 @@ void vcpu_arch_reset(struct vcpu *vcpu, vaddr_t entry)
     vcpu->regs.a0 = vcpu->arch.hart_id = vcpu->id;
     vcpu->regs.a1 = 0;  // according to sbi it should be the dtb load address
 
-    CSRW(CSR_HENVCFG, 0);
-
+    if(CPU_HAS_EXTENSION(CPU_EXT_SSTC)) {
+        CSRW(CSR_STIMECMP, -1);
+        CSRS(CSR_HENVCFG, HENVCFG_STCE);
+    } else {
+        CSRC(CSR_HENVCFG, HENVCFG_STCE);
+    }
     CSRW(CSR_HCOUNTEREN, HCOUNTEREN_TM);
     CSRW(CSR_HTIMEDELTA, 0);
     CSRW(CSR_VSSTATUS, SSTATUS_SD | SSTATUS_FS_DIRTY | SSTATUS_XS_DIRTY);
--- a/src/arch/riscv/inc/arch/csrs.h
+++ b/src/arch/riscv/inc/arch/csrs.h
@@ -32,6 +32,9 @@
 #define CSR_HTINST 0x64A
 #define CSR_HGATP 0x680
 #define CSR_HGEIP 0xE07
+/* Hypervisor Configuration */
+#define CSR_HENVCFG 0x60A
+#define CSR_HENVCFGH 0x61A
 
 #define STVEC_MODE_OFF (0)
 #define STVEC_MODE_LEN (2)
--- a/src/arch/riscv/inc/arch/csrs.h
+++ b/src/arch/riscv/inc/arch/csrs.h
@@ -17,6 +17,9 @@
 #define CSR_VSTVAL 0x243
 #define CSR_VSIP 0x244
 #define CSR_VSATP 0x280
+/* Sstc Extension */
+#define CSR_VSTIMECMP 0x24D
+#define CSR_VSTIMECMPH 0x25D
 
 #define CSR_HSTATUS 0x600
 #define CSR_HEDELEG 0x602
@@ -35,6 +38,10 @@
 /* Hypervisor Configuration */
 #define CSR_HENVCFG 0x60A
 #define CSR_HENVCFGH 0x61A
+
+/* Sstc Extension */
+#define CSR_STIMECMP 0x14D
+#define CSR_STIMECMPH 0x15D
 
 #define STVEC_MODE_OFF (0)
 #define STVEC_MODE_LEN (2)
@@ -189,6 +196,15 @@
 #define HSTATUS_VSXL_MSK (BIT_MASK(HSTATUS_VSXL_OFF, HSTATUS_VSXL_LEN))
 #define HSTATUS_VSXL_32 (1ULL << HSTATUS_VSXL_OFF)
 #define HSTATUS_VSXL_64 (2ULL << HSTATUS_VSXL_OFF)
+
+#define HENVCFG_FIOM (1ULL << 0)
+#define HENVCFG_CBIE_OFF (4)
+#define HSTATUS_CBIE_LEN (2)
+#define HSTATUS_CBIE_MSK (BIT_MASK(HSTATUS_CBIE_OFF, HSTATUS_CBIE_LEN))
+#define HENVCFG_CBCFE (1ULL << 6)
+#define HENVCFG_CBZE  (1ULL << 7)
+#define HENVCFG_PBMTE (1ULL << 62)
+#define HENVCFG_STCE  (1ULL << 63)
 
 #define HCOUNTEREN_CY (1ULL << 0)
 #define HCOUNTEREN_TM (1ULL << 1)
--- a/src/platform/qemu-riscv64-virt/inc/plat/platform.h
+++ b/src/platform/qemu-riscv64-virt/inc/plat/platform.h
@@ -8,4 +8,6 @@
 
 #include <drivers/sbi_uart.h>
 
+#define CPU_EXT_SSTC    1
+
 #endif


josecm:
@ninolomata are the sstc extensions available in qemu for testing? Or did you test this only on CVA6? If these are available in qemu, is it enabled by default? Or is there any option I should add? I'm thinking of something like `-cpu rv64,sstc=true`...

ninolomata:
The sstc extension is enabled by default on Qemu version 7.2. On Linux you enabled it on the device tree, just as follows: 
```

cpu@0 {
	phandle = <0x05>;
	device_type = "cpu";
	reg = <0x00>;
	status = "okay";
	compatible = "riscv";
	riscv,isa = "rv64imafdch_zicsr_zifencei_zihintpause_zba_zbb_zbc_zbs_sstc";
	mmu-type = "riscv,sv48";

	interrupt-controller {
		#interrupt-cells = <0x01>;
		interrupt-controller;
		compatible = "riscv,cpu-intc";
		phandle = <0x06>;
	};
};
```

src/arch/riscv/inc/arch/cpu.h
```diff
@@ -10,13 +10,22 @@
 
 extern cpuid_t CPU_MASTER;
 
+enum cpu_exts {
+    SSTC_EXTENSION,
+    EXTENSION_NOT_SUPPORTED
```
josecm, line 15: I'd suggest all variants start wih the same prefix. For example:

```suggestion
    CPU_EXT_SSTC,
    CPU_EXT_NUM,
```

src/arch/riscv/sbi.c
```diff
@@ -230,9 +230,12 @@ struct sbiret sbi_time_handler(unsigned long fid)
...
-
-    sbi_set_timer(stime_value);  // assumes always success
-    CSRC(CSR_HVIP, HIP_VSTIP);
+    if(cpu_arch_has_extension(SSTC_EXTENSION)) {
+        CSRW(CSR_VSTIMECMP, stime_value);
+    } else {
+        sbi_set_timer(stime_value);  // assumes always success
+        CSRC(CSR_HVIP, HIP_VSTIP);
+    }
     CSRS(sie, SIE_STIE);
```
josecm, line 239: Shouldn't this be inside the else branch above? We don't need to enable the hs-level timer interrupt if sstc is available...

src/arch/riscv/vm.c
```diff
@@ -32,7 +32,7 @@ void vcpu_arch_init(struct vcpu *vcpu, struct vm *vm) {
 void vcpu_arch_reset(struct vcpu *vcpu, vaddr_t entry)
 {
     memset(&vcpu->regs, 0, sizeof(struct arch_regs));
-    
+
```
josecm, line 35: Lookout for these white space modifications. Shouldn't appear in the commit.

src/arch/riscv/vm.c
```diff
@@ -82,6 +88,6 @@ void vcpu_arch_run(struct vcpu *vcpu){
         vcpu_arch_entry();
     } else {
         cpu_idle();
-    }
```
josecm, line 85: x2

src/arch/riscv/vm.c
```diff
@@ -1,5 +1,5 @@
 /**
- * SPDX-License-Identifier: Apache-2.0 
+ * SPDX-License-Identifier: Apache-2.0
```
josecm, line 2: This kind of modification is unecessary (even if it makes sense).

src/arch/riscv/vm.c
```diff
@@ -41,6 +41,12 @@ void vcpu_arch_reset(struct vcpu *vcpu, vaddr_t entry)
     vcpu->regs.a0 = vcpu->arch.hart_id = vcpu->id;
     vcpu->regs.a1 = 0;  // according to sbi it should be the dtb load address
 
+    if (cpu_arch_has_extension(SSTC_EXTENSION)) {
+        CSRW(CSR_STIMECMP, -1);
+        CSRW(CSR_HENVCFG, HENVCFG_STCE);
+    } else {
+        CSRW(CSR_HENVCFG, 0);
```
josecm, line 48: I know we don't have other extensions, but:

```suggestion
        CSRC(CSR_HENVCFG, HENVCFG_STCE);
```

josecm (changes requested):

src/arch/riscv/vm.c
ninolomata: Fixed

src/arch/riscv/vm.c
ninolomata: Fixed

src/arch/riscv/sbi.c
ninolomata: Fixed

src/arch/riscv/vm.c
```diff
@@ -41,6 +41,12 @@ void vcpu_arch_reset(struct vcpu *vcpu, vaddr_t entry)
     vcpu->regs.a0 = vcpu->arch.hart_id = vcpu->id;
     vcpu->regs.a1 = 0;  // according to sbi it should be the dtb load address
 
+    if (cpu_arch_has_extension(SSTC_EXTENSION)) {
+        CSRW(CSR_STIMECMP, -1);
+        CSRW(CSR_HENVCFG, HENVCFG_STCE);
```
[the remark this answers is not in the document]
ninolomata: Fixed

src/arch/riscv/cpu.c
```diff
@@ -32,3 +34,24 @@ void cpu_arch_idle()
                  "j cpu_idle_wakeup\n\r" ::"r"(&cpu()->stack[STACK_SIZE]));
     ERROR("returned from idle wake up");
 }
+
+void cpu_arch_probe_extensions() {
```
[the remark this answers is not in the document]
ninolomata: Fixed

src/arch/riscv/cpu.c
```diff
@@ -32,3 +34,24 @@ void cpu_arch_idle()
...
+    CSRC(CSR_HENVCFG, HENVCFG_STCE);
+}
+
+bool cpu_arch_has_extension(enum cpu_exts ext) {
+    unsigned long mask = (1ULL << ext) ;
+    if(cpu()->arch.cpu_features & mask)
+        return true;
+    else
+        return false;
+}
```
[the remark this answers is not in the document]
ninolomata: Fixed

src/arch/riscv/cpu.c
```diff
@@ -32,3 +34,24 @@ void cpu_arch_idle()
...
+    /* Probe SSTC Extension support */
+    CSRS(CSR_HENVCFG, HENVCFG_STCE);
+    val = CSRR(CSR_HENVCFG) & HENVCFG_STCE;
+    if (val) {
+        cpu()->arch.cpu_features |= (1ULL << SSTC_EXTENSION);
+    }
+    CSRC(CSR_HENVCFG, HENVCFG_STCE);
+}
+
+bool cpu_arch_has_extension(enum cpu_exts ext) {
```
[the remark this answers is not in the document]
ninolomata: This one I'm not sure how to fixed it, because this header is included in the core cpu.h header and so putting it static inline causes errors.

josecm:
@ninolomata I've tried this in qemu and it worked fine by default. However, when disabling sstc by passing `-cpu rv64,priv_spec=v1.12.0,sstc=false` as a qemu argument, bao crashes when trying to access stimecmp with an illegal instruction exception. It tries to access the register because previously the check for the extension is successful anyway. This seems like a bug in qemu that even when sstc is disabled, it still allows the writing of the STCE bit in henvcfg. Can you confirm this analysis is correct? If so, I 'll merge the PR and will probably send a patch to the qemu mailing list.

josecm:
Also, I'm kinda starting to question why we can't just set the STCE bit in henvcfg during initialization (ignoring it if it does not stick) and later directly check henvcfg.STCE instead of going through the cpu feature bitmap (also saving an extra memory access)

ninolomata:
> @ninolomata I've tried this in qemu and it worked fine by default. However, when disabling sstc by passing `-cpu rv64,priv_spec=v1.12.0,sstc=false` as a qemu argument, bao crashes when trying to access stimecmp with an illegal instruction exception. It tries to access the register because previously the check for the extension is successful anyway. This seems like a bug in qemu that even when sstc is disabled, it still allows the writing of the STCE bit in henvcfg. Can you confirm this analysis is correct? If so, I 'll merge the PR and will probably send a patch to the qemu mailing list.

Yes, see below a quote from the SSTC extension on this:
> Bit 63 of menvcfg (or bit 31 of menvcfgh) - named STCE (STimecmp Enable) - enables stimecmp for S-mode
when set to one, and the same bit of henvcfg enables vstimecmp for VS-mode. These STCE bits are WARL and
are hard-wired to 0 when this extension is not implemented

josecm:
> I 'll merge the PR and will probably send a patch to the qemu mailing list.

It seems the issue is fixed in qemu's latest version (v8.0.1). 

@ninolomata Nevertheless, I'd like you to address this:

> why we can't just set the STCE bit in henvcfg during initialization (ignoring it if it does not stick) and later directly check henvcfg.STCE instead of going through the cpu feature bitmap (also saving an extra memory access)

It seems redundant to keep this information in a separate variable.

ninolomata:
> why we can't just set the STCE bit in henvcfg during initialization (ignoring it if it does not stick) and later directly check henvcfg.STCE instead of going through the cpu feature bitmap (also saving an extra memory access)

My point is having a centralized mechanism with all cpu features, not only the ones you can query using the henvcfg . I understand that now it doesn't make sense since we only have the SSTC since we can just query the register and avoid memory access, but the same is not true for other extensions like Sscofpmf. My idea is to check which extensions are available at the start and then have this cpu_features variable store all the available features in the hart.

josecm:
> > why we can't just set the STCE bit in henvcfg during initialization (ignoring it if it does not stick) and later directly check henvcfg.STCE instead of going through the cpu feature bitmap (also saving an extra memory access)
> 
> My point is having a centralized mechanism with all cpu features, not only the ones you can query using the henvcfg . I understand that now it doesn't make sense since we only have the SSTC, and we can just query the register and avoid memory access, but the same is not true for other extensions like Sscofpmf. My idea is to check which extensions are available at the start and then have this cpu_features variable store all the available features in the hart.

I get your point and agree. My only issue here is that if this check is in some critical path (kind of what we have right now in 'sbi_time_handler')  and a miss in all cache levels occurs, that would significantly affect latency.

I was wondering... If we could assume all cores are homogenous in the extensions they provide, we could make this a compile-time conditional. What do you think?

@bao-project/core-team do any of you guys have an opinion here?

sandro2pinto:
I kind like the suggestion from @josecm.

Minimizing (interrupt) latency is at the core design on Bao, and the assumption (homogeneous cores) seems very reasonable - I'm not aware of any RISC-V SoC shipping cores with heterogenous extensions on the same SoC (at least so far).

josecm:
@ninolomata do you have any more feedback on Sandro's comment?

ninolomata:
I agree we can assume all cores are homogeneous and do this at compile time. My question now is how should I proceed ? Define a plain header file with all the configuration defines ? like a config.h ?

ninolomata:
I modified the code so that the SSTC extension is compile-time condicional @josecm.

DavidMCerdeira (commented):
Looks good. From my side I ask only to clean up the commit history.

src/arch/riscv/sbi.c
```diff
@@ -230,10 +230,13 @@ struct sbiret sbi_time_handler(unsigned long fid)
...
-    sbi_set_timer(stime_value);  // assumes always success
-    CSRC(CSR_HVIP, HIP_VSTIP);
-    CSRS(sie, SIE_STIE);
+    #if CPU_EXT_SSTC == 1
+        CSRW(CSR_VSTIMECMP, stime_value);
+    #else
+        sbi_set_timer(stime_value);  // assumes always success
+        CSRC(CSR_HVIP, HIP_VSTIP);
+        CSRS(sie, SIE_STIE);
+    #endif
```
josecm, line 239: I'd prefer:

```suggestion

   In some header:

   #define CPU_HAS_EXTENSION(EXT)    (DEFINED((EXT)))
   
   ----------------------------------------------------------------------------------
         
   Then:

    if (CPU_HAS_EXTENSION(CPU_EXT_SSTC) {
        CSRW(CSR_VSTIMECMP, stime_value);
    } else {
        sbi_set_timer(stime_value);  // assumes always success
        CSRC(CSR_HVIP, HIP_VSTIP);
        CSRS(sie, SIE_STIE);
    }
```

The compiler will take care of optimizing the unused branch away. In this way, we could easily pivot to your previous solution if we need the check for extensions to be more flexible (e.g. if different CPUs have a different extension set)

josecm (changes requested):

josecm (approved):

josecm:
Thanks, @ninolomata for following the suggestion. @DavidMCerdeira can you check the new modifications?

DavidMCerdeira (changes requested):
Looks good.

Please rewrite the commit history to prevent adding something in a commit that is remove immediately after. Specifically squash commits 425a533004decd33ff7b414e9764944583c09f3e and 
73280d891eaebe6065af2951fc5fe40949e8f9f1

ninolomata:
Done.

josecm (approved):

DavidMCerdeira (changes requested):
The commit history is still not as I expected.
There still is code being added on commit and being removed immediately afterwards.
The commits 3714f86e450f80850c22815c24b423a14be32948 and 73280d891eaebe6065af2951fc5fe40949e8f9f1 should be a single commit.

ninolomata:
@DavidMCerdeira done !

DavidMCerdeira (approved):
Good to go!

PR #53, opened 2023-02-17
RISC-V ACLINT Support

ninolomata:
Features:
- IPIs sent via ACLINT instead of SBI ECALLs

Notes:
Linux -> https://github.com/avpatel/linux/tree/riscv_aclint_v5
QEMU -> 7.2 with flag _-M aclint=on_

### DIFF
--- a/scripts/arch/riscv/platform_defs_gen.c
+++ b/scripts/arch/riscv/platform_defs_gen.c
@@ -1,0 +1,15 @@
+/**
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) Bao Project and Contributors. All rights reserved
+ */
+
+#include <stdio.h>
+#include <platform.h>
+
+void arch_platform_defs() {
+
+    if (platform.arch.aclint_sswi.base != 0) {
+        printf("#define ACLINT_SSWI 1\n");
+    }
+
+}
--- a/src/arch/riscv/arch.mk
+++ b/src/arch/riscv/arch.mk
@@ -18,6 +18,9 @@ irqc_arch_dir=$(cpu_arch_dir)/irqc/$(IRQC_DIR)
 src_dirs+=$(irqc_arch_dir)
 
 arch-cppflags+=-DIRQC=$(IRQC)
+ifeq ($(ACLINT_PRESENT), 1)
+arch-cppflags+=-DACLINT_PRESENT
+endif
 arch-cflags = -mcmodel=medany -march=rv64g -mstrict-align
 arch-asflags =
 arch-ldflags = 
--- a/src/arch/riscv/arch.mk
+++ b/src/arch/riscv/arch.mk
@@ -18,9 +18,6 @@ irqc_arch_dir=$(cpu_arch_dir)/irqc/$(IRQC_DIR)
 src_dirs+=$(irqc_arch_dir)
 
 arch-cppflags+=-DIRQC=$(IRQC)
-ifeq ($(ACLINT_PRESENT), 1)
-arch-cppflags+=-DACLINT_PRESENT
-endif
 arch-cflags = -mcmodel=medany -march=rv64g -mstrict-align
 arch-asflags =
 arch-ldflags = 
--- a/src/arch/riscv/inc/arch/aclint.h
+++ b/src/arch/riscv/inc/arch/aclint.h
@@ -1,0 +1,28 @@
+/**
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) Bao Project and Contributors. All rights reserved.
+ */
+
+#ifndef ACLINT_H
+#define ACLINT_H
+
+#include <bao.h>
+#include <platform.h>
+
+#define ACLINT_SSWI_MAX_HARTS   (4095)
+#define ACLINT_SSWI_SET_SETSSIP (0x1)
+
+struct aclint_sswi_hw {
+    uint32_t setssip[ACLINT_SSWI_MAX_HARTS];
+    uint32_t res;
+} __attribute__((__packed__, aligned(PAGE_SIZE)));
+
+extern volatile struct aclint_sswi_hw* aclint_sswi;
+
+void aclint_init();
+void aclint_send_ipi(cpuid_t hart);
+
+cpuid_t aclint_plat_sswi_index_to_hart_id(cpuid_t sswi_index);
+cpuid_t aclint_plat_hart_id_to_sswi_index(cpuid_t hard_id);
+
+#endif /* ACLINT_H */
--- a/src/arch/riscv/inc/arch/interrupts.h
+++ b/src/arch/riscv/inc/arch/interrupts.h
@@ -9,18 +9,20 @@
 #include <bao.h>
 #include <irqc.h>
 
-#define PLIC           (1)
-#define APLIC          (2)
+#define PLIC             (1)
+#define APLIC            (2)
+
+#define ACLINT_PRESENT() DEFINED(ACLINT_SSWI)
 
 /**
  * In riscv, the ipi (software interrupt) and timer interrupts dont actually have an ID as their
  * are treated differently from external interrupts routed by the external interrupt controller,
  * the PLIC. Will define their ids as the ids after the maximum possible in the PLIC.
  */
-#define SOFT_INT_ID    (IRQC_MAX_INTERRUPTS + 1)
-#define TIMR_INT_ID    (IRQC_MAX_INTERRUPTS + 2)
-#define MAX_INTERRUPTS (TIMR_INT_ID + 1)
+#define SOFT_INT_ID      (IRQC_MAX_INTERRUPTS + 1)
+#define TIMR_INT_ID      (IRQC_MAX_INTERRUPTS + 2)
+#define MAX_INTERRUPTS   (TIMR_INT_ID + 1)
 
-#define IPI_CPU_MSG    SOFT_INT_ID
+#define IPI_CPU_MSG      SOFT_INT_ID
 
 #endif /* __ARCH_INTERRUPTS_H__ */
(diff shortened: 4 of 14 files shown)


josecm:
I agree we should support the ACLINT to take advantage of faster IPIs in the platforms thave have one. However, I don't think virtualizing the ACLINT is useful because the cost of trap-and-emulate is larger than the direct hypercall. I'd say it would be better we don't support a virtual ACLINT, until maybe in the future the ACLINT adds support for virtualization or we find some guest that only supports IPIs through the ACLINT.

josecm (changes requested):
I agree we should support the ACLINT to take advantage of faster IPIs in the platforms thave have one. However, I don't think virtualizing the ACLINT is useful because the cost of trap-and-emulate is larger than the direct hypercall. I'd say it would be better we don't support a virtual ACLINT, until maybe in the future the ACLINT adds support for virtualization or we find some guest that only supports IPIs through the ACLINT.

src/arch/riscv/inc/arch/aclint.h
```diff
@@ -0,0 +1,28 @@
...
+#ifndef __ACLINT_H__
+#define __ACLINT_H__
+
+#include <bao.h>
+#include <platform.h>
+
+#define ACLINT_MAX_HARTS (4095)
+#define ACLINT_SET_SETSSIP (0x1)
+
+struct aclint_hw {
```
josecm, line 15: Calling this `struct aclint_sswi_hw` would be more clear.

src/arch/riscv/interrupts.c
```diff
@@ -28,6 +29,12 @@ void interrupts_arch_init()
         fence_sync();
 
         plic_init();
+
+        /* Initialize ACLINT if supported by the platform*/
+        if(platform.arch.aclint_base != 0) {
+            aclint = (void*) mem_alloc_map_dev(&cpu()->as, SEC_HYP_GLOBAL, INVALID_VA, 
+            platform.arch.aclint_base, NUM_PAGES(sizeof(struct aclint_hw)));
+        }
```
josecm, line 37: To follow the same logic of the plic, maybe this could be encapsulated in a `aclint_init` function.

src/arch/riscv/interrupts.c
```diff
@@ -43,7 +50,11 @@ void interrupts_arch_init()
 
 void interrupts_arch_ipi_send(cpuid_t target_cpu, irqid_t ipi_id)
 {
-    sbi_send_ipi(1ULL << target_cpu, 0);
+    if(platform.arch.aclint_base == 0) {
+        sbi_send_ipi(1ULL << target_cpu, 0);
+    } else {
+        aclint_send_ipi(target_cpu);
+    }
```
josecm, line 57: Could we try to make this compile-time constant? Maybe using the platform script to generate a `CLINT_PRESENT` macro based on `platform.arch.aclint_base`...

src/arch/riscv/inc/arch/aclint.h
```diff
@@ -0,0 +1,28 @@
+/**
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) Bao Project and Contributors. All rights reserved.
+ */
+
+#ifndef __ACLINT_H__
+#define __ACLINT_H__
```
josecm, line 7:
```suggestion
#ifndef ACLINT_H
#define ACLINT_H
```

Underscore started identifiers are reserved by the C standard. I know there are many headers like this in the rest of the code base, but these need to be removed.

josecm (changes requested):

src/arch/riscv/inc/arch/vm.h
```diff
@@ -44,6 +44,7 @@
 
 struct arch_vm_platform {
     paddr_t plic_base;
+    paddr_t aclint_base;
```
josecm, line 47: This is not necessary as we are not emulating the aclint for the VM.

src/arch/riscv/inc/arch/platform.h
```diff
@@ -10,6 +10,7 @@
 
 struct arch_platform {
     paddr_t plic_base;
+    paddr_t aclint_base;
```
josecm, line 13: Since each aclint module can have completely independent bases, as per the spec, this should be `aclint_sswi_base`.

src/arch/riscv/interrupts.c
josecm: Also, shouldn't we be mapping only the sswi module region?

josecm (changes requested):

josecm:
Please also, squash the two commits in a single one.

ninolomata:
@josecm I addressed all your comments in this recent update.

josecm:
Actually, since the information about the client being present or not is already present in the platform description I was thinking about adding to the platform define scripts the automatic generation of the macro. In a scripts/arch/riscv/platform_defs_gen.c

```c
...
    if (platform.arch.aclint_sswi.base != 0) {
        printf("#define ACLINT_PRESENT");
    }
...
```

I'll be happy to do this myself if you don't have the time.

ninolomata:
@josecm I did as you ask and created a platform specific script. If you want, I can create a separate commit just for that.

src/platform/qemu-riscv64-virt/virt_desc.c
```diff
@@ -26,6 +26,8 @@ struct platform platform = {
         #else 
         #error "unknown IRQC type " IRQC
         #endif
+        // Disable ACLINT by default
+        .aclint_sswi.base = 0
```
josecm, line 30: Could we add some kind of option compile time option to enable it similar to how we select the interrupt controller?

josecm:
> @josecm I did as you ask and created a platform specific script. If you want, I can create a separate commit just for that.

Thanks! Yes, a separate commit for that would be great!

src/arch/riscv/arch.mk
```diff
@@ -18,6 +18,10 @@ irqc_arch_dir=$(cpu_arch_dir)/irqc/$(IRQC_DIR)
 src_dirs+=$(irqc_arch_dir)
 
 arch-cppflags+=-DIRQC=$(IRQC)
+arch-cppflags+=-DIPIC=$(IPIC)
+ifeq ($(ACLINT_PRESENT), 1)
+arch-cppflags+=-DACLINT_PRESENT
+endif
```
josecm, line 24: Is this still required after generating the macro in the scripts?

src/arch/riscv/objects.mk
```diff
@@ -15,3 +15,6 @@ cpu-objs-y+=cpu.o
 cpu-objs-y+=cache.o
 cpu-objs-y+=iommu.o
 cpu-objs-y+=relocate.o
+ifeq ($(IPIC),IPIC_ACLINT)
+cpu-objs-y+=aclint.o
+endif
```
josecm, line 20: I understand that we have this so that we can use both configurations for qemu. However, this is not really that useful for other "real" platforms that either have an SSWI or not. I don't like the idea of "polluting" the arch makefiles with these option. Maybe IPIC should only be present on  platform.mk.

scripts/arch/riscv/platform_defs_gen.c
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright (c) Bao Project and Contributors. All rights reserved
+ */
+
+#include <stdio.h>
+#include <platform.h>
+
+void arch_platform_defs() {
+
+    if (platform.arch.aclint_sswi.base != 0) {
+        printf("#define ACLINT_PRESENT 1\n");
```
josecm, line 12: Don't really need the `1` if we are using the `DEFINED` macro.

src/arch/riscv/arch.mk
```diff
@@ -18,6 +18,10 @@ irqc_arch_dir=$(cpu_arch_dir)/irqc/$(IRQC_DIR)
 src_dirs+=$(irqc_arch_dir)
 
 arch-cppflags+=-DIRQC=$(IRQC)
+arch-cppflags+=-DIPIC=$(IPIC)
```
josecm, line 21: Should be only on qemu's makefile.

src/arch/riscv/inc/arch/interrupts.h
```diff
@@ -12,6 +12,9 @@
 #define PLIC  (1)
 #define APLIC (2)
 
+#define IPIC_SBI (1)
+#define IPIC_ACLINT (2)
```
josecm, line 16: I think the selection of IPIC should be circumscribed to the qemu platforms files (see other comment)

src/arch/riscv/interrupts.c
```diff
@@ -14,11 +14,15 @@
 #include <vm.h>
 #include <arch/csrs.h>
 #include <fences.h>
+#include <arch/aclint.h>
 
 void interrupts_arch_init()
 {
     if (cpu()->id  == CPU_MASTER) {  
         irqc_init();
+        if(DEFINED(ACLINT_PRESENT)){
```
josecm, line 23: Just from an aesthetics perspective, could we somehow rename the generated macro and create a new macro `#define ACLINT_PRESENT() DEFINED(ACLINT_BASE_ADDR)` so that:

```suggestion
        if(ACLINT_PRESENT()){
```

Something like this would be more readable.

josecm (changes requested):

josecm (approved):

josecm:
@ninolomata Can you resolve the conflicts in *src/platform/qemu-riscv64-virt/virt_desc.c* ?

ninolomata:
@josecm all good now.

josecm:
Now the code seems fine, but I think we need to add also a build check for compiling with `IPIC == IPIC_ACLINT`. I'll add this later today.

josecm (approved):

josecm:
@DavidMCerdeira @danielRep @sandro2pinto

danielRep (approved):

DavidMCerdeira (approved):

sandro2pinto (approved):
All good to me.

PR #55, opened 2023-03-14
Add ticked spinlock

ninolomata:
This commit replaces the current spinlock implementation to a ticked spinlock.

### DIFF
--- a/src/arch/armv8/aarch32/inc/arch/spinlock.h
+++ b/src/arch/armv8/aarch32/inc/arch/spinlock.h
@@ -43,7 +43,7 @@ static inline void spin_lock(spinlock_t* lock)
         "bne 1b \n\t"
         /* Wait for your turn */
         "2:\n\t"
-        "ldaex  %r1, %4\n\t"
+        "ldr    %r1, %4\n\t"
         "cmp    %r0, %r1\n\t"
         "beq   3f\n\t"
         "wfe \n\t"
--- a/src/arch/armv8/aarch32/inc/arch/spinlock.h
+++ b/src/arch/armv8/aarch32/inc/arch/spinlock.h
@@ -46,6 +46,7 @@ static inline void spin_lock(spinlock_t* lock)
         "ldaex  %r1, %4\n\t"
         "cmp    %r0, %r1\n\t"
         "beq   3f\n\t"
+        "wfe \n\t"
         "b 2b\n\t"
         "3:\n\t"
         : "=&r"(ticket), "=&r"(next), "=&r"(temp)
@@ -61,6 +62,8 @@ static inline void spin_unlock(spinlock_t* lock)
         "ldr    %r0, %1\n\t"
         "add    %r0, %r0, #1\n\t"
         "stl    %r0, %1\n\t"
+        "dsb ish\n\t"
+        "sev\n\t"
         : "=&r"(temp)
         : "Q"(lock->next));
 }
--- a/src/arch/armv8/aarch32/inc/arch/spinlock.h
+++ b/src/arch/armv8/aarch32/inc/arch/spinlock.h
@@ -8,30 +8,61 @@
 
 #include <bao.h>
 
-typedef volatile uint32_t spinlock_t;
+typedef struct {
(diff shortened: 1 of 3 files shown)


josecm:
I've taken the liberty of adding an implementation for the ticket lock for aarch64. So that we follow the same logic for all the supported architectures, @AfonsoSantos96 could you implement this for aarch32?

Still need to test this more thoroughly. @ninolomata did you test the risc-v implementation this in CVA6?

ninolomata:
No, but I can do it.  I  will also try on the Openpiton because we were having some troubles there with the spinlocks.

AfonsoSantos96:
> I've taken the liberty of adding an implementation for the ticket lock for aarch64, so that we follow the same logic for all the supported architectures. @AfonsoSantos96 could you implement this for aarch32?

Sure, I will carry out this implementation to aarch32.

josecm:
> No, but I can do it. I will also try on the Openpiton because we were having some troubles there with the spinlocks.

That'd be great!

> Sure, I will carry out this implementation to aarch32.

Thanks!

ninolomata:
Just tested on both it's working.

josecm:
> Just tested on both it's working.

Thanks! Leave it with me and @AfonsoSantos96  to wrap up the PR.

josecm:
@AfonsoSantos96 thanks for the commit. Where did you test this? Just the fvp model? Or any real hardware?

src/arch/armv8/aarch32/inc/arch/spinlock.h
```diff
@@ -8,30 +8,62 @@
...
+        "cmp  %r2, #0\n\t"
         "bne 1b \n\t"
-        : "=&r"(tmp), "+Q"(*lock)
-        : "r"(ONE) : "memory");
+        /* Wait for your turn */
+        "2:\n\t"
+        "ldaex  %r1, %4\n\t"
+        "cmp    %r0, %r1\n\t"
+        "beq   3f\n\t"
+        "b 2b\n\t"
```
DavidMCerdeira, line 50: is there a reason for not using wfe here?

src/arch/armv8/aarch64/inc/arch/spinlock.h
```diff
@@ -3,38 +3,67 @@
...
 static inline void spin_unlock(spinlock_t* lock)
 {
-    asm volatile("stlr wzr, %0\n\t" ::"Q"(*lock) : "memory");
+    uint32_t temp;
+
+    __asm__ volatile(
+        /* increment to next ticket */
+        "ldr    %w0, %1\n\t"
+        "add    %w0, %w0, 1\n\t"
+        "stlr   %w0, %1\n\t"
```
DavidMCerdeira, line 63: There should be a `sev` here no?

src/arch/riscv/inc/arch/spinlock.h
```diff
@@ -8,29 +8,48 @@
...
 
 static inline void spin_lock(spinlock_t* lock)
 {
-    spinlock_t const ONE = 1;
-    spinlock_t tmp = SPINLOCK_INITVAL;
+    uint32_t const INCR = 1;
+    uint32_t ticket;
+    uint32_t serving;
 
     asm volatile("1:\n\t"
```
DavidMCerdeira, line 30: Why have `1:` here?

DavidMCerdeira:
Although it might be obvious in this case, it would be nice to elaborate a bit on why this change is necessary/useful

src/arch/armv8/aarch32/inc/arch/spinlock.h
josecm: Added.

src/arch/armv8/aarch64/inc/arch/spinlock.h
josecm: Added

josecm (dismissed):

AfonsoSantos96 (approved):

josecm (approved):

PR #59, opened 2023-04-21
Update CPU master selection

josecm:
Most platforms start with only a single CPU which is then responsible for waking up others, but this is not the case for the newer Arm cortex-R where every CPU wakes up in unison. Because of this, Bao employed a lottery that made use of atomic load/store exclusive (arm) and load-acquire/store-conditional (riscv) instructions to select the master CPU that will subsequently coordinate system initialization. This happens right at the start of execution before MMU/MPU is enabled. This is a problem because the used atomic instructions might only work with translation enabled in Arm platforms or if the used atomic data is part of a non-reservable PMA region in RISC-V, which could result in multiple CPUs assuming they were the boot master.

To circumvent this problem, this PR first reverts the lottery selection to the assumption that only one CPU wakes first, therefore removing the need for synchronization, and then introduces a facility that alternatively allows a platform to statically define its master CPU.

### DIFF
--- a/scripts/platform_defs_gen.c
+++ b/scripts/platform_defs_gen.c
@@ -12,6 +12,9 @@ int main() {
 
     printf("#define PLAT_CPU_NUM (%ld)\n", platform.cpu_num);
     printf("#define PLAT_BASE_ADDR (0x%lx)\n", platform.regions[0].base);
+    if (platform.cpu_master_fixed) {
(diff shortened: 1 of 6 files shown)


DavidMCerdeira (approved):

danielRep (approved):
Seems good to me.

AfonsoSantos96 (approved):

ninolomata (approved):

PR #60, opened 2023-04-28
fix(hypercall/risc-v): don't exclude A0 register from args retrieval in hypercall()

filippofontana:
The current multi-arch implementation does not include RISC-V A0 register when running Bao on top of a RISC-V arch. A0 contains the ipc_id and it is not passed to the ipc_ipercall() function causing possible errors due to the right argument missing.

To fix this, the HYPCALL_ARG_REG(ARG) now starts with offset 0 (now A0 is the first register considered for RISC-V), but then the same interface function must be changed for ARM (by adding +1 to the updated offset).

To improve readability and avoid misunderstandings, the argument retrieved by A0 in RISC-V and by X1 in ARM is now called ipc_id.

### DIFF
--- a/src/arch/armv8/inc/arch/hypercall.h
+++ b/src/arch/armv8/inc/arch/hypercall.h
@@ -6,6 +6,6 @@
 #ifndef ARCH_HYPERCALL_H
 #define ARCH_HYPERCALL_H
 
-#define HYPCALL_ARG_REG(ARG)  (ARG)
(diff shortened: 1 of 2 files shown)


josecm (approved):
Thanks for the PR @filippofontana. LGTM! I believe I introduced this issue when modifying the hypercall ABI to comply with both SBI and SMCCC.

PR #63, opened 2023-05-09
Bug fix barrier: missing isb()

JorgeMVP:
Spec says: "Where an instruction results in an update to a System register, as is the case with the AT * address translation
instructions, explicit synchronization must be performed before the result is guaranteed to be visible to subsequent
direct reads of the PAR_EL1."

Signed-off-by: JorgeMVP <jorgepereira89@gmail.com>

### DIFF
--- a/src/arch/armv8/mem.c
+++ b/src/arch/armv8/mem.c
@@ -49,6 +49,7 @@ bool mem_translate(addr_space_t* as, void* va, uint64_t* pa)
     else
         asm volatile("AT S12E1W, %0" ::"r"(va));
 
+    asm volatile("isb" ::: "memory");
     par = MRS(PAR_EL1);
     MSR(PAR_EL1, par_saved);
     if (par & PAR_F) {


josecm:
It does need that barrier to guarantee the result in PAR is ready by the time we read it. I'd just ask you to, instead of directly using inline asm, replacing it with the[ ISB macro](https://github.com/bao-project/bao-hypervisor/blob/76a779129d4cafee80f5cd04f47acbf575247ede/src/arch/armv8/inc/arch/fences.h#L15) before I merge it.

danielRep:
> It does need that barrier to guarantee the result in PAR is ready by the time we read it. I'd just ask you to, instead of directly using inline asm, replacing it with the[ ISB macro](https://github.com/bao-project/bao-hypervisor/blob/76a779129d4cafee80f5cd04f47acbf575247ede/src/arch/armv8/inc/arch/fences.h#L15) before I merge it.

Guys this is already solved on the latest version of bao. You can check it here https://github.com/bao-project/bao-hypervisor/blob/76a779129d4cafee80f5cd04f47acbf575247ede/src/arch/armv8/armv8-a/mem.c#L43
I'm closing the PR.

PR #65, opened 2023-06-05
fix(arch/armv8): fixed typo in 'mpidr' variable and function

clachan:


josecm (approved):
@clachan Nice catch! Thanks for the PR!

PR #68, opened 2023-06-28
feat(armv8/boot): linearize cpu id according to number of clusters

AfonsoSantos96:
With this PR, during the generation of the cpu_id for each core, the boot will be aware of the number of clusters on the platform, as well as the number of cores in the cluster.

This information about the _number of clusters_ and _cores per cluster_ must be explicit on the _platform_ struct from the _\<platform\>\_desc.c_ file. This must be defined on the _arch_ field, as exemplified next.

Defining 2 cluster with 4 cores each:
```c
    .arch = {
        .clusters =  {
            .num = 2,
            .core_num = (size_t[]) {4, 4}
        },
```


josecm:
Can you update the commit message to highlight this is just for the aarch32 implementation and that it follows essentially the same procedure as the already existing aarch64 CPU linearization?

josecm (changes requested):
Please uniformize the assembly identation.

AfonsoSantos96:
> Please uniformize the assembly identation.

On my IDE it is already with the correct identation...guess it is something with github...

josecm (approved):

src/arch/armv8/aarch32/boot.S
```diff
@@ -52,11 +52,31 @@ _reset_handler:
...
+    ldr r4, [r4, r5]
+    ldr r5, =BAO_VAS_BASE
+    sub r4, r4, r5
+    add r4, r4, r1
+    mov r5, #0
+    mov r7, #0
+1:
+    cmp r5, r3
+    bge	2f
+    ldrb r6, [r4]
```
DavidMCerdeira, line 72: Reading a byte but the array element is size_t not a uint8_t

DavidMCerdeira (changes requested):

src/arch/armv8/aarch32/boot.S
josecm: I've added this correct in #73. So I'd suggest only merging this after that PR is merged and @AfonsoSantos96 uses the macro in the aarch32 code also.

josecm: #73 was merged. @AfonsoSantos96 could you re-write this PR following that fix?

josecm (changes requested):

src/arch/armv8/aarch32/boot.S
```diff
@@ -52,11 +52,31 @@ _reset_handler:
...
+    add r4, r4, #1
+    add r5, r5, #1
+    add r7, r7, r6
+    b 	1b
+2:
+    and r0, r0, #0xff
+    add r0, r0, r7
```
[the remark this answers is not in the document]
AfonsoSantos96: Nicely spotted! Followed your suggestion, thanks!

src/arch/armv8/aarch32/boot.S
AfonsoSantos96: Just fixed with the changes from #73

src/arch/armv8/aarch32/boot.S
```diff
@@ -52,12 +52,41 @@ _reset_handler:
...
+     * Linearize cpu id according to the number of clusters and processors
+     * per cluster. We are only considering two levels of affinity.
+     * TODO: this should be done some other way. We shouldn't depend on the platform
+	 * description this early in the initialization.
+     */
+    mov r3, r0, lsr #8
+    and r3, r3, #0xff
+    ldr r4, =platform
+    mov r5, #(PLAT_ARCH_OFF+PLAT_ARCH_CLUSTERS_OFF+PLAT_CLUSTERS_CORES_NUM_OFF)
+    ldr r4, [r4, r5]
```
josecm, line 69: At this point r4 contains a virtual address, you need it to first convert it to a virtual address or use a relative address instruction such as `adr` in line 67.

src/arch/armv8/aarch32/boot.S
```diff
@@ -52,12 +52,41 @@ _reset_handler:
...
+    sub r4, r4, r5
+    add r4, r4, r1
+    mov r5, #0
+    mov r7, #0
+1:
+    cmp r5, r3
+    bge	2f
+#if PLAT_CLUSTERS_CORES_NUM_SIZE != 8
+#error "struct platform cluster core_num field type size is not 8"
+#endif
```
josecm, line 80: Did you compile this? In aarch32 this assert should fail since the size of `size_t` is 4 bytes.

josecm (changes requested):

src/arch/armv8/aarch32/boot.S
AfonsoSantos96: You are right! Fixed using the macro `get_phys_addr` defined at the beginning of the file.

src/arch/armv8/aarch32/boot.S
AfonsoSantos96: My bad!... It indeed is 4 and not 8...Thanks!

josecm (dismissed):

DavidMCerdeira (dismissed):
There are some source code comment misalignment issues, but this should be fixed when we use the automatic formatter

AfonsoSantos96:
> There are some source code comment misalignment issues, but this should be fixed when we use the automatic formatter

Or I can correct it now

DavidMCerdeira (approved):

PR #69, opened 2023-06-28
RISC-V/PLIC: set proper nomenclature for threshold offset (plus typos fixing)

filippofontana:
Hi, I made this PR just to "correct" the name of some variables/macros which does not respect the nomenclature provided by the [PLIC specs](https://github.com/riscv/riscv-plic-spec/blob/master/riscv-plic.adoc#2-memory-map). 
As you can see in RISC-V PLIC specifications, offset `0x200000` is used to access the `THRESHOLD` for `context 0` instead of `CLAIM/COMPLETE` (`0x200004`) for `context 0`, so the related code has been renamed accordingly.
At the same time, I've also fixed some typos in "threshold related" functions.

Hope you will find this useful.


josecm (changes requested):
I get your point. In this setting, 'threshold' is more accurate than 'claim/complete'. But now that you mention it, a better alternative would be 'context'.  Do you agree?

ninolomata:
I was about to suggest the same, it seems more accurate to me to call it _context_.

josecm:
> So you're saying to change `PLIC_CLAIMCMPLT_OFF` to a sort of `PLIC_CNTXT_OFF` and `plic_claimcomplte_emul` to `plic_cntxt_emul`?

Exactly. 

> It can be a good solution, the only thing I'm a bit concerned about is that `context` is used to distinguish the execution contexts in also other fields (like the enabled interrupt sources).

But I get your point here. They don't specify a name for that "interrupt handling" region as they do for the APLIC (i.e. IDC). So I feel there is no perfect solution for this, so I'd be ok to leave your PR as is.

josecm (approved):

ninolomata (approved):

josecm:
Thanks again @filippofontana !

filippofontana:
Thank you guys for accepting this solution. I hope that this will be clear also for others!

PR #71, opened 2023-07-04
fix(riscv/vm): fix vplic_init call 

D3boker1:
Fix vplic_init call in riscv/vm.c to initialize the vplic with config's plic base instead platform's.

I have tested it in QEMU running Linux to ensure that the modifications perform as intended. To guarantee this, I configured the Linux VM to access the PLIC at address 0xd000000 instead of the physical address 0xc000000. With these modifications, it worked successfully.


josecm (approved):
Nice catch! Thanks!

ninolomata (approved):

PR #73, opened 2023-07-07
fix(armv8/aarch64): assert 'cluster_num' type size in asm code 

josecm:
In armv8, the platform's cluster information is used in the boot assembly code to linearize the CPU ID from Arm's hierarchical "affinity ID". However,  the code assumed byte-width entries in the `cluster_num` array, which were, in fact, of `size_t` type, which is a machine-word-sized type. In the past, the type was indeed `uint8_t` but was changed recently.

This PR fixes this issue and generates a macro with the type size used by a pre-processor assert to prevent this issue from resurfacing if the type is changed again in the future.

### DIFF
--- a/src/arch/armv8/aarch64/boot.S
+++ b/src/arch/armv8/aarch64/boot.S
@@ -76,8 +76,11 @@ _reset_handler:
 1:
 	cmp x5, x3
 	b.ge	2f
-	ldrb w6, [x4]
-	add x4, x4, #1
+#if PLAT_CLUSTERS_CORES_NUM_SIZE != 8
+#error "struct platform cluster core_num field type size is not 8"
+#endif
+	ldr w6, [x4]
+	add x4, x4, #PLAT_CLUSTERS_CORES_NUM_SIZE
 	add x5, x5, #1
 	add x7, x7, x6
 	b 	1b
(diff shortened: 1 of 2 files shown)


DavidMCerdeira (approved):

src/arch/armv8/aarch64/boot.S
```diff
@@ -76,8 +76,11 @@ _reset_handler:
 1:
 	cmp x5, x3
 	b.ge	2f
-	ldrb w6, [x4]
-	add x4, x4, #1
+#if PLAT_CLUSTERS_CORES_NUM_SIZE != 8
+#error "struct platform cluster core_num field type size is not 8"
+#endif
```
danielRep, line 81: I would prefer to define the value 8 as a macro that represents the platform word size.

danielRep (changes requested):

src/arch/armv8/aarch64/boot.S
DavidMCerdeira: @josecm thoughts?

danielRep: Maybe I'm being to pedantic, but my issue was to give more clarity to the meaning of the value. But it's fine by me, the warning message is probably enough.

danielRep (approved):

josecm:
@DavidMCerdeira, I'd say this is ready to be merged.

PR #75, opened 2023-07-13
fix(core/interrupts): fix intp reserve condition

D3boker1:
This commit fixes the interrupts_reserve function. The current condition allows the hypervisor to reserve an interrupt by only checking if it is in the correct range. By introducing the interrupt_is_reserved() function and incorporating it into the conditional statement, we avoid reassigning an interrupt ID that has already been reserved.

### DIFF
--- a/src/core/interrupts.c
+++ b/src/core/interrupts.c
@@ -46,9 +46,22 @@ inline void interrupts_init()
     interrupts_cpu_enable(IPI_CPU_MSG, true);
 }
 
-static inline bool interrupt_is_reserved(irqid_t int_id)
+static inline bool interrupt_assigned_to_hyp(irqid_t int_id)
 {
     return bitmap_get(hyp_interrupt_bitmap, int_id);
+}
+
+/**
+ * @brief For a given interrupt intp_id, return if this interrupt
+ *        is already reserved by VMM or any VM
+ * 
+ * @param int_id interrupt ID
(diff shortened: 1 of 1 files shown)


josecm:
Thanks, s this check should have been there already! I think we could take this a step forward. Add another function that checks if the interrupt is in  `global_interrupt_bitmap`  because the interrupt might be reserved for a guest also. Maybe renaming the `interrupt_is_reserved` function is in order to something like `interrupt_assigned_to_hyp` :thinking:

josecm:
@D3boker1 could you address this? If you don't have any idle cycles to spare, I'd be happy to do it.

DavidMCerdeira (changes requested):
It's better to name `interrupt_globally_assigned` as `interrupt_assigned`. Interrupts are not assigned globally, they are marked as assigned in a global bitmap.

Otherwise looks good.

D3boker1:
Hi @DavidMCerdeira
Please find the suggested change in commit 73730d0ee2f921369fd4f7e16fe1d2ba282ee965

josecm (approved):

DavidMCerdeira (changes requested):
Looks good.
Please clean the git history to avoid having the commit with the function name change.

Either squash de commits, or keep the two other commits with the correct name from the start

DavidMCerdeira (approved):

PR #76, opened 2023-07-17
fix(core/ipc): fix ipc interrupt delivery

clachan:
Add interrupt_offset to struct ipc for offsetting the IRQ number before IPC interrupt injection into guest.

Take Linux guest as an example, GIC driver (driver/irqchip/irq-gic-v3.c) adds 32 to the SPI's IRQ number. Thus, to successfully deliver the IPC interrupt to the guest, an offset is added to the IRQ number.

For example, below code snippet defines the ipc for Linux VM and assigns 52 as the IRQ number. However, the actual IRQ number in Linux is 84. By offsetting 32 to the IRQ number, the IPC interrupt can be delivered.

                .ipc_num = 1,
                .ipcs = (struct ipc[]) {
                    {
                        .base = 0xf0000000,
                        .size = 0x00010000,
                        .shmem_id = 0,
                        .interrupt_num = 1,
                        .interrupts = (irqid_t[]) {52},
                        .interrupt_offset = 32
                    }
                },

On FreeRTOS or Zephyr, the offset is not needed.

### DIFF
--- a/src/core/inc/ipc.h
+++ b/src/core/inc/ipc.h
@@ -15,6 +15,7 @@ struct ipc {
     size_t shmem_id;
     size_t interrupt_num;
     irqid_t *interrupts;
+    size_t interrupt_offset;
 };
 
 struct vm_config;
--- a/src/core/ipc.c
+++ b/src/core/ipc.c
@@ -49,7 +49,7 @@ static void ipc_notify(size_t shmem_id, size_t event_id) {
     struct ipc* ipc_obj = ipc_find_by_shmemid(cpu()->vcpu->vm, shmem_id);
     if(ipc_obj != NULL && event_id < ipc_obj->interrupt_num) {
         irqid_t irq_id = ipc_obj->interrupts[event_id];
-        vcpu_inject_hw_irq(cpu()->vcpu, irq_id);
+        vcpu_inject_hw_irq(cpu()->vcpu, irq_id + ipc_obj->interrupt_offset);
     }
 }
 


josecm:
I don't think this is really an issue. We are aware that on Linux [GIC device tree bindings](https://github.com/torvalds/linux/blob/e599e16c16a16be9907fb00608212df56d08d57b/Documentation/devicetree/bindings/interrupt-controller/arm%2Cgic.yaml#L71) SPI IDs are offseted by 32, thus IDs are in the range [0-987].  However,  the IDs you identify in Bao's config are the "true" ID in the full GIC interrupt ID range, i.e., [0-1023].  Therefore, this must be taken into account when writing the Linux guest device tree.

For example, in bao-demos for [zcu104  config](https://github.com/bao-project/bao-demos/blob/e6f1f6f5556c8e4787516404c3c7eaa5d2feb774/demos/linux%2Bfreertos/configs/zcu104.c#L58) the interrupt for the uart peripheral has ID 54, the "true" interrupt ID. But in the [guest's dts](https://github.com/bao-project/bao-demos/blob/e6f1f6f5556c8e4787516404c3c7eaa5d2feb774/demos/linux%2Bfreertos/devicetrees/zcu104/linux.dts#L84), we define it as (54 - 32 =) 22, or 0x16. 

In conclusion, this offset must be subtracted on the device tree itself. Does this make sense to you @clachan ?

clachan:
Hi @josecm,

Thanks for the feedback. It makes sense to me.

Now I realized that the irq number in BAO IPC doesn't need to be the same as that in the device tree. Deducting the number in the device tree by 32, I am able to receive the interrupt in the driver I submitted to bao-demos.

This PR can be closed.

Thanks,
Clay

PR #77, opened 2023-07-20
feat(clang): add clang support

ninolomata:


josecm:
Just a small note, could you add some information to the PR description on where you've tested this?

Also, for arm targets, we should test this with armclang, not just clang.

danielRep:
For Ubuntu 20.04, I've used 

```
16:43 $ ~/toolchains/arm/clang+llvm-16.0.0-x86_64-linux-gnu-ubuntu-18.04/bin/clang-cpp --version
clang version 16.0.0
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /home/daniel/toolchains/arm/clang+llvm-16.0.0-x86_64-linux-gnu-ubuntu-18.04/bin
```
and I've effectively run bao on `qemu-aarch64-virt` platform:
```
=> go 0x50000000
## Starting application at 0x50000000 ...
Bao Hypervisor w/ clang compilation
Bao bare-metal test guest
cpu 1 up
cpu 2 up
cpu 3 up
cpu 0 up
cpu0: timer_handler
cpu1: ipi_handler
cpu2: ipi_handler
cpu3: ipi_handler
QEMU: Terminated
```

danielRep:
@ninolomata , could you separate the changes into x number of meaningful commits? It would be easier to understand some of the changes made. I feel that e.g. with the bao code changes it was an iterative process, maybe that would be sufficient to map it to significant commits.

josecm:
Also, @ninolomata , could you clarify exactly what is need to compile the code with clang? I'm guessing:

```
make PLATFORM=... CONFIG=... LLVM=/path/to/llvm/install
```

Is my assumption correct?

I've installed clang via a package manager. So all the executables are in */usr/bin*. Could you clarify how you have installed clang also?

josecm:
Another point is that we should include  build checks for clang now that we have a CI pipeline.

DavidMCerdeira:
I've rebase this branch with main, and replaced the proposed flag with `CC_IS_*` to distinguish between GCC and CLANG compilers both on the Makefile and in the source code (e.g., on src/arch/armv8/aarch64/boot.S).

I also propose to use the LLVM variable as a full path, so if one wants to use the package manager (to install clang llvm lld)  it is possible to simply pass `LLVM=/usr/bin/` as a Makefile argument.

Another change is replacing the use of `CROSS_COMPILE` variable to create the target architecture variable for clang, with `ARCH_TARGET`. I feel that `CROSS_COMPILE` variable having different meaning depending on the compiler used can be confusing.

josecm:
@miguelafsilva5 I don't think those commit tags make sense

Diogo21Costa:
Tested on Ubuntu 22.04 with the following setup:

```sh
Ubuntu clang version 18.1.8 (++20240731024944+3b5b5c1ec4a3-1~exp1~20240731145000.144)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

Successfully ran the Bao bare-metal demo from the [bao demos](https://github.com/bao-project/bao-demos) on `qemu-aarch64-virt` platform

Diogo21Costa (dismissed):

danielRep:
> Tested on Ubuntu 22.04 with the following setup:
> 
> ```shell
> Ubuntu clang version 18.1.8 (++20240731024944+3b5b5c1ec4a3-1~exp1~20240731145000.144)
> Target: x86_64-pc-linux-gnu
> Thread model: posix
> InstalledDir: /usr/bin
> ```
> 
> Successfully ran the Bao bare-metal demo from the [bao demos](https://github.com/bao-project/bao-demos) on `qemu-aarch64-virt` platform

Double-checked. LGTM.

danielRep (approved):

DavidMCerdeira (approved):

danielRep (approved):

josecm (approved):

AfonsoSantos96 (approved):

Issue #78, opened 2023-07-24
Bitmap library throws error when compiled with optimizations level 2


ninolomata:
There is an error when compiling Bao for RISC-V in `sbi_ipi_handler` function for gcc version 12.1.0

```
struct sbiret sbi_ipi_handler(unsigned long fid)
{
    if (fid != SBI_SEND_IPI_FID) return (struct sbiret){SBI_ERR_NOT_SUPPORTED};

    unsigned long hart_mask = vcpu_readreg(cpu()->vcpu, REG_A0);
    unsigned long hart_mask_base = vcpu_readreg(cpu()->vcpu, REG_A1);

    struct cpu_msg msg = {
        .handler = SBI_MSG_ID,
        .event = SEND_IPI,
    };

    for (size_t i = 0; i < sizeof(hart_mask) * 8; i++) {
        if (bitmap_get((bitmap_t*)&hart_mask, i)) {
            vcpuid_t vhart_id = hart_mask_base + i;
            cpuid_t phart_id = vm_translate_to_pcpuid(cpu()->vcpu->vm, vhart_id);
            if(phart_id != INVALID_CPUID) cpu_send_msg(phart_id, &msg);
        }
    }

    return (struct sbiret){SBI_SUCCESS};
}
```
The following error occurs:
```
In function 'bitmap_get',
    inlined from 'sbi_ipi_handler.part.0' at /media/ninolomata/Nino/bao-demos/wrkdir/srcs/bao/src/arch/riscv/sbi.c:260:13:
/media/ninolomata/Nino/bao-demos/wrkdir/srcs/bao/src/lib/inc/bitmap.h:43:16: error: 'hart_mask' is used uninitialized [-Werror=uninitialized]
   43 |     return (map[bit / BITMAP_GRANULE_LEN] & (ONE << (bit % BITMAP_GRANULE_LEN)))
      |             ~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
/media/ninolomata/Nino/bao-demos/wrkdir/srcs/bao/src/arch/riscv/sbi.c: In function 'sbi_ipi_handler.part.0':
/media/ninolomata/Nino/bao-demos/wrkdir/srcs/bao/src/arch/riscv/sbi.c:251:19: note: 'hart_mask' declared here
  251 |     unsigned long hart_mask = vcpu_readreg(cpu()->vcpu, REG_A0);
      |                   ^~~~~~~~~
cc1: all warnings being treated as errors
```
[...]

PR #79, opened 2023-07-25
Fix forced integer to bitmap casts

josecm:
In order to use some of the operations defined for the `bitmap_t*` type, some functions were casting integer types pointers to `bitmap_t*`.  This is not the best practice and was causing some issues with the latest GCC versions, as pointed out in #78.

This PR fixes this issue by refactoring these operations using simple bit manipulation functions available in *bit/util/bit.h*. However, it was still needed to add the `bit_ffs` and `bit_count` functions so that all functionality needed was available.


ninolomata (approved):

ninolomata:
Just tested on RISC-V in QEMU issue #78 is fixed.

josecm:
@DavidMCerdeira

DavidMCerdeira (approved):

PR #93, opened 2023-09-30
feat(types): unsigned long width sanity check

josecm:


sandro2pinto (approved):
Good to me

src/core/inc/types.h
```diff
@@ -11,6 +11,16 @@
 #include <stddef.h>
 #include <limits.h>
 
+/**
+ * We assume LP64 and ILP32 for 64- and 32-bit architectrures, respectively, as
+ * throughout the code `unsigned long` is the type used for values of the
+ * architecture's word width. This is just a sanity check to verify this is the
+ * ABI the compiler is effectily using.
+ */
```
danielRep, line 19:
```suggestion
/**
 * We assume LP64 and ILP32 for 64- and 32-bit architectures, respectively, as
 * throughout the code `unsigned long` is the type used for values of the
 * architecture's word width. This is just a sanity check to verify this is the
 * ABI the compiler is effectively using.
 */
```

danielRep (changes requested):

src/core/inc/types.h
josecm: Typos fixed!

danielRep (approved):

DavidMCerdeira (approved):

PR #94, opened 2023-09-30
fix(arm/vgic): use word width for generic access

josecm:
We were always using 64-bit types for a generic vgic access, even for aarch32 which is unecessary. First because of the inneficiency associated with using 64-bit values in a 32-bit architecture, second because we don't support traps for the 64-bit access instructions available in aarch32 (this is because most hardware does not provide a valid ESR on traps due to these instructions).


sandro2pinto (approved):
All good to me.

DavidMCerdeira (approved):

PR #95, opened 2023-10-01
feat: add version symbol to final object file

josecm:
This is meant to provide a way to, by inspecting the executable object file, discover Bao's version. I'm wondering if you should actually embed it as a string in the binary so that this can also be discovered just by inspecting the image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,10 @@ OPTIMIZATIONS:=2
 CONFIG=
 PLATFORM=
 
+# Setup version
+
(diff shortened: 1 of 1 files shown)


sandro2pinto:
Yes, I would embed it as a string in the binary

danielRep:
> Yes, I would embed it as a string in the binary

I also agree. Maybe we can introduce that in another PR. I'm accepting this as it is.

danielRep (approved):

sandro2pinto (approved):
All good to me.

PR #98, opened 2023-10-08
fix(riscv/sbi): rfence hart mask adjustment and check

josecm:
Right now, the sbi rfence implementation only supports 64 harts, so the hart mask has to be rectified for always base 0 and return error if an with hart id >= 64 is being targeted. Both this check and the hart adjusment were being incorrectly performed.

This issue was detected by @D3boker1.


D3boker1 (approved):
Hi @josecm!

Everything looks good to me!

danielRep:
ping @ninolomata

ninolomata (approved):

danielRep (approved):

PR #101, opened 2023-10-13
New improved printk implementation

josecm:
The quality of the previous printk implementation was less than desirable. This PR introduces a new improved, and more robust printk implementation that allows to print strings larger than the allocated print buffers and also adds the "format printf" attribute so that incorrect type identifiers in the format string are detected at compile-time.

### DIFF
--- a/src/arch/armv8/aarch32/aborts.c
+++ b/src/arch/armv8/aarch32/aborts.c
@@ -10,10 +10,10 @@
 void internal_abort_handler(unsigned long gprs[]) {
 
     for(ssize_t i = 14; i >= 0; i--) {
-        printk("x%d:\t\t0x%0lx\n", i, gprs[14 - i]);
+        console_printk("x%d:\t\t0x%0lx\n", i, gprs[14 - i]);
     }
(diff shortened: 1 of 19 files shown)


DavidMCerdeira (approved):

DavidMCerdeira:
@josecm at least one CI check is failing

danielRep (approved):

josecm:
@DavidMCerdeira 

> @josecm at least one CI check is failing

The error was due to an "undefined reference to `__aeabi_uldivmod" error in 32-bit arm architectures since we don't link the GCC lib. I don't fully understand why this intrinsic was being invoked for one of the divisions but not for the others. Anyway, I believe I fixed this by removing the use of `long long` since the print functions don't support these types in the first place.

danielRep (approved):
I have tested `console_printk` locally with some unit tests and it seems to been working properly.

josecm:
> I have tested `console_printk` locally with some unit tests and it seems to been working properly.

Thanks!

ninolomata (approved):

AfonsoSantos96 (approved):

PR #103, opened 2023-10-20
Format code using clang format and introduce CI checks for it

josecm:
This PR is initially marked as a draft to achieve consensus on the formatting rules defined in the bao-project/bao-ci _.clang_format_ file. I'd ask the @bao-project/core-team to scan the code base and to provide feedback in this sense if they don't agree with any specific previously discussed format decisions so we can discuss further and update the clang-format options accordingly.

Temporarily the ci branch is set to[ bao-project/bao-ci's exp/clang-format-options branch ](https://github.com/bao-project/bao-ci/tree/exp/clang-format-options)so that we can play around with the CI options before merging them into bao-project/bao-ci 's main accordingly and set the submodule in this PR to it.

Depends on bao-project/bao-ci#70.


src/platform/drivers/pl011_uart/inc/drivers/pl011_uart.h
```diff
@@ -1,210 +1,213 @@
...
+
+#define NUM_UART         6
+
+#define UART_CLK         19200000
+#define UART_BAUD_RATE   115200
+
+/* UART Data Register */
+
+#define UART_DATA_DATA   0xFFFFFF00
+#define UART_DATA_FE             (1 << 8
```
danielRep, line 37: missing ')' and not aligned with below macros

src/platform/drivers/pl011_uart/inc/drivers/pl011_uart.h
```diff
@@ -1,210 +1,213 @@
...
+    volatile uint32_t line_control; // UART Line Control Register
+    volatile uint32_t control; // UART Control Register
+    volatile uint32_t isr_fifo_level_sel; // UART Interrupt FIFO level Select
+                                          // Register
+    volatile uint32_t isr_mask; // UART Interrupt Mask Set/Clear Register
+    volatile uint32_t raw_isr_status; // UART Raw Interrupt Status Register
+    volatile uint32_t masked_isr_status; // UART Masked Interrupt Status
+                                         // Register
+    volatile uint32_t isr_clear; // UART Interrupt Clear Register
+    volatile uint32_t DMA_control; // UART DMA control Register
```
danielRep, line 198: I would prefer this trailing comments to be aligned. Is this the AlignTrailingComments option?

danielRep (commented):
In the overall, I agree with clang format options that we are using. I tried to keep my eyes wide open to check for inconsistencies on the formatted files, but for sure we need to run this in all boards.

Regarding the clang options:
-  SpaceBeforeParensOptions: My a personal taste, I wouldn't use the space before control statements if we are not using it for functions
- AlignTrailingComments: this options also controls the comments in structs right? If so, I would force it. In large structs keeping the comments unaligned does not help the readability of each member of the struct.

src/platform/drivers/pl011_uart/inc/drivers/pl011_uart.h
josecm: Fixed.

src/arch/armv8/armv8-a/inc/arch/page_table.h
```diff
@@ -10,125 +10,131 @@
...
-     PTE_AF)
+        PTE_AF)
 
 #define PTE_VM_DEV_FLAGS \
     (PTE_MEMATTR_DEV_GRE | PTE_SH_NS | PTE_S2AP_RW | PTE_AF)
 
 #ifndef __ASSEMBLER__
 
-typedef uint64_t pte_t;
+    typedef uint64_t pte_t;
```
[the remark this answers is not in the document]
josecm: I can't figure out what might be triggering this!

src/platform/drivers/pl011_uart/inc/drivers/pl011_uart.h
josecm: I've added this option. Please check if it formatted this as you expected.

josecm:
I'm thinking we could define a max of 100 columns instead of 80. This is kind of being stuck in the past... Modern monitors are wide enough that this does not become an hassle.

danielRep:
> I'm thinking we could define a max of 100 columns instead of 80. This is kind of being stuck in the past... Modern monitors are wide enough that this does not become an hassle.

I totally agree with this. Other projects already follow the 100 columns too.

src/arch/armv8/aborts.c
```diff
@@ -133,15 +139,17 @@ void sysreg_handler(unsigned long iss, unsigned long far, unsigned long il, unsi
...
         struct emul_access emul;
         emul.addr = reg_addr;
         emul.width = 8;
         emul.write = iss & ESR_ISS_SYSREG_DIR ? false : true;
-        emul.reg = bit64_extract(iss, ESR_ISS_SYSREG_REG_OFF, ESR_ISS_SYSREG_REG_LEN);
-        emul.reg_high = bit64_extract(iss, ESR_ISS_SYSREG_REG2_OFF, ESR_ISS_SYSREG_REG2_LEN);
+        emul.reg =
+            bit64_extract(iss, ESR_ISS_SYSREG_REG_OFF, ESR_ISS_SYSREG_REG_LEN);
+        emul.reg_high = bit64_extract(iss, ESR_ISS_SYSREG_REG2_OFF,
+            ESR_ISS_SYSREG_REG2_LEN);
```
DavidMCerdeira, line 150: In cases like this personally I'd prefer to have AlignAfterOpenBracket=BAS_Align

DavidMCerdeira (dismissed):

src/arch/armv8/armv8-a/inc/arch/page_table.h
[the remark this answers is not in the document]
danielRep: Harmless, so lets resolve it.

danielRep (dismissed):

src/arch/armv8/aborts.c
josecm: My opinion is that this becomes a bit weird in some corner cases, as not aligning is pretty much linear. See an example of what I'm talking about below:

```c
void as_init(struct addr_space* as, enum AS_TYPE type, asid_t id, pte_t* root_pt, colormap_t colors)
{
    as->type = type;
    as->pt.dscr = type == AS_HYP || type == AS_HYP_CPY ? hyp_pt_dscr : vm_pt_dscr;
    as->colors = colors;
    as->lock = SPINLOCK_INITVAL;
    as->id = id;

    if (root_pt == NULL) {
        size_t n = NUM_PAGES(pt_size(&as->pt, 0));
        root_pt = (pte_t*)mem_alloc_page(n,
                                         type == AS_HYP || type == AS_HYP_CPY ? SEC_HYP_PRIVATE :
                                                                                SEC_HYP_VM,
                                         true);
        memset((void*)root_pt, 0, n * PAGE_SIZE);
    }
    as->pt.root = root_pt;

    as_arch_init(as);
}
```

josecm: Code could be written to avoid these cases, but nevertheless..

josecm:
Updated to use a 100 char column limit.

src/arch/armv8/aborts.c
danielRep: IMO, i would prefer to defer that option since it leads to strange formats like @josecm as showcase..

DavidMCerdeira: IMO this case in particular would be better written by separating the type identification from the function invocation into different lines and I don't see it as reason to not use BAS_Align. IMO formatting should also discourage writing code that is harder to read.

josecm: Yes, I totally get your point and agree. But we then need to manually scan the current code and make this refactoring so we don't get other cases like this.

danielRep: For the time being, I would for leaving the code as it is for now

DavidMCerdeira: I made the changes here
https://github.com/bao-project/bao-hypervisor/tree/tmp/style_clang_format_align

there are two commits, one where I think align works well, the other where I think it's not ideal.
IMO the benefits outweigh the costs, but I'm fine with whatever decision for the time being

josecm: When enabling that option, I can't find a way to get around of it putting multiple entries per line on arrays, instead on one entry per line. Like [here](https://github.com/bao-project/bao-hypervisor/blob/a0adc4adeedc7e9ba0ed83768fd06fec090f9399/src/arch/armv8/aborts.c#L154C1-L159C3). I'd suggest we keep it as it is right now.

josecm:
If no one has further comments, I bumping this to full PR.

josecm:
We need to first merge bao-project/bao-ci#70, then update the ci submodule here, and only then can we merge this.

src/arch/armv8/armv8-a/inc/arch/smmuv2.h
```diff
@@ -8,108 +8,108 @@
...
+#define SMMUV2_IDR2_PTFSv8_4kB_BIT      (0x1 << 12)
+#define SMMUV2_IDR2_OAS_OFF             (4)
+#define SMMUV2_IDR2_OAS_LEN             (4)
+#define SMMUV2_IDR2_IAS_OFF             (0)
+#define SMMUV2_IDR2_IAS_LEN             (4)
+
+#define SMMUV2_IDR7_MAJOR_OFF           (4)
+#define SMMUV2_IDR7_MAJOR_LEN           (4)
+
+#define SMMU_SMR_ID_OFF                 0
```
danielRep, line 42: is there any option that can force this macro defines to be enclosed in parenthesis? To maintain consistency

src/core/mem.c
```diff
@@ -264,19 +269,21 @@ bool mem_vm_img_in_phys_rgn(struct vm_config* vm_config) {
...
         struct ppages ppages = mem_ppages_get(vm_cfg->image.load_addr, n_pg);
 
         // If the vm image is part of a statically allocated region of the same
         // vm, we defer the reservation of this memory to when we reserve the
-        // physical region below. Note that this not allow partial overlaps. If the
-        // image must be entirely inside a statically allocated region, or
+        // physical region below. Note that this not allow partial overlaps. If
+        // the image must be entirely inside a statically allocated region, or
         // completely outside of it. This avoid overcamplicating the reservation
         // logic while still covering all the useful use cases.
```
danielRep, line 288: After this PR is accepted we need to rewrap comment lines to 100ch limit.

src/lib/inc/util.h
```diff
@@ -9,81 +9,78 @@
...
+#define XSTR(s)   STR(s)
 
 #ifndef __ASSEMBLER__
 
 #define DEFINE_OFFSET(SYMBOL, STRUCT, FIELD) \
     asm volatile("\n-> " XSTR(SYMBOL) " %0 \n" : : "i"(offsetof(STRUCT, FIELD)))
 
-#define DEFINE_SIZE(SYMBOL, TYPE) \
-    asm volatile("\n-> " XSTR(SYMBOL) " %0 \n" : : "i"(sizeof(TYPE)))
+#define DEFINE_SIZE(SYMBOL, TYPE) asm volatile("\n-> " XSTR(SYMBOL) " %0 \n" : : "i"(sizeof(TYPE)))
```
danielRep, line 39: both macros are being wrapped differently.

danielRep (dismissed):

src/core/vm.c
```diff
@@ -167,29 +165,27 @@ static void vm_init_ipc(struct vm* vm, const struct vm_config* config)
...
         }
-        
+
         spin_lock(&shmem->lock);
         shmem->cpu_masters |= (1ULL << cpu()->id);
         spin_unlock(&shmem->lock);
 
-        struct vm_mem_region reg = {
-            .base = ipc->base,
+        struct vm_mem_region reg = { .base = ipc->base,
```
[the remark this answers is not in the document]
josecm: Missed a trailing comma, that forces clang-format to make initialization one field per line. Fixed.

src/core/mem.c
josecm: We can do it in this PR.

src/lib/inc/util.h
josecm: This happens because the macro definition now fits the 100 chars.

DavidMCerdeira (dismissed):

danielRep (approved):

src/core/mem.c
danielRep: Wrapped all comments to 100 in commit [59fecb5](https://github.com/bao-project/bao-hypervisor/pull/103/commits/59fecb5de26527b45a54f99ad9dd530dd8336c90)

danielRep (approved):

DavidMCerdeira (approved):

src/core/mem.c
josecm: thanks!

src/arch/armv8/aborts.c
josecm: We'll leave this as is for now.

ninolomata (approved):

PR #107, opened 2023-11-17
Refactor riscv sstc ext check

josecm:
This PR centralizes the enabling of the RISCV sstc extension in the vmm architectural initialization function. It also adds the sanity checking that the extension is effectively present if the hypervisor was configured to used it (via a platform-defined macro).

Depends on #106

### DIFF
--- a/src/arch/riscv/vmm.c
+++ b/src/arch/riscv/vmm.c
@@ -22,6 +22,24 @@ void vmm_arch_init()
     CSRW(CSR_HEDELEG, HEDELEG_ECU | HEDELEG_IPF | HEDELEG_LPF | HEDELEG_SPF);
 
     /**
+     * Enable and sanity check presence of Sstc extension if the hypervisor was
+     * configured to use it (via the CPU_EXT_SSTC macro). Otherwise, make sure
+     * it is disabled.
+     */
+    if (CPU_HAS_EXTENSION(CPU_EXT_SSTC)) {
+        CSRS(CSR_HENVCFG, HENVCFG_STCE);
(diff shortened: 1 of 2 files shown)


src/arch/riscv/vmm.c
```diff
@@ -21,6 +21,24 @@ void vmm_arch_init()
...
     CSRW(CSR_HEDELEG, HEDELEG_ECU | HEDELEG_IPF | HEDELEG_LPF | HEDELEG_SPF);
 
+    /**
+     * Enable and sanity check presence of Sstc extension if the hypervisor was
+     * configured to use it (via the CPU_EXT_SSTC macro). Otherwise, make sure
+     * it is disabled.
+     */
+    if (CPU_HAS_EXTENSION(CPU_EXT_SSTC)) {
+        CSRS(CSR_HENVCFG, HENVCFG_STCE);
+        bool sttc_present = (CSRR(CSR_HENVCFG) & HENVCFG_STCE) != 0;
```
ninolomata, line 31: there is a typo here sttc_present should be sstc_present.

ninolomata (changes requested):

src/arch/riscv/vmm.c
josecm: fixed!

josecm: fixed!

ninolomata (approved):

DavidMCerdeira (approved):

AfonsoSantos96 (approved):

PR #108, opened 2023-11-23
Remove debug build by default and use debug optimizations

josecm:


danielRep (approved):

DavidMCerdeira (commented):
We should use -Og for debugging

https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html
>-Og: Optimize debugging experience. -Og should be the optimization level of choice for the standard edit-compile-debug cycle, offering a reasonable level of optimization while maintaining fast compilation and a good debugging experience. It is a better choice than -O0 for producing debuggable code because some compiler passes that collect debug information are disabled at -O0.

DavidMCerdeira (approved):

danielRep (approved):

PR #116, opened 2023-12-10
Add support for 32-bit RISC-V

josecm:
This PR introduces support for RV32. Because the way the instruction set is written the difference is not much when comparing to RV64. The main areas of modification are:

- modifying assembly instructions that used 64-bit load/store directly by the LOAD/STORE macro which translates to either ld/st or lw/sw depending on if we are targetting RV64 or RV32, respectively;
- support for the sv32 translation mode. This also requires adapting the boot code where the bootstrap page-tables are built manually.
- adapt CSR access to CSRs that have a H variant (e.g., henvcfg, henvcfgh), and other small related issues;
- fix small issues that assumed 64-bit literals.

The PR also includes support for qemu-system-riscv32 virt machine, which is essentially a wrapper over the 64-bit variant.

The main issue with this port is related to the fact that we are required to map all available physical memory to perform page-table walks. Because the address space is so small (4GiB), this might not be possible as for RV64 we simply identity map all physical memory and assume these addresses won't collide with the high virtual addresses used by the hypervisor. This is not the case and we might not be able to map all physical addresses, which already happens for qemu-riscv32-virt if we configure it with memory larger than 1 GiB, since it starts memory at 0x80000000. We might need a dynamic mechanism similar to Linux's "highmem" in the future. I only think this is worth it if a real platform that requires it becomes available. Also, the main purpose of supporting RV32 is to target future MMU-less systems.

Test guest:
- [x]  Baremetal app
- [x] Linux

Depends on #115

### DIFF
--- a/src/arch/riscv/vm.c
+++ b/src/arch/riscv/vm.c
@@ -46,6 +46,9 @@ void vcpu_arch_reset(struct vcpu* vcpu, vaddr_t entry)
 
     if (CPU_HAS_EXTENSION(CPU_EXT_SSTC)) {
         csrs_stimecmp_write(~0U);
+        csrs_henvcfg_set(HENVCFG_STCE);
+    } else {
+        csrs_henvcfg_clear(HENVCFG_STCE);
     }
 
     csrs_hcounteren_write(HCOUNTEREN_TM);
--- a/src/arch/riscv/vm.c
+++ b/src/arch/riscv/vm.c
@@ -38,7 +38,10 @@ void vcpu_arch_reset(struct vcpu* vcpu, vaddr_t entry)
 
     csrs_sscratch_write((uintptr_t)&vcpu->regs);
 
-    vcpu->regs.hstatus = HSTATUS_SPV | HSTATUS_VSXL_64;
+    vcpu->regs.hstatus = HSTATUS_SPV;
+    if (RV64) {
+        vcpu->regs.hstatus |= HSTATUS_VSXL_64;
+    }
     vcpu->regs.sstatus = SSTATUS_SPP_BIT | SSTATUS_FS_DIRTY | SSTATUS_XS_DIRTY;
     vcpu->regs.sepc = entry;
     vcpu->regs.a0 = vcpu->arch.hart_id = vcpu->id;
--- a/src/arch/riscv/arch.mk
+++ b/src/arch/riscv/arch.mk
@@ -1,7 +1,21 @@
 ## SPDX-License-Identifier: Apache-2.0
 ## Copyright (c) Bao Project and Contributors. All rights reserved.
 
+ARCH_SUB?=riscv64
+
+ifeq ($(ARCH_SUB), riscv64)
 CROSS_COMPILE ?= riscv64-unknown-elf-
+riscv_march:=rv64imac_zicsr
+riscv_mabi:=lp64
+ld_emulation:=elf64lriscv
+else ifeq ($(ARCH_SUB), riscv32)
+CROSS_COMPILE ?= riscv32-unknown-elf-
+riscv_march:=rv32imac_zicsr
+riscv_mabi:=ilp32
+ld_emulation:=elf32lriscv
+else
+$(error RISC-V $(ARCH_SUB) not supported!)
+endif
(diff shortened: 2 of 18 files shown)


josecm:
I have booted 32-bit Linux over Bao too. 

I think this is ready to be fully reviewed  @ninolomata @DavidMCerdeira @danielRep

danielRep:
I still couldn't finish my review , but as a first note we still need a PR on bao-demos to support 32-bit RISC-V as a means to test this PR.

src/arch/riscv/vm.c
```diff
@@ -46,6 +46,9 @@ void vcpu_arch_reset(struct vcpu* vcpu, vaddr_t entry)
 
     if (CPU_HAS_EXTENSION(CPU_EXT_SSTC)) {
         csrs_stimecmp_write(~0U);
+        csrs_henvcfg_set(HENVCFG_STCE);
+    } else {
+        csrs_henvcfg_clear(HENVCFG_STCE);
```
danielRep, line 54: Is this clear necessary?

"These STCE bits are WARL and are hard-wired to 0 when this extension is not implemented."

src/core/mmu/vmm.c
```diff
@@ -16,14 +16,14 @@ struct vm_install_info vmm_get_vm_install_info(struct vm_allocation* vm_alloc)
 {
     struct vm_install_info info = {
         .base = vm_alloc->base,
-        .vm_section_pte = *pt_get_pte(&cpu()->as.pt, 0, vm_alloc->base),
+        .vm_section_pte = *pt_get_pte(&cpu()->as.pt, VM_SHARED_PT_LVL, vm_alloc->base),
```
danielRep, line 19: On file  `src/core/mmu/mem.c` , shouldn't the parameter on the functions `pt_get_pte` (line  810/812) be changed to use `VM_SHARED_PT_LVL` ?

danielRep (commented):
While I don't see a huge problem on using compilation macros macros (RV32 and RV64), wouldn't be more cleaner to separate rv32 and rv64 profiles on different `arch` folders as we do for `aarch32` and `aarch64`?

josecm:
A branch for bao-demos featuring demos targetting qemu-riscv32-virt is available at https://github.com/josecm/bao-demos/tree/feat/rv32

src/core/mmu/vmm.c
josecm: I think this was already fixed by another PR. It was "fixed" after a rebase

src/arch/riscv/vm.c
josecm: This is not really relevant for this PR. But I believe this is already fixed in the main branch.

josecm:
> While I don't see a huge problem on using compilation macros macros (RV32 and RV64), wouldn't be more cleaner to separate rv32 and rv64 profiles on different `arch` folders as we do for `aarch32` and `aarch64`?

The difference between rv32 and rv64 is not as stark as from aarch32 to aarch64, so it does not seem necessary to introduce both architectures separately, only a few of those RV32/RV64 macros and a wrapper around csr access so that rv32 can access "64-bit" csrs is enough and avoids a lot of code duplication and indirection.

danielRep:
> > While I don't see a huge problem on using compilation macros macros (RV32 and RV64), wouldn't be more cleaner to separate rv32 and rv64 profiles on different `arch` folders as we do for `aarch32` and `aarch64`?
> 
> The difference between rv32 and rv64 is not as stark as from aarch32 to aarch64, so it does not seem necessary to introduce both architectures separately, only a few of those RV32/RV64 macros and a wrapper around csr access so that rv32 can access "64-bit" csrs is enough and avoids a lot of code duplication and indirection.

Seems reasonable. Ok, look good to me.

danielRep:
> A branch for bao-demos featuring demos targetting qemu-riscv32-virt is available at https://github.com/josecm/bao-demos/tree/feat/rv32. I'll create a PR for that demo once this PR is accepted.

Good!

DavidMCerdeira (approved):

ninolomata (approved):

PR #120, opened 2024-01-26
Add compatibility with standard C11

miguelafsilva5:
## PR description

The usage of a C standard without any extensions is a required rule in MISRA compliance.
This PR will serve as a hub for most modifications regarding the standard C11 and additonal MISRA rules that can be verified by the compiler.
Ideally, each PR made to this WIP will represent a rule or a subset of MISRA rules that are verified with additional compiling flags.

- [ ] C11 with no extensions
- [ ] TBD

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -189,7 +189,7 @@ ifeq ($(DEBUG), y)
 	OPTIMIZATIONS:=g
 endif
 
-override CFLAGS+=-O$(OPTIMIZATIONS) -Wall -Werror -ffreestanding -std=gnu11 \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -235,7 +235,7 @@ $(build_dir)/%.d : $(src_dir)/%.[c,S]
 
 $(objs-y):
 	@echo "Compiling source	$(patsubst $(cur_dir)/%, %, $<)"
-	@$(cc) $(CFLAGS) -c $< -o $@
+	-@$(cc) $(CFLAGS) -c $< -o $@
```
josecm, line 238: We should not be suppressing build errors.

miguelafsilva5: Should be fixed in the recent commit.

josecm:
Closing this at it is already addressed in #122

PR #124, opened 2024-02-09
feat(core/shmem): move shared memory logic to dedicated source file

joaopeixoto13:
This patch relocates the initialization and shared memory access steps from the hardcoded settings in the existing IPC file to a generic shared memory file. Previously, the code assumed the existence of only one object working with the shared memory (ipc). With this update, the system is now prepared for future versions where multiple objects can be added and working with the shared memory.

### DIFF
--- a/src/core/inc/ipc.h
+++ b/src/core/inc/ipc.h
@@ -21,6 +21,5 @@ struct vm_config;
 
 long int ipc_hypercall(unsigned long arg0, unsigned long arg1, unsigned long arg2);
 void ipc_init(void);
-struct shmem* ipc_get_shmem(size_t shmem_id);
 
 #endif /* IPC_H */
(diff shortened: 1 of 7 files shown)


danielRep (dismissed):
Look's good to me :+1:

DavidMCerdeira (commented):
Seems to be working. I've tested this for riscv64.

My only issue is the wording of the commit title.
I don't think "custom" is the correct term here.

Something like this is better:
feat(core/shmem): move shared memory logic to dedicated source file

DavidMCerdeira (approved):

DavidMCerdeira:
Needs rebase with main @joaopeixoto13

josecm (approved):

danielRep (approved):

PR #133, opened 2024-03-25
fix(smmuv2): fix smmu_cntxt_hw struct

Diogo21Costa:
This pull request addresses an issue in the SMMUv2 struct definition (`smmu_cntxt_hw`) where `PMEVCNTRm` and `PMEVTYPERm` were incorrectly defined as single 32-bit registers, when in fact, they are arrays of 15 positions of 32-bit registers each. This fix ensures that the struct accurately matches the smmuv2 specification of these registers within the SMMU context hardware.


danielRep (approved):
Seems correct per the ARM SMMU architecture v2.0, page 150

> The SMMU architecture permits a maximum of:
> • 256 implemented event counters in total.
> • 256 Counter groups.
> • 15 counters in a Counter group.

DavidMCerdeira (approved):

PR #157, opened 2024-07-09
Guest reboot support

josecm:
This PR adds support for VM reset. Guest can request a VM reset by issuing standard system calls, i.e., PSCI_RESET and SBI_SRST, for arm and risc-v systems, respectively.

It starts by adding support for allowing CPUs to go idle in the middle of an interrupt. This is necessary because CPUs will receive an IPI from the CPU that emitted the reset and will be required to go idle if they are not the VM's primary CPU.

Besides the generic VM rest logic and the arch-specific calls for requesting a reset, this PR adds arch-specific logic to reset the virtual interrupt controllers.


danielRep (commented):
Just tested with the baremetal demo and it seems to be correctly working for `qemu-aarch64-virt`, but for `qemu-riscv64-virt` bao crashes after some reboot iterations.

PR #164, opened 2024-07-28
fix(armv8/vgic): fix comparison on pend_found

Clara-Huang:
If a suitable pending interrupt is found, it should be replaced rather than checking if `pend_found` is greater than 1. Thus, I have changed `> 1` to `>= 1`.
If my understanding is incorrect, I will close this PR. Thank you for your review!

### DIFF
--- a/src/arch/armv8/vgic.c
+++ b/src/arch/armv8/vgic.c
@@ -360,7 +360,7 @@ bool vgic_add_lr(struct vcpu* vcpu, struct vgic_int* interrupt)
             }
         }
 
-        if (pend_found > 1) {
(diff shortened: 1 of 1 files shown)


josecm:
I'm not sure this fix is correct. The logic here is that if we found more than one pending interrupt, we can replace one of them, because as suggested by the GIC spec, if there are any pending interrupts, at least one pending interrupt should be in the list. Do you agree @Clara-Huang @DavidMCerdeira ?

josecm:
@Clara-Huang I'm closing this for now, unless there is any new arguments that can convince me my rationale is wrong.

PR #167, opened 2024-08-13
Allow for ci submodule not be present

josecm:
If this reposiroty is not cloned with '--recursive', the ci subdmoudle will not be prenset which will result in any invocation of make to fail. This modification allows that if the ci submodule is not present, it will simply be ignored and typical make calls for building the hypervisor will continue normally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ scripts_dir:=$(cur_dir)/scripts
 ci_dir:=$(cur_dir)/ci
 src_dirs:=
 
-include $(ci_dir)/ci.mk
+-include $(ci_dir)/ci.mk
 
 targets:=$(MAKECMDGOALS)
 ifeq ($(targets),)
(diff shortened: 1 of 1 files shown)


danielRep (dismissed):

Makefile
```diff
@@ -47,13 +47,13 @@ scripts_dir:=$(cur_dir)/scripts
...
 
-include $(ci_dir)/ci.mk
+-include $(ci_dir)/ci.mk
 
 targets:=$(MAKECMDGOALS)
 ifeq ($(targets),)
 targets:=all
 endif
-non_build_targets+=ci clean
+non_build_targets+=ci cleanc
```
DavidMCerdeira, line 56: Is this intended? Seems like it may be a typo.

josecm: Fixed!

danielRep (approved):

DavidMCerdeira (approved):

PR #172, opened 2024-09-14
feat(inc/list): add list size function

joaopeixoto13:


DavidMCerdeira (approved):

josecm:
This PR looks good! However, without other commits it does not make any sense, because nowhere in the hypervisor it is being used for now. I say this commit should be part of the larger PR where it is actually used!

DavidMCerdeira:
That's a good point @josecm !
Even though there's no other problems with it, I'll close the PR
@joaopeixoto13 follow @josecm and add this as part of a larger PR

PR #173, opened 2024-09-14
fix(config): vm_dev_region struct field order, type casting, and macro usage

joaopeixoto13:


DavidMCerdeira (dismissed):

src/core/inc/vm.h
```diff
@@ -30,9 +30,9 @@ struct vm_dev_region {
     paddr_t pa;
     vaddr_t va;
     size_t size;
+    deviceid_t id; /* bus master id for iommu effects */
     size_t interrupt_num;
     irqid_t* interrupts;
-    deviceid_t id; /* bus master id for iommu effects */
```
josecm, line 35: WHy do we need to change this?

josecm:
@joaopeixoto13 please make sure you indicate this is in the example config in the commit message

src/core/inc/vm.h
joaopeixoto13: When structs are initialized, fields are processed in the order they are defined. If a pointer field is declared first, it may be implicitly initialized to _NULL_, which is generally safe. However, when a scalar field is placed after a scalar pointer field, the compiler may not apply proper initialization, resulting in the scalar field being left uninitialized. Compilers with the `-Wuninitialized` flag (flag included in this [PR](https://github.com/bao-project/bao-hypervisor/commit/dea07f2d410360eee22c1d3fd47915b99016af85) may raise an error in such cases. By placing the scalar field first, we ensure it is properly initialized, as pointers are more lenient and typically default to _NULL_, avoiding the warning.

joaopeixoto13:
> @joaopeixoto13 please make sure you indicate this is in the example config in the commit message

The two commit messages already include the header `fix(configs/example)`

src/core/inc/vm.h
joaopeixoto13: > > When structs are initialized, fields are processed in the order they are defined. If a pointer field is declared first, it may be implicitly initialized to _NULL_, which is generally safe. However, when a scalar field is placed after a scalar pointer field, the compiler may not apply proper initialization, resulting in the scalar field being left uninitialized.
> 
> I don't really understand why this is. Can you point me to some C or GCC documentation that delves into it?

I attempted to find proper documentation regarding this issue but was unsuccessful. It's peculiar because the warning only appears for pointers to scalar values (e.g., irqid_t* interrupts), but not for pointers to structs (e.g., struct ipc* ipcs). It seems that placing the scalar value before the scalar pointer resolved the issue, and the code now works as expected. This behavior might be due to a potential compiler bug, as I'm using `aarch64-none-elf-gcc (GNU Toolchain for the Arm Architecture 11.2-2022.02 (arm-11.14)) 11.2.1 20220111`, rather than the newer `aarch64-none-elf-gcc (Arm GNU Toolchain 13.2.rel1 (Build arm-13.7)) 13.2.1 20231009`, which is defined in [bao-ci's container](https://github.com/bao-project/bao-ci/blob/main/docker/Dockerfile). If you tested with the latter version and if it works, let me know.

josecm: We don't get that error with the latest version. @miguelafsilva5 can you please shime in?

> However, when a scalar field is placed after a scalar pointer field, the compiler may not apply proper initialization, resulting in the scalar field being left uninitialized.

@joaopeixoto13 If there is no information available, how did you arrive at this conclusion?

joaopeixoto13: > We don't get that error with the latest version. @miguelafsilva5 can you please chime in?
> 
> > However, when a scalar field is placed after a scalar pointer field, the compiler may not apply proper initialization, resulting in the scalar field being left uninitialized.
> 
> @joaopeixoto13 If there is no information available, how did you arrive at this conclusion?

Checking the compiler error message

joaopeixoto13: > We don't get that error with the latest version. @miguelafsilva5 can you please chime in?
> 
> > However, when a scalar field is placed after a scalar pointer field, the compiler may not apply proper initialization, resulting in the scalar field being left uninitialized.
> 
> @joaopeixoto13 If there is no information available, how did you arrive at this conclusion?

@josecm I tested with both compilers, and the most recent version does not present any issues.

josecm: > @josecm I tested with both compilers, and the most recent version does not present any issues.

Great! Nevertheless, I think @miguelafsilva5 should have the last word if this is really necessary or not.

> Checking the compiler error message

Could you please post it here? I would like to understand the issue better.

joaopeixoto13: > > @josecm I tested with both compilers, and the most recent version does not present any issues.
> 
> Great! Nevertheless, I think @miguelafsilva5 should have the last word if this is really necessary or not.
> 
> > Checking the compiler error message
> 
> Could you please post it here? I would like to understand the issue better.

Here is the full compiler error message:
image

Below is the configuration file with the initialized `id` field that triggers the compiler error:
```
{
      // Virtio devices
      .pa = 0xa003000,                    			        
      .va = 0xa003000,                    			       
      .size = 0x1000,                     			            
      .interrupt_num = 8,                 			        
      .interrupts = (irqid_t[]) {72,73,74,75,76,77,78,79}
}
```

joaopeixoto13: > I get the error. But don't get how you arrived at your conclusion. Trial and error? You moved the fields around and saw the error disappeared? If so, since the latest compiler version does not trigger the error, I say we should close the PR.

Yes, I tried several configurations and found that moving the scalar value declaration before the scalar pointer resolved the issue.

miguelafsilva5: I believe I had a similar error with previous compiler versions. If the most recent version does not issue any error, it should be ok. 

Specially since the error occurs in the config file, where we can argue if the `-Wuninitialized` should be applied to it.

DavidMCerdeira: @joaopeixoto13 please confirm that this is not issue with current compiler versions. And if it is not, remove this commit from the PR

joaopeixoto13: @DavidMCerdeira Done!

DavidMCerdeira (approved):

miguelafsilva5 (approved):

PR #179, opened 2024-09-27
feat(armv8/psci): Handle unexpected psci behavior

DavidMCerdeira:
We've observed that some platforms behave unexpectedly when performing power down, and others when performing standby. In these cases, the CPU cores are not awaken by interrupts as expected. These commit solve this, by adding the ability to skip the respective firmware call, instead relying on the fallback mechanism of using wfi.


danielRep (approved):

josecm:
@DavidMCerdeira Shouldn't we address other calls to `psci_power_down` such as in `psci_cpu_suspend_handler`

DavidMCerdeira:
This is a good point
Let me first make a few tests to make sure the PSCI_WAKEUP_IDLE behavior is the same as the PSCI_STATE_TYPE_STANDBY

josecm:
> This is a good point Let me first make a few tests to make sure the PSCI_WAKEUP_IDLE behavior is the same as the PSCI_STATE_TYPE_STANDBY

Please make sure we are also inline with the PSCI in this respect.

src/arch/armv8/armv8-a/psci.c
```diff
@@ -82,6 +82,12 @@ void psci_wake(uint32_t handler_id)
 
 int32_t psci_standby()
 {
+    /* We've observed that some platforms behave unexpectedly when performing
+     * standby. In these cases, after standby, the CPU cores are not awaken
+     * by interrupts. */
+    if (!DEFINED(PLAT_PSCI_STANDBY_NOT_SUPPORTED)) {
```
josecm, line 88: Should if be:

```suggestion
    if (DEFINED(PLAT_PSCI_STANDBY_NOT_SUPPORTED)) {
```

src/arch/armv8/armv8-a/psci.c
```diff
@@ -90,6 +96,13 @@ int32_t psci_standby()
 
 int32_t psci_power_down(enum wakeup_reason reason)
 {
+    /* We've observed that some platforms behave unexpectedly when performing
+     * power down. In these cases, after powerdown, the CPU cores are not awaken
+     * by interrupts as expected by Bao. */
+    if (!DEFINED(PLAT_PSCI_POWERDOWN_NOT_SUPPORTED)) {
```
josecm, line 102:
```suggestion
    if (DEFINED(PLAT_PSCI_POWERDOWN_NOT_SUPPORTED)) {
```

Same as above

src/arch/armv8/psci.c
```diff
@@ -69,14 +69,11 @@ static int32_t psci_cpu_suspend_handler(uint32_t power_state, unsigned long entr
...
-         * understand why. To circunvent this, we directly emmit a wfi
-         */
-        // ret = psci_standby();
-        __asm__ volatile("wfi\n\r");
-        ret = PSCI_E_SUCCESS;
+        ret = psci_standby();
+        if (ret == PSCI_E_NOT_SUPPORTED) {
+            __asm__ volatile("wfi\n\r");
+            ret = PSCI_E_SUCCESS;
+        }
```
josecm, line 76: I think we could incorporate this inside the `psci_standby` function itself. Because `wfi`is a legal implementation of that PSCI service.

josecm (changes requested):

danielRep (approved):

josecm (approved):

danielRep (approved):

PR #184, opened 2024-10-15
fix(aarch32/sysregs): fix the 64-bit register read function

AfonsoSantos96:
The usage of `unsigned long long` for both variables may introduce unexpected values in the OR operation in the return statement, as the `_temp`  variable is not guaranteed to be initialized to 0.

### DIFF
--- a/src/arch/armv8/aarch32/inc/arch/subarch/sysregs.h
+++ b/src/arch/armv8/aarch32/inc/arch/subarch/sysregs.h
@@ -40,7 +40,7 @@
     {                                                                                             \
         unsigned long long _temp, _tempH;                                                         \
         __asm__ volatile("mrrc p15, " #op1 ", %0, %1, " #crm "\n\r" : "=r"(_temp), "=r"(_tempH)); \
-        return ((_tempH << 32) | _temp);                                                          \
+        return ((_tempH << 32) | ((unsigned long)_temp));                                         \
     }                                                                                             \
     static inline void sysreg_##reg##_write(unsigned long long val)                               \
     {                                                                                             \


src/arch/armv8/aarch32/inc/arch/subarch/sysregs.h
```diff
@@ -38,7 +38,8 @@
 #define SYSREG_GEN_ACCESSORS_64(reg, op1, crm)                                                    \
     static inline unsigned long long sysreg_##reg##_read(void)                                    \
     {                                                                                             \
-        unsigned long long _temp, _tempH;                                                         \
+        unsigned long long _tempH;                                                                \
+        unsigned long _temp;                                                                      \
         __asm__ volatile("mrrc p15, " #op1 ", %0, %1, " #crm "\n\r" : "=r"(_temp), "=r"(_tempH)); \
         return ((_tempH << 32) | _temp);                                                          \
```
josecm, line 44: I see the problem. But maybe we can add the explicit cast at the end:

```suggestion
        return ((_tempH << 32) | ((unsigned long) _temp));                                                          \
```

josecm (changes requested):

src/arch/armv8/aarch32/inc/arch/subarch/sysregs.h
AfonsoSantos96: Yes, that also works! I'll change to your suggestion.
Thanks!

miguelafsilva5 (dismissed):

josecm (approved):

miguelafsilva5 (approved):

PR #186, opened 2024-10-22
PSCI: fix some PSCI issues and decouple CPU standby from CPU power down

joaopeixoto13:
This PR is divided into three parts:
- Fix QEMU PSCI power down and standby states
- Remove unnecessary cpu_idle and vcpu_arch_run calls
- Decouple CPU standby from CPU power down


josecm:
@Diogo21Costa could you check this is inline with your previous work in distinguishing idle cpu states?

src/arch/armv8/armv8-a/cpu.c
```diff
@@ -27,7 +27,15 @@ void cpu_arch_profile_init(cpuid_t cpuid, paddr_t load_addr)
...
-void cpu_arch_profile_idle()
+void cpu_arch_profile_standby()
+{
+    int32_t err = psci_standby();
+    if (err) {
+        ERROR("PSCI cpu%d standby failed with error %ld", cpu()->id, err);
+    }
+}
+
+void cpu_arch_profile_powerdown()
```
Diogo21Costa, line 38: In the recent changes, we implemented checks for active interrupts before calling the `powerdown` function, as detailed in this [commit](https://github.com/AFOliveira/bao-hypervisor-mem-throttling/commit/b908dd5e80dadf6ec6046c8d469625bd1428651a). The same logic applies to the `standby` function.

@josecm, do you think this approach is also appropriate in this context?

src/arch/armv8/armv8-a/cpu.c
```diff
@@ -27,7 +27,15 @@ void cpu_arch_profile_init(cpuid_t cpuid, paddr_t load_addr)
     }
 }
 
-void cpu_arch_profile_idle()
+void cpu_arch_profile_standby()
+{
+    int32_t err = psci_standby();
```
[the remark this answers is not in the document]
joaopeixoto13: Hmm, I don’t think so. I've kept the `psci_standby` function the same. 

Both `psci_power_down` and `psci_standby` functions return a 32-bit integer (int32_t). 
However, in the invocation of `psci_power_down` within the `cpu_arch_profile_powerdown` function (and previously in `cpu_arch_profile_idle`), we cast it to a 64-bit variable. 
There's apparently no need for the return value to be 64-bit, so the best approach might be to change it to a 32-bit variable, similar to the `psci_standby` calls.

src/arch/armv8/armv8-a/cpu.c
josecm: I believe that is not needed for now. We only need to check if an interrupt is being handled if you go to idle/powerdown in the middle of an interrupt handler, which is not the case at the moment.

src/arch/armv8/armv8-a/cpu.c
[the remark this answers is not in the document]
josecm: Yes! Can you fix it in this PR?

src/arch/armv8/armv8-a/psci.c
```diff
@@ -58,7 +58,7 @@ static void psci_wake_from_powerdown(void)
 
 static void psci_wake_from_idle(void)
 {
-    cpu_idle_wakeup();
+    cpu_standby_wakeup();
```
josecm, line 61: We can change this "idle" nomenclature to "standby"

src/core/vm.c
```diff
@@ -350,9 +350,7 @@ __attribute__((weak)) cpumap_t vm_translate_to_vcpu_mask(struct vm* vm, cpumap_t
 
 void vcpu_run(struct vcpu* vcpu)
 {
-    if (vcpu->active == false) {
-        cpu_idle();
-    } else {
-        vcpu_arch_run(vcpu);
+    if (vcpu_arch_is_on(vcpu) && vcpu->active == true) {
```
josecm, line 353:
```suggestion
    if (vcpu_arch_is_on(vcpu) && vcpu->active) {
```

Shouldn't we call a `cpu_powerdown` if the `!vcpu_arch_is_on(vcpu)`? And also call `cpu_standby` if `vcpu_arch_is_on(vcpu) && !vcpu->active`?

src/platform/qemu-aarch64-virt/inc/plat/psci.h
```diff
@@ -9,8 +9,9 @@
 #define PSCI_POWER_STATE_LVL_0    0x0000000
 #define PSCI_POWER_STATE_LVL_1    0x1000000
 #define PSCI_POWER_STATE_LVL_2    0x2000000
-#define PSCI_STATE_TYPE_STANDBY   0x00000
+#define PSCI_STATE_TYPE_STANDBY   0x1
 #define PSCI_STATE_TYPE_BIT       (1UL << 16)
-#define PSCI_STATE_TYPE_POWERDOWN (0x1)
+#define PSCI_STATE_ID_RETENTION   0x2
+#define PSCI_STATE_TYPE_POWERDOWN PSCI_STATE_TYPE_BIT | PSCI_STATE_ID_RETENTION
```
josecm, line 15: Shouldn't powerdown be non-retentive?

josecm (changes requested):
This PR should be dependent (and rebased) on #187 so that #187 is merged first, and the risc-v `cpu_arch_standby` implementation correctly uses the sbi suspend function.

src/platform/qemu-aarch64-virt/inc/plat/psci.h
joaopeixoto13: I believe there may be some confusion with the power states on RISC-V. In RISC-V, platform-specific suspend states for a hart can be either retentive or non-retentive: a retentive suspend state preserves hart registers and CSR values across all privilege modes, while a non-retentive state does not. 
However, the terminology and structure differ on ARM, where the nomenclature for power states is not the same.

An ARM core can exist in one of four power states:
- Run: The core is fully powered and operational
- Standby: The core remains powered, but energy-saving measures are applied (`wfi`), preserving all core states (this power state is similar to RISC-V retention state)
- Retention: The core is partially powered down, with its state saved and restored upon resuming normal operation (this power state is similar to RISC-V non-retention state)
- Powerdown: The core is fully powered off.

More details:
The ARM `CPU_SUSPEND` API enables a topology node to enter a low-power state. 
The `power_state` bit 16 in the CPU suspend call specifies the state type and bits [15:0] the State ID. 
For power-down, bit 16 is set to 1, and the State ID represents a platform-specific state. 
In [ATF’s QEMU implementation](https://github.com/bao-project/arm-trusted-firmware/blob/bao/master/plat/qemu/common/qemu_pm.c#L54), three idle states are defined: Standby (StateId = 0x1) and two power-down states — one for the core level and one for the cluster level — both categorized as retention states (StateId = 0x2)

josecm: Ok! That makes sense!

josecm:
@joaopeixoto13 please rebase this on main and adapt the RISC-V suspend calls

src/arch/armv8/armv8-a/cpu.c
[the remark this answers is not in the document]
joaopeixoto13: Done [here](https://github.com/bao-project/bao-hypervisor/pull/186/commits/080a24e8bff2f4941d6363ed8f1662e6cda2ae4b)

src/arch/armv8/armv8-a/psci.c
joaopeixoto13: Done [here](https://github.com/bao-project/bao-hypervisor/pull/186/commits/1dba63da1c20589c1bafba8bdf6575df578b2fcb)

joaopeixoto13:
> @joaopeixoto13 please rebase this on main and adapt the RISC-V suspend calls

Done

josecm (dismissed):

src/core/vm.c
```diff
@@ -350,9 +350,11 @@ __attribute__((weak)) cpumap_t vm_translate_to_vcpu_mask(struct vm* vm, cpumap_t
...
-    if (vcpu->active == false) {
-        cpu_idle();
-    } else {
-        vcpu_arch_run(vcpu);
+    if (vcpu_arch_is_on(vcpu) && vcpu->active) {
+        vcpu_arch_entry();
+    } else if (!vcpu_arch_is_on(vcpu)) {
+        cpu_powerdown();
+    } else if (vcpu_arch_is_on(vcpu) && !vcpu->active) {
+        cpu_standby();
```
DavidMCerdeira, line 358: This flow is a bit confusing IMO

Please re-write it like this:
```
void vcpu_run(struct vcpu* vcpu)
{
    if (vcpu_arch_is_on(vcpu)){
        if(vcpu->active) {
            vcpu_arch_entry();
        } else  {
            cpu_standby();
        }
    } else {
        cpu_powerdown();
    }
}
```

src/arch/armv8/armv8-a/cpu.c
```diff
@@ -37,7 +37,7 @@ void cpu_arch_profile_standby()
 
 void cpu_arch_profile_powerdown()
 {
-    int64_t err = psci_power_down(PSCI_WAKEUP_IDLE);
+    int32_t err = psci_power_down(PSCI_WAKEUP_IDLE);
```
DavidMCerdeira, line 40: Please provide a bit of rationale behind this decision/change in the commit message.
From this it is not clear why we are doing this.

src/arch/armv8/armv8-a/psci.c
```diff
@@ -56,17 +56,17 @@ static void psci_wake_from_powerdown(void)
...
 }
 
 void psci_wake_from_off(void);
 
 void (*psci_wake_handlers[PSCI_WAKEUP_NUM])(void) = {
     [PSCI_WAKEUP_CPU_OFF] = psci_wake_from_off,
     [PSCI_WAKEUP_POWERDOWN] = psci_wake_from_powerdown,
-    [PSCI_WAKEUP_IDLE] = psci_wake_from_idle,
+    [PSCI_WAKEUP_IDLE] = psci_wake_from_standby,
```
DavidMCerdeira, line 69: > PSCI_WAKEUP_IDLE

Shouldn't this be renamed to standby or maybe suspend?

src/core/vm.c
joaopeixoto13: > This flow is a bit confusing IMO
> 
> Please re-write it like this:
> 
> ```
> void vcpu_run(struct vcpu* vcpu)
> {
>     if (vcpu_arch_is_on(vcpu)){
>         if(vcpu->active) {
>             vcpu_arch_entry();
>         } else  {
>             cpu_standby();
>         }
>     } else {
>         cpu_powerdown();
>     }
> }
> ```

Done [here](https://github.com/bao-project/bao-hypervisor/pull/186/commits/8adff2474cd9cabb4b61c1bf40482538571af909)

DavidMCerdeira (approved):

josecm (approved):

src/arch/armv8/armv8-a/psci.c
josecm: I think João changed this already.

PR #196, opened 2024-11-27
Fix(extra mem):  Fix additonal memory allocation for MPU-based systems

miguelafsilva5:
## PR description

This PR fixes a problem identified for systems based on MPUs, where additional memory was being allocated for VMs in the linkerscript. This is only required for MMU-based ones. 
Additionally, this PR also fixes a small mistake where the root pool bitmap base was being allocated in a wrong place for systems with MPU.


DavidMCerdeira (commented):
Code looks alright, but commit message is confusing.
Here's my proposal:

> For MMU systems we manage the location counter to relocate VM images virtual address space.
This allows us to maintain a contiguous virtual address for each cpu between the hypervisor image and it’s private memory. However, for non-mmu systems this operation results in extra memory being erroneously allocated for the vms, as this space is already allocated between vm_image_start and vm_image_end.

josecm:
Shouldn't this also address calculations in `pp_root_reserve_hyp_mem`?

miguelafsilva5:
This PR is obsolete with the introduction of PR #218

PR #197, opened 2024-12-07
Fix: CPU power down

joaopeixoto13:


src/core/cpu.c
```diff
@@ -127,6 +127,11 @@ void cpu_standby_wakeup(void)
 
 void cpu_powerdown_wakeup(void)
```
josecm, line 128: I think we can merge this with `cpu_standby_wakeup` and have a single wake up function

DavidMCerdeira: For me it's ok how it is currently
IMO in this case it is better to have the two separate but similar functions

DavidMCerdeira (approved):

josecm (approved):

Issue #199, opened 2024-12-26
Possibly incorrect allocation in function pp_alloc_clr


nkosmatov:
While working on formal verification of cache coloring and page coloring mechanisms in Bao, we discovered two issues in function `pp_alloc_clr`

The following line (currently line 138 in commit c306b0f  in file src/core/mmu/mem.c)
`
index ++;
`

should be removed. Otherwise , in some situations, a previously allocated page can be allocated again,  or other unintended behavior can occur. 

The following line (currently line 161 in commit c306b0f  in file src/core/mmu/mem.c)
`
 index = 0;
`
should be replaced by 
`
index = pp_next_clr ( pool->base , 0 , colors ) ;
`
Otherwise , in some situations, a previously allocated page can be allocated again.

After the proposed modifications we were able to prove a (slightly simplified) corrected version of cache coloring and page coloring mechanisms in Bao.

josecm:
Hello @nkosmatov! 

> After the proposed modifications we were able to prove a (slightly simplified) corrected version of cache coloring and page coloring mechanisms in Bao.

This is amazing!! Thank you for doing that work.

Your corrections seem to make sense. I'd propose you send a PR with the fixes you point out.

josecm:
Fixed by #202

PR #202, opened 2025-01-02
fix(mem): correctly allocate colored pages

nkosmatov:
fix(mem): correctly allocate colored pages
Fixes Issue #199

### DIFF
--- a/src/core/mmu/mem.c
+++ b/src/core/mmu/mem.c
@@ -134,8 +134,6 @@ bool pp_alloc_clr(struct page_pool* pool, size_t n, colormap_t colors, struct pp
                 allocated++;
                 index = pp_next_clr(pool->base, ++index, colors);
             }
-
-            index++;
         }
 
         if (allocated == n) {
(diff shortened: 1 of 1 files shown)


josecm:
@nkosmatov thanks for sending this fix! 

The fixes themselves look great. Just a couple of nits regarding the PR itself that I'd ask to change:

- For commit messages we follow the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) scheme. So I'd ask that you replace the messaage with somehting like: `fix(mem): correctly allocate colored pages`
- We have a coding style enforced by clang-format. You can apply it to your modifications simply by running `make format`.

nkosmatov:
Thank you for considering my commit. I am not a Bao developer and I am not used to the conventions. I created the commit from the website `github.com `and do not see how I can take into account your requests from the current commit. Should I create a new commit with another title? But I do not think that `make format` can be run from the website? I can try to do if you send me step-by-step instructions. Or maybe a more experienced Bao developer will be able to do the necessary PR properly and it will be faster, as the corrections are very limited.

josecm (approved):

DavidMCerdeira (approved):

PR #209, opened 2025-03-31
fix(riscv/irqc): align IMSIC handler to AIA specifications

filippofontana:
"The RISC-V Advanced Interrupt Architecture" specifications explain in Chapter 3 (pages 26-27) how interrupts should be handled with IMSIC, in particular, interrupt claiming has to be done right after reading the STOPEI register or in the same read instruction with the csrrw instruction.

This commit aligns the IMSIC interrupt handler to that specs. Without this change, if an interrupt with higher priority becomes pending while handling a lower one, writing to STOPEI after `interrupts_handle()` executes will claim the higher IRQ, freezing the execution.


filippofontana:
Had to force push because of the failing format check, I wanted to keep the change small but clang-format doesn't like it :smiling_face_with_tear:.

josecm (approved):

josecm:
Thanks again for fixing this @filippofontana!

josecm (changes requested):
I'd just maybe ask you to split the PR in two commits. One for the new csrs accessors the other for the interrupt handle fix.

filippofontana:
Ok, sure!

josecm (approved):

PR #212, opened 2025-04-14
AIA ipi fix

filippofontana:
This commits adds back a change (generic irqc function for sending IPIs in RISC-V) which probably got lost during the integration of AIA inside the main codebase.


josecm (changes requested):
@filippofontana we have updated the internal IPI API and the `ipi_id` parameter not longer exists. I have rebased and fixed and the issue and placed it [here](https://github.com/bao-project/bao-hypervisor/tree/aia-ipi-fix). However, I don't have permissions over the fork's branch to do it.  If you can please either add the “Allow edits from maintainers” to the branch in the fork, or update the fork's branch with the changes on the fix I've point to.

Thanks, and sorry for the (huge) delay

src/arch/riscv/irqc/aia/inc/irqc.h
```diff
@@ -64,13 +64,12 @@ static inline irqid_t irqc_reserve(irqid_t pintp_id)
...
 
-static inline void irqc_send_ipi(cpuid_t target_cpu, irqid_t ipi_id)
+static inline void irqc_send_ipi(cpuid_t target_cpu)
 {
 #if (IRQC == APLIC)
-    UNUSED_ARG(ipi_id);
     sbi_send_ipi(1UL << target_cpu, 0);
 #elif (IRQC == AIA)
-    imsic_send_msi(target_cpu, ipi_id);
+    imsic_send_msi(target_cpu, interrupts_ipi_id);
```
filippofontana: @josecm I think the cleanest approach is to remove this variable from here (it gets defined while compile `core/interrupts.c`, but `irqc.h` is also included during config generation (this causes the build error) and to put it directly inside `imsic_send_msi`, like this:

```diff
M src/arch/riscv/irqc/aia/imsic.c
@@ -80,9 +80,9 @@ void imsic_inject_pend(size_t guest_file, irqid_t intp_id)
     csrs_vsireg_clear(1UL << imsic_eie_bit(intp_id));
 }
 
-void imsic_send_msi(cpuid_t target_cpu, irqid_t ipi_id)
+void imsic_send_msi(cpuid_t target_cpu)
 {
-    imsic[target_cpu]->s_file.seteipnum_le = ipi_id;
+    imsic[target_cpu]->s_file.seteipnum_le = interrupts_ipi_id;
 }
```

In this way we can avoid clunky inclusion mechanism. What do you think?

If this sounds good to you I will prepare a new commit for this.

josecm: My mistake, I built with the modifications but stupidly forgot to make the IRQC=AIA. We need to make AIA the default interrupt controller for RISC-V. 

This patch looks good!

filippofontana: In the end I decided to keep everything in the same commit, let me know if you prefer to separate some parts of it.

josecm: No need! Thanks!

josecm (approved):

PR #216, opened 2025-05-22
ref(vmpu): Standardize vMPU driver and update armv8-r MPU

miguelafsilva5:
## PR Description

This PR aims to stantardize the vMPU driver and to the some extend the MPU API. This was needed to support different MPU implementations.
A major goal of this PR is remove the complexity from the MPU driver and move it to the vMPU.

In order to comply with these modifications to the vMPU, the MPU driver for armv8-r was also updated

PR #218, opened 2025-06-06
feat(non-unified-mem): introduce support for non unified memory platforms

danielRep:
## PR Description

This PR builds on top of PR #216 and adds support for platforms with non-unified memory architectures—i.e., platforms where code (e.g., flash) and data (e.g., SRAM) reside in separate memory regions. Previously, Bao only supported platforms with unified DRAM-based memory.
To support this architectural change, the following modifications were made:

- Refactored Bao’s memory reservation logic to handle the distinction between code and data memory.
- Removed the `load_addr` argument from function calls within `mem.c`, as it is now a constant global variable. Passing it explicitly no longer made sense in the context of non-unified memory platforms.
- Added memory permissions classification to reflect platform's memory restrictions (e.g., flash is read-only)
- Refactored the linkerscript to add the concept of VMA (Virtual Memory Address) and LMA (Load Memory Address)

### Refactored LinkerScript 

The following table details the difference between the memory layout of unified vs non-unified platforms. Unified platforms keep the same layout as expected while non-unified platforms split addresses between VMA and LMA addresses to cope with the typical separation of code and data memory architecture. 

#####  fvp-r-aarch32

| **Sections**            | **Address (bottom)** | **Linker Symbols**                 |
|-------------------------|----------------------|------------------------------------|
|                         | (top) 0x0001fe40     | _image_end; _dmem_beg              |
| .bss                    | 0x0001cec0           | _image_noload_start; _vm_image_end |
| .vm_images              | 0x0000cec0           | _image_load_end; _vm_image_start   |
| .ipi_cpumsg_handlers_id | ...                  |                                    |
| .ipi_cpumsg_handlers    | ...                  |                                    |
| .data                   | 0x0000c880           |                                    |
| .rodata                 | 0x0000ba00           |                                    |
| .text                   | 0x00000300           |                                    |
| .boot                   | 0x00000000           | _image_start                       |

##### qemu-aarch64-virt 

| **Sections**                | **Addresses (bottom)** | **Linker Symbols**                     |
|-------------------------|--------------------|------------------------------------|
|                         | 0xfd800002c000     | _image_noload_start; _vm_image_end |
|                         | 0xfd8000020000     | _image_end; _dmem_beg              |
| .glb_page_tables        | 0xfd800001c000     |                                    |
| .bss                    | 0xfd8000013000     |                                    |
| .vm_images              | 0xfd8000013000     | _image_load_end; _vm_image_start   |
| .ipi_cpumsg_handlers_id | ...                |                                    |
| .ipi_cpumsg_handlers    | ...                |                                    |
| .data                   | 0xfd8000012000     |                                    |
| .rodata                 | 0xfd8000010000     |                                    |
| .text                   | 0xfd8000000800     |                                    |
| .boot                   | 0xfd8000000000     | _image_start                       |

##### tricore tc49

| Sections                    | Addresses (bottom) | Linker Symbols                   |
|-----------------------------|--------------------|----------------------------------|
|                             | 0x90008e40         | _image_end; _dmem_beg            |
| .bss                        | 0x900014c0         | _image_noload_start; _bss_start  |
|                             | 0x800172c0         | _vm_image_end                    |
| .vm_images                  | 0x800130c0         | _image_load_end; _vm_image_start |
| .ipi_cpumsg_handlers_id     | ...                |                                  |
| .ipi_cpumsg_handlers        | ...                |                                  |
| .data                       | 0x90000000         | _data_vma_start                  |
| .data : AT(_data_lma_start) | 0x80011c00         | _data_lma_start                  |
| .rodata                     | 0x80010f80         |                                  |
| .text                       | 0x80000300         |                                  |
| .boot                       | 0x80000000         | _image_start                     |


src/core/mem.c
```diff
@@ -326,15 +399,21 @@ static bool mem_create_ppools(struct mem_region* root_mem_region)
...
     struct mem_region* root_mem_region = NULL;
     for (size_t i = 0; i < platform.region_num; i++) {
         struct mem_region* region = &(platform.regions[i]);
-        bool is_in_rgn = range_in_range(load_addr, image_size, region->base, region->size);
+        bool is_in_rgn;
+        if (!DEFINED(MEM_NON_UNIFIED)) {
+            is_in_rgn = range_in_range(_img_addr, root_mem_size, region->base, region->size);
+        } else {
+            is_in_rgn = range_in_range(_data_addr, root_mem_size, region->base, region->size);
+        }
```
DavidMCerdeira: invert the branches of the if for improved readability

``` c
if (DEFINED(MEM_NON_UNIFIED)) {
         is_in_rgn = range_in_range(_data_addr, root_mem_size, region->base, region->size);
} else {
         is_in_rgn = range_in_range(_img_addr, root_mem_size, region->base, region->size);
}
```

src/core/mem.c
```diff
@@ -14,10 +14,37 @@
...
 
 struct list page_pool_list;
 
+static size_t calc_root_mem_size(void)
+{
+    if (!DEFINED(MEM_NON_UNIFIED)) {
+        return (size_t)(&_image_end - &_image_start);
+    } else {
+        return (size_t)(&_image_end - &_data_vma_start);
+    }
```
DavidMCerdeira: invert the if branches for improved readability

src/core/cpu.c
```diff
@@ -31,13 +31,15 @@ static size_t ipi_cpumsg_handler_num;
 
 struct cpuif cpu_interfaces[PLAT_CPU_NUM];
 
-void cpu_init(cpuid_t cpu_id, paddr_t load_addr)
+extern uint32_t _img_addr;
```
DavidMCerdeira: This should be const

src/arch/armv8/aarch32/boot.S
```diff
@@ -23,6 +23,10 @@
  */
 _barrier: .4byte 0
 
+.balign 4
+.global _img_addr
```
DavidMCerdeira: what's the reason for starting the symbol name with '_' ?

src/linker.ld
```diff
@@ -30,9 +30,15 @@ SECTIONS
...
-	.data : {
+	. = ALIGN(PAGE_SIZE); /* start RW sections in separate page */
+	_data_lma_start = .;
+
+#ifdef MEM_NON_UNIFIED
+	. = PLAT_DATA_MEM;	/* Changed location counter to VMA address */
+	_data_vma_start = .;
+#endif
+
+	.data : AT (_data_lma_start){
```
DavidMCerdeira: I'm finding it a bit hard to understand this modification
Can you elaborate the commit message to make this easier to grasp?

src/core/mem.c
danielRep: Done. Follow the same for other similar branches.

src/core/mem.c
danielRep: Done.

src/core/cpu.c
danielRep: Done

src/arch/armv8/aarch32/boot.S
danielRep: Changed to the normal format `img_addr`

src/linker.ld
danielRep: Done.

DavidMCerdeira:
In the PR description, alongside the other modifications, I also suggest adding that it is now possible to setup memory permissions to reflect platform's memory restrictions (e.g., flash is read-only)

scripts/platform_defs_gen.c
```diff
@@ -14,6 +14,15 @@ int main() {
...
     printf("#define PLAT_CPU_NUM (%ld)\n", platform.cpu_num);
     printf("#define PLAT_BASE_ADDR (0x%lx)\n", platform.regions[0].base);
+
+    for(size_t i = 1; i < platform.region_num; i++)
+    {
+        if(platform.regions[i].perms == RWX)
+        {
+            printf("#define PLAT_DATA_MEM (0x%lx)\n", platform.regions[i].base);
+            break;
+        }
```
DavidMCerdeira: I believe we should make it explicit that the first "data" memory region is the one Bao will use for RAM
We should add a comment here to explain this, and in the future possibly in the "affected" platforms' platform description as well

src/linker.ld
```diff
@@ -51,15 +57,27 @@ SECTIONS
...
+    }
+
+#ifdef MEM_NON_UNIFIED
+    /* Restore the location counter (VMA) */
+    . = _vma_before_vm_images;
+#endif
+
 	. = ALIGN(PAGE_SIZE);
-	_vm_image_end = .;
+	_vm_image_end = _vm_image_start + SIZEOF(.vm_images);
```
DavidMCerdeira: Is SIZEOF page aligned?

scripts/platform_defs_gen.c
danielRep: Added the required comment.

src/linker.ld
danielRep: Forced the alignment.

josecm:
@danielRep what is holding this for review?

src/core/inc/mem.h
```diff
@@ -15,6 +15,11 @@
 
 #ifndef __ASSEMBLER__
 
+enum MEM_PERMISSIONS {
+    RWX,
+    RX,
```
josecm: Could we make these variants MEM_RWX, etc?

src/core/inc/mem.h
```diff
@@ -15,6 +15,11 @@
 
 #ifndef __ASSEMBLER__
 
+enum MEM_PERMISSIONS {
+    RWX,
```
josecm: The platform description needs to assume this variant is zero so we don't have to specify it for all regions. Let's make it explicit and
```suggestion
    RWX = 0, // This variant always has to be zero and therefore the first one
```

danielRep:
> @danielRep can we rebase this on ref/vmpu?

Done.

src/linker.ld
```diff
@@ -30,9 +30,19 @@ SECTIONS
...
+	_data_lma_start = .;
+
+#ifdef MEM_NON_UNIFIED
+	. = PLAT_DATA_MEM;	/* Changed location counter to VMA address */
+	_data_vma_start = .;
+#else
+	/* Dummy initialization to avoid the compiler triggering an undefined reference to
+	 _data_vma_start with -O0, so we need to provide it even if MEM_NON_UNIFIED is not defined. */
+	_data_vma_start = _data_lma_start;
+#endif
```
josecm: Can't we just do:


```suggestion
#ifdef MEM_NON_UNIFIED
	. = PLAT_DATA_MEM;	/* Changed location counter to VMA address */
#endif

	_data_vma_start = .;
```

I think this would be enough since if the location counter is not updated to PLAT_DATA_MEM, it will be equal do _data_lma_start.

src/core/mem.c
```diff
@@ -430,9 +509,11 @@ void mem_init(paddr_t load_addr)
 
     cpu_sync_and_clear_msgs(&cpu_glb_sync);
 
+#ifdef MEM_PROT_MMU
```
josecm: Why do we need to have this conditional compilation now?

src/core/mem.c
```diff
@@ -327,15 +400,21 @@ static bool mem_create_ppools(struct mem_region* root_mem_region)
...
     struct mem_region* root_mem_region = NULL;
     for (size_t i = 0; i < platform.region_num; i++) {
         struct mem_region* region = &(platform.regions[i]);
-        bool is_in_rgn = range_in_range(load_addr, image_size, region->base, region->size);
+        bool is_in_rgn;
+        if (DEFINED(MEM_NON_UNIFIED)) {
+            is_in_rgn = range_in_range(data_addr, root_mem_size, region->base, region->size);
+        } else {
+            is_in_rgn = range_in_range(img_addr, root_mem_size, region->base, region->size);
+        }
```
josecm:
```suggestion
        vaddr_t root_base_addr;
        if (DEFINED(MEM_NON_UNIFIED)) {
            root_base_addr = data_addr;
        } else {
            root_base_addr = img_addr;
        }
        is_in_rgn = range_in_range(img_addr, root_mem_size, region->base, region->size);
```

src/core/mem.c
```diff
@@ -258,6 +325,12 @@ static bool mem_reserve_physical_memory(struct page_pool* pool)
         return false;
     }
 
+    if (DEFINED(MEM_NON_UNIFIED)) {
+        if (!pp_reserve_hyp_image_load(pool)) {
+            return false;
+        }
+    }
```
josecm: Just so I fully get why this is here. We need to try to reserve the hypervisor image for every pool since it is not guaranteed that the image is in the root pool for non-unified mem plaforms. Is that correct?

src/core/mem.c
```diff
@@ -158,39 +181,83 @@ static bool root_pool_set_up_bitmap(paddr_t load_addr, struct page_pool* root_po
...
+{
+    if (DEFINED(MEM_NON_UNIFIED)) {
+        return pp_reserve_hyp_data(root_pool);
+
+    } else {
+        bool hyp_image_load_mem = pp_reserve_hyp_image_load(root_pool);
+        bool hyp_image_noload_mem = pp_reserve_hyp_image_noload(root_pool);
+        bool cpus_mem = pp_reserve_cpus(root_pool);
+        return hyp_image_load_mem && hyp_image_noload_mem && cpus_mem;
+    }
```
josecm: Shouldnt we reserve the cpus in the mem_non_unified case also?

src/core/mem.c
```diff
@@ -158,39 +181,83 @@ static bool root_pool_set_up_bitmap(paddr_t load_addr, struct page_pool* root_po
...
+    paddr_t data_base_addr = (paddr_t)&_data_vma_start;
+    paddr_t cpu_base_addr = data_base_addr + data_size;
+
+    struct ppages data_ppages = mem_ppages_get(data_base_addr, NUM_PAGES(data_size));
     struct ppages cpu_ppages = mem_ppages_get(cpu_base_addr, NUM_PAGES(cpu_size));
 
-    bool image_load_reserved = mem_reserve_ppool_ppages(root_pool, &images_load_ppages);
-    bool image_noload_reserved = mem_reserve_ppool_ppages(root_pool, &images_noload_ppages);
+    bool data_reserved = mem_reserve_ppool_ppages(root_pool, &data_ppages);
     bool cpu_reserved = mem_reserve_ppool_ppages(root_pool, &cpu_ppages);
```
josecm: Can't we take the reservation of cpus from here and call `pp_reserve_cpus` ?

src/core/mem.c
```diff
@@ -158,39 +181,83 @@ static bool root_pool_set_up_bitmap(paddr_t load_addr, struct page_pool* root_po
...
+
+    return mem_reserve_ppool_ppages(pool, &cpu_ppages);
+}
+
+static bool pp_reserve_hyp_data(struct page_pool* root_pool)
+{
+    size_t data_size = (size_t)(&_image_end - &_data_vma_start);
+    size_t cpu_size = platform.cpu_num * mem_cpu_boot_alloc_size();
+    paddr_t data_base_addr = (paddr_t)&_data_vma_start;
+    paddr_t cpu_base_addr = data_base_addr + data_size;
```
josecm: How is this guaranteed to be aligned to PAGE_SIZE?

src/core/mem.c
```diff
@@ -158,39 +181,83 @@ static bool root_pool_set_up_bitmap(paddr_t load_addr, struct page_pool* root_po
...
+    size_t cpu_size = platform.cpu_num * mem_cpu_boot_alloc_size();
+    paddr_t image_noload_addr = img_addr + image_load_size + vm_image_size;
+    paddr_t cpu_base_addr = image_noload_addr + image_noload_size;
+
+    struct ppages cpu_ppages = mem_ppages_get(cpu_base_addr, NUM_PAGES(cpu_size));
+
+    return mem_reserve_ppool_ppages(pool, &cpu_ppages);
+}
+
+static bool pp_reserve_hyp_data(struct page_pool* root_pool)
```
josecm: Why are we calling the argument `root_pool` here and not in the previous functions?

src/core/inc/bao.h
```diff
@@ -22,7 +22,12 @@
     console_printk("BAO ERROR: " __VA_ARGS__); \
     while (true) { };
 
-void init(cpuid_t cpu_id, paddr_t load_addr);
+void init(cpuid_t cpu_id);
+
+/* The address where the Bao image is loaded in memory */
+extern const uintptr_t img_addr;
+/* The address where the data section is loaded in memory */
+extern const uintptr_t data_addr;
```
josecm: Probably it would make more sense to have this in mem.h

scripts/platform_defs_gen.c
```diff
@@ -14,6 +14,20 @@ int main() {
...
+         * Selects the first memory region with RWX (read, write, execute) permissions, and defines
+         * it as PLAT_DATA_MEM. This region is considered the main data memory that Bao will use
+         * for its own purposes.
+         */
+        if(platform.regions[i].perms == RWX)
+        {
+            printf("#define PLAT_DATA_MEM (0x%lx)\n", platform.regions[i].base);
+            break;
+        }
+    }
```
josecm: I am thinking we could also allow a user to define the the data base in the hyp section of the configuration. Otherwise it is too rigid to make the user use that region for hypervisor data in all configurations

josecm (changes requested):

src/core/mem.c
danielRep: Yes, correct.

src/core/inc/mem.h
danielRep: Fixed in f13b28ff

src/core/inc/mem.h
danielRep: Fixed in f13b28ff

src/linker.ld
danielRep: Fixed in 8d57c1c0

src/core/mem.c
danielRep: Fixed in 96aee16c

src/core/mem.c
danielRep: Fixed in 2dc87ae2

src/core/mem.c
danielRep: Fixed in 662d2977

src/core/mem.c
danielRep: Fixed in 662d2977

src/core/mem.c
danielRep: Fixed in f33f6133

src/core/inc/bao.h
danielRep: Fixed in 60ac1ad6

scripts/platform_defs_gen.c
danielRep: Fixed in 187e1ca9

src/core/inc/mem.h
```diff
@@ -89,6 +95,11 @@ bool mem_translate(struct addr_space* as, vaddr_t va, paddr_t* pa);
 
 extern struct list page_pool_list;
 
-#endif /* |__ASSEMBLER__ */
+/* The address where the Bao image is loaded in memory */
+extern uintptr_t img_addr;
+/* The address where the data section is loaded in memory */
+extern uintptr_t data_addr;
```
josecm: A bit pedantic, but can we use `vaddr_t`?
```suggestion
extern vaddr_t img_addr;
/* The address where the data section is loaded in memory */
extern vaddr_t data_addr;
```

src/core/mem.c
josecm: I don't think you got my point here. I'd prefer that all necessary fuinctions would identify the pool as the root pool. Even better would be:

```suggestion
static bool pp_root_reserve_hyp_data(struct page_pool* root_pool)
```

And the same in other hypervisor memory reservation functions that implicitly target the root pool.

src/core/mem.c
josecm: @danielRep did you miss this one?

src/core/inc/mem.h
danielRep: Done 5b2aae0ff46486bd55446c57b4f6dfee17158488

src/core/mem.c
danielRep: Done 04f6748a

src/core/mem.c
danielRep: Done adb3b369b18a760f1ace2d231ff6b5d816020821

josecm:
@danielRep please rebase on main and clean up commit history before proceeding

josecm:
@DavidMCerdeira @miguelafsilva5 please give another pass when possible

miguelafsilva5:
> @DavidMCerdeira @miguelafsilva5 please give another pass when possible

@josecm I've given it another pass and rebased it to TC4. Everything is working and seems OK

miguelafsilva5 (approved):

PR #220, opened 2025-06-09
fix(objpool): mark objpool func as static to avoid missing symbol at O0

danielRep:
## PR Description

Small PR to mark `objpool_get_by_id` as `static inline` to ensure the function is always available in translation units that use it. This prevents linker errors at optimization level 0, where the compiler will throw an undefined reference error regarding that symbol.


DavidMCerdeira (approved):

DavidMCerdeira:
Should we consider adding optimization flags to the code building matrix?

danielRep:
> Should we consider adding optimization flags to the code building matrix?

I'm inclined to agree

PR #222, opened 2025-07-08
ref(ipi): Update IPI API to support additional architectures

miguelafsilva5:
## PR Description

This PR introduces small changes to the current IPI API moving the interrupt_ipi_id dependency to the arch implementation. 
It also introduces an additional interface _interrupts_arch_ipi_init_ to support architectures that have different IPI mechanisms (i.e TC4).

Additionally, the current supported architectures were updated to match the new interfaces

### DIFF
--- a/src/arch/armv8/interrupts.c
+++ b/src/arch/armv8/interrupts.c
@@ -22,10 +22,11 @@ void interrupts_arch_init()
     gic_init();
 }
 
-void interrupts_arch_ipi_send(cpuid_t target_cpu, irqid_t ipi_id)
+void interrupts_arch_ipi_send(cpuid_t target_cpu)
 {
-    if (ipi_id < GIC_MAX_SGIS) {
-        gic_send_sgi(target_cpu, ipi_id);
+    // This check may be removed? Might be an assert or something
+    if (interrupts_ipi_id < GIC_MAX_SGIS) {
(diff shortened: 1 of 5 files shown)


src/arch/armv8/interrupts.c
```diff
@@ -22,10 +22,11 @@ void interrupts_arch_init()
     gic_init();
 }
 
-void interrupts_arch_ipi_send(cpuid_t target_cpu, irqid_t ipi_id)
+void interrupts_arch_ipi_send(cpuid_t target_cpu)
 {
-    if (ipi_id < GIC_MAX_SGIS) {
-        gic_send_sgi(target_cpu, ipi_id);
+    // This check may be removed? Might be an assert or something
+    if (interrupts_ipi_id < GIC_MAX_SGIS) {
```
DavidMCerdeira: The interrupts_ipi_id is now global state and not a parameter, thus, as the code comment implies, this check could potentially be removed.
I think we should remove this and make the check upon interrupts_ipi_id initialization only.
@josecm what do you think?

josecm: Yes. completely agree. @miguelafsilva5 please follow through with this suggestion

miguelafsilva5: Done

DavidMCerdeira (approved):

PR #223, opened 2025-07-11
ref: Replace hardcoded booleans with named macros

DavidMCerdeira:
This commit updates mainly memory and bitmap operations to use named macros like MEM_DONT_FREE_PAGES, MEM_ALIGN_PPAGES, and BITMAP_DONT_SET instead of raw boolean values on respective functions.

### DIFF
--- a/src/arch/armv8/inc/arch/vgic.h
+++ b/src/arch/armv8/inc/arch/vgic.h
@@ -14,6 +14,9 @@ struct vm;
 struct vcpu;
 struct vgic_dscrp;
 
+#define VGIC_GICR_ACCESS     (true)
+#define VGIC_NOT_GICR_ACCESS (false)
+
 /**
  * TODO: optimize the vgic_int struct's size
  */
--- a/src/core/inc/mem.h
+++ b/src/core/inc/mem.h
@@ -15,6 +15,12 @@
 
 #ifndef __ASSEMBLER__
 
+#define MEM_ALIGN_REQ       (true)
+#define MEM_ALIGN_NOT_REQ   (false)
+
+#define MEM_FREE_PAGES      (true)
+#define MEM_DONT_FREE_PAGES (false)
+
 enum MEM_PERMISSIONS {
     MEM_RWX = 0, // This variant always has to be zero and therefore the first one
     MEM_RX,
--- a/src/core/mpu/mem.c
+++ b/src/core/mpu/mem.c
@@ -13,6 +13,12 @@
 #include <config.h>
 #include <arch/mpu.h>
 
+#define MEM_BROADCAST      (true)
+#define MEM_DONT_BROADCAST (false)
+
+#define MEM_LOCKED         (true)
+#define MEM_NOT_LOCKED     (false)
+
 struct shared_region {
     enum AS_TYPE as_type;
     asid_t asid;
@@ -174,7 +180,7 @@ static void mem_init_boot_regions(void)
 #endif
         .as_sec = SEC_HYP_IMAGE,
     };
-    mem_map(&cpu()->as, &mpr, false, true);
+    mem_map(&cpu()->as, &mpr, MEM_DONT_BROADCAST, MEM_LOCKED);
 
     if (separate_noload_region) {
         mpr = (struct mp_region){
@@ -188,7 +194,7 @@ static void mem_init_boot_regions(void)
             .mem_flags = PTE_HYP_FLAGS,
             .as_sec = SEC_HYP_IMAGE,
         };
-        mem_map(&cpu()->as, &mpr, false, true);
+        mem_map(&cpu()->as, &mpr, MEM_DONT_BROADCAST, MEM_LOCKED);
     }
 
     mpr = (struct mp_region){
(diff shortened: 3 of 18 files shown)


josecm:
@DavidMCerdeira looks good so far. Please let me know when its ready for final review

DavidMCerdeira:
@josecm this is ready for review

danielRep (dismissed):
LGTM

src/arch/armv8/inc/arch/vgic.h
```diff
@@ -71,6 +71,9 @@ struct vgic_priv {
     struct vgic_int interrupts[GIC_CPU_PRIV];
 };
 
+#define VGIC_GICR_ACCESS    true
+#define VGIC_NO_GICR_ACCESS false
```
josecm: Shouldnt these macros have surrounding parentheses?

```suggestion
#define VGIC_GICR_ACCESS    (true)
#define VGIC_NO_GICR_ACCESS (false)
```

josecm: I'd prefer the phrasing:

```suggestion
#define VGIC_GICR_ACCESS    true
#define VGIC_NOT_GICR_ACCESS false
```

src/core/inc/mem.h
```diff
@@ -55,6 +55,12 @@ struct shmem {
     spinlock_t lock;
 };
 
+#define MEM_PPAGES_ALIGNED     true
+#define MEM_PPAGES_NOT_ALIGNED false
```
josecm:
```suggestion
#define MEM_PPAGES_ALIGNED_REQUIRED     true
#define MEM_PPAGES_ALIGNED_NOT_REQUIRED false
```

src/core/mpu/mem.c
```diff
@@ -13,6 +13,12 @@
 #include <config.h>
 #include <arch/mpu.h>
 
+#define MEM_BROADCAST      true
+#define MEM_DONT_BROADCAST false
+
+#define MEM_LOCKED         true
+#define MEM_UNLOCKED       false
```
josecm: Don't love the "unlocked" phrasing. To keep consistent with the other macros maybe `MEM_NOT_LOCKED`

src/arch/armv8/inc/arch/vgic.h
DavidMCerdeira: Does this make a difference? Or is it part of a coding guideline I'm not aware of?

src/core/inc/mem.h
DavidMCerdeira: I see your point! However the result seems to me an overly long macro name. 
How about `MEM_PP_ALIGN_OPT` (optional), `MEM_PP_ALIGN_REQ` (required)?

src/core/mpu/mem.c
DavidMCerdeira: `MEM_NOT_LOCKED` doesn't sound right me either.
Just realized "pinned" might be the better term for this actually. 
What do you think about `MEM_PINNED` and `MEM_NOT_PINNED`?

src/arch/armv8/inc/arch/vgic.h
josecm: It is good practice for expressions definitions in macros, including literals, to always be between parenthesis. We will include a "Parentheses shall be used to enclose every macro parameter." guideline in the documentation.

src/core/inc/mem.h
josecm: The latter seems ok for me.

src/core/mpu/mem.c
josecm: What about following the logic in other macros: `MEM_DONT_LOCK`?

DavidMCerdeira: I decided `MEM_NOT_LOCKED` was ok

src/arch/armv8/inc/arch/vgic.h
DavidMCerdeira: Done

DavidMCerdeira: Done

src/core/inc/mem.h
DavidMCerdeira: Done

josecm: Actually I was suggesting `MEM_PP_ALIGN_REQ` and `MEM_PP_ALIGN_NOT_REQ`.  Or just `MEM_ALIGN_REQ` and `MEM_ALIGN_NOT_REQ` to keep them shorter?

danielRep: I'd prefer `MEM_ALIGN_REQ` and `MEM_ALIGN_NOT_REQ`

src/core/inc/mem.h
```diff
@@ -15,6 +15,12 @@
 
 #ifndef __ASSEMBLER__
 
+#define MEM_PP_ALIGN_REQ    (true)
+#define MEM_PP_ALIGN_OPT    (false)
```
danielRep:
```suggestion
#define MEM_ALIGN_REQ       (true)
#define MEM_ALIGN_NOT_REQ   (false)
```

src/core/inc/mem.h
DavidMCerdeira: I agree
Done

src/core/inc/mem.h
DavidMCerdeira: Done

PR #224, opened 2025-07-14
fix(rv32): use ul for sbi_send_ipi call

josecm:
Compilation is failing for rv32. I guess we need to add this variant to CI.


DavidMCerdeira:
Ping @ninolomata

PR #230, opened 2025-08-04
fix(armv8/aarch32): restore/save extra lr on vm entry/exit

josecm:
Hyp mod uses lr from user mode. This means we were only saving/restoring the lr from this mode. But in addition to this lr we need to restore the lr of the mode we just preemted (e.g., lr_irq) so that we can inspect and manipulate the state of the mode that was trapped.

### DIFF
--- a/src/arch/armv8/aarch32/exceptions.S
+++ b/src/arch/armv8/aarch32/exceptions.S
@@ -46,14 +46,14 @@
 .macro VM_EXIT
 
     push {r14}
-    SAVE_SP
+    SAVE_LR_SP
     push {r0-r12}
     SAVE_ELR_SPSR
 
 .endm
 
 
-.macro SAVE_SP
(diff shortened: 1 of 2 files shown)


danielRep (changes requested):
Looks good to me.

src/arch/armv8/aarch32/exceptions.S
```diff
@@ -46,14 +46,14 @@
...
     SAVE_ELR_SPSR
 
 .endm
 
 
-.macro SAVE_SP
+.macro SAVE_LR_SP
 
     mrs r14, SPSR_hyp
     and r14, r14, #0x1F
```
danielRep, line 59: Please use a macro for this mask.

josecm:
I think there is a better way to do this. I'll send another PR.

PR #233, opened 2025-08-07
Add missing spinlock initializations

malejo97:
This PR adds missing initialization of spinlocks across the source code.
Uninitialized control structures are a potential cause of random behavior issues.

The PR also deletes the unused `spinlock_init` function, as spinlock initialization in Bao is done via the `SPINLOCK_INITVAL` macro.

### DIFF
--- a/src/arch/armv8/aarch32/inc/arch/spinlock.h
+++ b/src/arch/armv8/aarch32/inc/arch/spinlock.h
@@ -15,12 +15,6 @@ typedef struct {
 
 static const spinlock_t SPINLOCK_INITVAL = { 0, 0 };
 
-static inline void spinlock_init(spinlock_t* lock)
-{
-    lock->ticket = 0;
-    lock->next = 0;
-}
-
 /**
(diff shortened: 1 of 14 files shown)


josecm:
Nice catch @malejo97 ! 

I just realized we are not using `spinlock_init` anywhere and instead prefer to use that init value macro. Could you also delete the `spinlock_init` functions and add this to the PR? 

Thanks!

malejo97:
> Nice catch @malejo97 !
> 
> I just realized we are not using `spinlock_init` anywhere and instead prefer to use that init value macro. Could you also delete the `spinlock_init` functions and add this to the PR?
> 
> Thanks!

Done!

josecm:
Should we check all locks are being initialized?

malejo97:
@josecm I've found a bunch of other locks without initialization.
I've squashed all added initializations into a single commit, and left the deleting of `spinlock_init` in another one.

josecm (approved):

PR #235, opened 2025-08-09
fix(mem): don't adjust pp_alloc start when alignment is unecessary

josecm:
We were adjusting curr event if aligned was set to false which resulted in skipping available pages even if alignment was not required, creating fragementation. Worse, it precluded allocating contigous regions which is of the upmost importance to allow merging (and consequent saving) of mpu entries.

### DIFF
--- a/src/core/mem.c
+++ b/src/core/mem.c
@@ -34,8 +34,15 @@ bool pp_alloc(struct page_pool* pool, size_t num_pages, bool aligned, struct ppa
     /**
      * If we need a contigous segment aligned to its size, lets start at an already aligned index.
      */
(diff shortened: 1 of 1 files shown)


DavidMCerdeira (commented):
The commit message as minor spelling mistakes. Here's a fix, slightly tweaked:

> We were adjusting `curr` even if `aligned` was set to false which resulted in skipping available pages even if alignment was not required, creating fragmentation. Worse, it precluded allocating contiguous regions which is of the upmost importance to allow merging (and consequent saving) of mpu entries.

Otherwise lgtm

danielRep (approved):
lgtm

DavidMCerdeira (approved):

PR #238, opened 2025-08-19
feat/static-bitmaps: change bitmaps to be statically allocated

danielRep:
This pull request refactors the memory bitmap allocation logic to use a preallocated bitmap pool. We introduce a global bitmap pool and update relevant functions to allocate bitmaps from this pool instead of allocating memory dynamically. With this PR we optimize the use of MPU regions in MMU-less platforms, as the global bitmap is allocated in the .bss section as part of the hyp image.
To calculate the size of the global bitmap pool based on platform regions, we added `PLAT_BITMAP_POOL_SIZE` in `scripts/platform_defs_gen.c`.

### DIFF
--- a/scripts/platform_defs_gen.c
+++ b/scripts/platform_defs_gen.c
@@ -24,7 +24,7 @@ int main() {
 
         reg_size = platform.regions[i].size;
 
-        bitmap_array_size += reg_size / (8 * PAGE_SIZE) + ((reg_size % (8 * PAGE_SIZE) != 0) ? 1 : 0);
+        bitmap_array_size += BITMAP_SIZE_IN_BYTES(NUM_PAGES(reg_size));
     }
 
     printf("#define PLAT_BITMAP_POOL_SIZE (0x%lx)\n", bitmap_array_size);
--- a/scripts/platform_defs_gen.c
+++ b/scripts/platform_defs_gen.c
@@ -5,6 +5,7 @@
 
 #include <stdio.h>
 #include <platform.h>
+#include <bao.h>
 
 __attribute__((weak)) void arch_platform_defs(void){
     return;
@@ -12,8 +13,22 @@ __attribute__((weak)) void arch_platform_defs(void){
 
 int main() {
 
+    size_t bitmap_array_size = 0;
+
     printf("#define PLAT_CPU_NUM (%ld)\n", platform.cpu_num);
     printf("#define PLAT_BASE_ADDR (0x%lx)\n", platform.regions[0].base);
+
+    for(size_t i = 0; i < platform.region_num; i++)
+    {
+        size_t reg_size;
+
+        reg_size = platform.regions[i].size;
+
+        bitmap_array_size += reg_size / (8 * PAGE_SIZE) + ((reg_size % (8 * PAGE_SIZE) != 0) ? 1 : 0);
+    }
+
+    printf("#define PLAT_BITMAP_POOL_SIZE (0x%lx)\n", bitmap_array_size);
+
     if (platform.cpu_master_fixed) {
         printf("#define CPU_MASTER_FIXED (%ld)\n", platform.cpu_master);
     }
--- a/src/core/mem.c
+++ b/src/core/mem.c
@@ -18,6 +18,28 @@ extern uint8_t _image_start, _image_load_end, _image_end, _vm_image_start, _vm_i
 
 struct list page_pool_list;
 
+static bool pp_bitmap_alloc(size_t pool_num_pages, bitmap_t** bitmap)
+{
+    static uint8_t bitmap_pool[PLAT_BITMAP_POOL_SIZE];
+    static spinlock_t bitmap_lock = SPINLOCK_INITVAL;
+    static size_t last_index = 0;
+    bool allocated = false;
+
+    spin_lock(&bitmap_lock);
+
+    if (((PLAT_BITMAP_POOL_SIZE - last_index) * 8) >= pool_num_pages) {
+        size_t prev_last_index = last_index;
+
+        last_index += BITMAP_SIZE_IN_BYTES(pool_num_pages);
+        allocated = true;
+        *bitmap = (bitmap_t*)&bitmap_pool[prev_last_index];
+    }
+
+    spin_unlock(&bitmap_lock);
+
+    return allocated;
+}
+
 bool pp_alloc(struct page_pool* pool, size_t num_pages, bool aligned, struct ppages* ppages)
 {
     ppages->colors = 0;
@@ -145,25 +167,9 @@ void* mem_alloc_page(size_t num_pages, enum AS_SEC sec, bool phys_aligned)
 
 static bool root_pool_set_up_bitmap(paddr_t load_addr, struct page_pool* root_pool)
 {
-    size_t image_size = (size_t)(&_image_end - &_image_start);
-    size_t vm_image_size = (size_t)(&_vm_image_end - &_vm_image_start);
-    size_t cpu_size = platform.cpu_num * mem_cpu_boot_alloc_size();
+    UNUSED_ARG(load_addr);
 
-    size_t bitmap_size = root_pool->num_pages / 8 + ((root_pool->num_pages % 8 != 0) ? 1 : 0);
-    size_t bitmap_num_pages = NUM_PAGES(bitmap_size);
-
-    if (root_pool->num_pages <= bitmap_num_pages) {
-        return false;
-    }
-    size_t bitmap_base = load_addr + image_size + vm_image_size + cpu_size;
-
-    struct ppages bitmap_pp = mem_ppages_get(bitmap_base, bitmap_num_pages);
-    bitmap_t* root_bitmap = (bitmap_t*)mem_alloc_map(&cpu()->as, SEC_HYP_GLOBAL, &bitmap_pp,
-        INVALID_VA, bitmap_num_pages, PTE_HYP_FLAGS);
-    root_pool->bitmap = root_bitmap;
-    memset((void*)root_pool->bitmap, 0, bitmap_num_pages * PAGE_SIZE);
-
-    return mem_reserve_ppool_ppages(root_pool, &bitmap_pp);
+    return pp_bitmap_alloc(root_pool->num_pages, &root_pool->bitmap);
 }
 
 static bool pp_root_reserve_hyp_mem(paddr_t load_addr, struct page_pool* root_pool)
@@ -206,39 +212,25 @@ static bool pp_root_init(paddr_t load_addr, struct mem_region* root_region)
     return true;
 }
 
-static void pp_init(struct page_pool* pool, paddr_t base, size_t size)
+static bool pp_init(struct page_pool* pool, paddr_t base, size_t size)
 {
-    struct ppages pages;
-
     if (pool == NULL) {
-        return;
+        return false;
     }
 
     memset((void*)pool, 0, sizeof(struct page_pool));
     pool->base = ALIGN(base, PAGE_SIZE);
     pool->num_pages = NUM_PAGES(size);
-    size_t bitmap_size = pool->num_pages / 8 + !!(pool->num_pages % 8 != 0);
-    size_t bitmap_num_pages = NUM_PAGES(bitmap_size);
 
-    if (size <= bitmap_num_pages) {
-        return;
+    if (!pp_bitmap_alloc(pool->num_pages, &pool->bitmap)) {
+        return false;
     }
-
-    pages = mem_alloc_ppages(cpu()->as.colors, bitmap_num_pages, false);
-    if (pages.num_pages != bitmap_num_pages) {
-        return;
-    }
-
-    if ((pool->bitmap = (bitmap_t*)mem_alloc_map(&cpu()->as, SEC_HYP_GLOBAL, &pages, INVALID_VA,
-             bitmap_size, PTE_HYP_FLAGS)) == NULL) {
-        return;
-    }
-
-    memset((void*)pool->bitmap, 0, bitmap_num_pages * PAGE_SIZE);
 
     pool->last = 0;
     pool->free = pool->num_pages;
     pool->lock = SPINLOCK_INITVAL;
+
+    return true;
 }
 
 static bool mem_vm_img_in_phys_rgn(struct vm_config* vm_config)
@@ -324,7 +316,9 @@ static bool mem_create_ppools(struct mem_region* root_mem_region)
             struct mem_region* reg = &platform.regions[i];
             struct page_pool* pool = &reg->page_pool;
             if (pool != NULL) {
-                pp_init(pool, reg->base, reg->size);
+                if (!pp_init(pool, reg->base, reg->size)) {
+                    return false;
+                }
                 if (!mem_reserve_physical_memory(pool)) {
                     return false;
                 }
--- a/src/core/objpool.c
+++ b/src/core/objpool.c
@@ -9,7 +9,7 @@
 void objpool_init(struct objpool* objpool)
 {
     memset(objpool->pool, 0, objpool->objsize * objpool->num);
-    memset(objpool->bitmap, 0, BITMAP_SIZE(objpool->num));
+    memset(objpool->bitmap, 0, BITMAP_SIZE_IN_BYTES(objpool->num));
     objpool->lock = SPINLOCK_INITVAL;
 }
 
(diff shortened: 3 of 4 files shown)


src/core/mem.c
```diff
@@ -17,6 +17,30 @@
 extern uint8_t _image_start, _image_load_end, _image_end, _vm_image_start, _vm_image_end;
 
 struct list page_pool_list;
+uint8_t bitmap_pool[PLAT_BITMAP_POOL_SIZE];
```
josecm: This should also be static

src/core/mem.c
```diff
@@ -17,6 +17,30 @@
...
+static spinlock_t bitmap_lock = SPINLOCK_INITVAL;
```
josecm: Both of these could static inside `mem_bitmap_pool_alloc`

src/core/mem.c
```diff
@@ -17,6 +17,30 @@
...
+    spin_lock(&bitmap_lock);
+
+    if (PLAT_BITMAP_POOL_SIZE - last_index * 8 * PAGE_SIZE > bitmap_size) {
+        size_t prev_last_index = last_index;
+
+        last_index += bitmap_size;
+        allocated = true;
+        *bitmap = (bitmap_t*)&bitmap_pool[prev_last_index];
+
+        return true;
```
josecm: If we have `allocated = true` we don't need to return here.

josecm: Actually, we can't! We need to unlock the lock

src/core/mem.c
```diff
@@ -145,26 +169,28 @@ void* mem_alloc_page(size_t num_pages, enum AS_SEC sec, bool phys_aligned)
...
+    UNUSED_ARG(load_addr);
+    bool bitmap_alloced = false;
 
     size_t bitmap_size =
         root_pool->size / (8 * PAGE_SIZE) + ((root_pool->size % (8 * PAGE_SIZE) != 0) ? 1 : 0);
     size_t bitmap_num_pages = NUM_PAGES(bitmap_size);
 
-    if (root_pool->size <= bitmap_size) {
-        return false;
+    if (root_pool->size <= bitmap_num_pages) {
```
josecm: Shouldn't this be:

```suggestion
    if (root_pool->size <= bitmap_size) {
```

src/core/mem.c
```diff
@@ -17,6 +17,30 @@
...
+static spinlock_t bitmap_lock = SPINLOCK_INITVAL;
+
+static bool mem_bitmap_pool_alloc(size_t bitmap_size, bitmap_t** bitmap)
+{
+    static size_t last_index = 0;
+    bool allocated = false;
+
+    spin_lock(&bitmap_lock);
+
+    if (PLAT_BITMAP_POOL_SIZE - last_index * 8 * PAGE_SIZE > bitmap_size) {
```
josecm: Why are we multiplying by 8*PAGE_SIZE? The argument is already the size of the bitmap in bytes, correct? Also the comparison should be greater or equal. For example, for a plat with a single pool the PLAT_BIMTAP_POOL will be (for example) 10 and the bitmap size is also 10: (10 - 0) > 10 is false and allocation will fail.  Shouldn't this be:

```suggestion
    if ((PLAT_BITMAP_POOL_SIZE - last_index) >= bitmap_size) {
```

src/core/mem.c
```diff
@@ -145,26 +169,28 @@ void* mem_alloc_page(size_t num_pages, enum AS_SEC sec, bool phys_aligned)
 
 static bool root_pool_set_up_bitmap(paddr_t load_addr, struct page_pool* root_pool)
 {
-    size_t image_size = (size_t)(&_image_end - &_image_start);
-    size_t vm_image_size = (size_t)(&_vm_image_end - &_vm_image_start);
-    size_t cpu_size = platform.cpu_num * mem_cpu_boot_alloc_size();
+    UNUSED_ARG(load_addr);
+    bool bitmap_alloced = false;
```
josecm:
```suggestion
    bool bitmap_allocated = false;
```

src/core/mem.c
```diff
@@ -145,26 +169,28 @@ void* mem_alloc_page(size_t num_pages, enum AS_SEC sec, bool phys_aligned)
...
 
     size_t bitmap_size =
         root_pool->size / (8 * PAGE_SIZE) + ((root_pool->size % (8 * PAGE_SIZE) != 0) ? 1 : 0);
     size_t bitmap_num_pages = NUM_PAGES(bitmap_size);
 
-    if (root_pool->size <= bitmap_size) {
-        return false;
+    if (root_pool->size <= bitmap_num_pages) {
+        return bitmap_alloced;
```
josecm: let's just return false. I don't think it is worth it to use bitmap_allocated in this function since there are so many return points with an obivous outcome.

src/core/mem.c
josecm: I'm confused, we are passing as an argument for bitmap_size "NUM_PAGES(root_pool->size)" which is the bitmap_size in bits. So is this check correct? Now I'm thinking it should be:


```suggestion
    if (((PLAT_BITMAP_POOL_SIZE - last_index) * 8) > bitmap_size) {
```

src/core/mem.c
```diff
@@ -17,6 +17,30 @@
...
+{
+    static size_t last_index = 0;
+    bool allocated = false;
+
+    spin_lock(&bitmap_lock);
+
+    if (PLAT_BITMAP_POOL_SIZE - last_index * 8 * PAGE_SIZE > bitmap_size) {
+        size_t prev_last_index = last_index;
+
+        last_index += bitmap_size;
```
josecm: And this shouldn't be:

```suggestion
        last_index += (bitmap_size/8);
```

josecm: Probably need to deal with remainder too.

src/core/mem.c
```diff
@@ -224,16 +248,14 @@ static void pp_init(struct page_pool* pool, paddr_t base, size_t size)
...
-    }
+    bitmap_t* pool_bitmap = NULL;
 
-    if ((pool->bitmap = (bitmap_t*)mem_alloc_map(&cpu()->as, SEC_HYP_GLOBAL, &pages, INVALID_VA,
-             bitmap_size, PTE_HYP_FLAGS)) == NULL) {
+    if (!mem_bitmap_pool_alloc(bitmap_size, &pool_bitmap)) {
         return;
     }
 
+    pool->bitmap = pool_bitmap;
```
josecm: Can't we just pass poo->bitmap directly?

src/core/mem.c
```diff
@@ -145,26 +169,28 @@ void* mem_alloc_page(size_t num_pages, enum AS_SEC sec, bool phys_aligned)
...
 
-    if (root_pool->size <= bitmap_size) {
-        return false;
+    if (root_pool->size <= bitmap_num_pages) {
+        return bitmap_alloced;
+    }
+
+    bitmap_t* root_bitmap = NULL;
+
+    if (!mem_bitmap_pool_alloc(NUM_PAGES(root_pool->size), &root_bitmap)) {
```
josecm: Shouldn't this be 

```suggestion
    if (!mem_bitmap_pool_alloc(bitmap_size, &root_bitmap)) {
```

josecm (commented):
The meaning of the bitmap_size parameter in `mem_bitmap_pool_alloc` is confusing. Is it the size of the bitmap in bits (i.e., the number of bits in the bitmap) or the size of the bitmap in actual bytes? The arguments passed in the two invocations of `mem_bitmap_pool_alloc` are inconsistent. Please clarify this.

src/core/mem.c
danielRep: Fixed in 85d57091.

src/core/mem.c
danielRep: Fixed in https://github.com/bao-project/bao-hypervisor/commit/85d57091c35c69d094582a3fb776202793c8c0e0.

src/core/mem.c
danielRep: The `mem_bitmap_pool_alloc` expects number of pages. Fixed in c57de40d.

src/core/mem.c
danielRep: Fixed in 65ea3cc1.

src/core/mem.c
danielRep: Fixed in 65ea3cc1.

src/core/mem.c
danielRep: Fixed in c57de40d.

src/core/mem.c
danielRep: Fixed in 19b0f7ee.

src/core/mem.c
danielRep: Read https://github.com/bao-project/bao-hypervisor/pull/238#issuecomment-3241910683

src/core/mem.c
danielRep: Read https://github.com/bao-project/bao-hypervisor/pull/238#issuecomment-3241910683

scripts/platform_defs_gen.c
```diff
@@ -5,15 +5,30 @@
...
     printf("#define PLAT_CPU_NUM (%ld)\n", platform.cpu_num);
     printf("#define PLAT_BASE_ADDR (0x%lx)\n", platform.regions[0].base);
+
+    for(size_t i = 0; i < platform.region_num; i++)
+    {
+        size_t reg_size;
+
+        reg_size = platform.regions[i].size;
+
+        bitmap_array_size += reg_size / (8 * PAGE_SIZE) + ((reg_size % (8 * PAGE_SIZE) != 0) ? 1 : 0);
```
josecm:
```suggestion
        bitmap_array_size += (reg_size / (8 * PAGE_SIZE)) + (((reg_size % (8 * PAGE_SIZE)) != 0) ? 1 : 0);
```

src/core/mem.c
```diff
@@ -18,6 +18,28 @@ extern uint8_t _image_start, _image_load_end, _image_end, _vm_image_start, _vm_i
 
 struct list page_pool_list;
 
+static bool mem_bitmap_pool_alloc(size_t bitmap_num_pages, bitmap_t** bitmap)
```
josecm:
```suggestion
static bool pp_bitmap_alloc(size_t pool_num_pages, bitmap_t** bitmap)
```

This would make more sense:

- the function is a page_pool function and in line with other function names should start with pp (page pool bitmap allocation function)
  
- the argument is actually the number of pages in the pool, not the number of pages of the bitmap

src/core/mem.c
```diff
@@ -18,6 +18,28 @@ extern uint8_t _image_start, _image_load_end, _image_end, _vm_image_start, _vm_i
...
+    static spinlock_t bitmap_lock = SPINLOCK_INITVAL;
+    static size_t last_index = 0;
+    bool allocated = false;
+
+    spin_lock(&bitmap_lock);
+
+    if (((PLAT_BITMAP_POOL_SIZE - last_index) * 8) >= bitmap_num_pages) {
+        size_t prev_last_index = last_index;
+
+        last_index += bitmap_num_pages / 8;
```
josecm: Aren't we missing the remainder? Something like:

```suggestion
        last_index += (bitmap_num_pages / 8) + ((bitmap_num_pages % 8) > 0 ? 1 : 0);
```

I'm also thinking we could define a macro for this operation in the bitmap header. For example:

```
#define BITMAP_SIZE_IN_BYTES(NUM_BITS) (((NUM_BITS) / 8) + (((NUM_BITS) % 8) > 0 ? 1 : 0))
```

And replace this operation whenever it appears in the code.

What do you think?

src/core/mem.c
```diff
@@ -224,18 +228,10 @@ static void pp_init(struct page_pool* pool, paddr_t base, size_t size)
...
     }
 
-    pages = mem_alloc_ppages(cpu()->as.colors, bitmap_num_pages, false);
-    if (pages.num_pages != bitmap_num_pages) {
-        return;
-    }
-
-    if ((pool->bitmap = (bitmap_t*)mem_alloc_map(&cpu()->as, SEC_HYP_GLOBAL, &pages, INVALID_VA,
-             bitmap_size, PTE_HYP_FLAGS)) == NULL) {
+    if (!mem_bitmap_pool_alloc(bitmap_num_pages, &pool->bitmap)) {
```
josecm: Now that `mem_bitmap_pool_alloc` parameter is the number of pages in the page pool, shouldn't we just use `pool->num_pages` as we do in `root_pool_set_up_bitmap` ?

src/core/mem.c
```diff
@@ -224,18 +228,10 @@ static void pp_init(struct page_pool* pool, paddr_t base, size_t size)
...
         return;
```
josecm: Should we return a boolean value in this function so we can check if the allocation fails?

scripts/platform_defs_gen.c
danielRep: Fix in 1a9c7d312761b72dad909f8f93b65650981918b1

src/core/mem.c
danielRep: Totally agree. Fixed in 69938be4

src/core/mem.c
danielRep: Yes, it is more consistent. Changed in 7831c633

src/core/mem.c
danielRep: Fixed in 7831c633

src/core/mem.c
danielRep: Yes. Fixed in 7831c633

scripts/platform_defs_gen.c
```diff
@@ -5,15 +5,30 @@
...
     printf("#define PLAT_CPU_NUM (%ld)\n", platform.cpu_num);
     printf("#define PLAT_BASE_ADDR (0x%lx)\n", platform.regions[0].base);
+
+    for(size_t i = 0; i < platform.region_num; i++)
+    {
+        size_t reg_size;
+
+        reg_size = platform.regions[i].size;
+
+        bitmap_array_size += (reg_size / (8 * PAGE_SIZE)) + (((reg_size % (8 * PAGE_SIZE)) != 0) ? 1 : 0);
```
josecm:
```suggestion
        bitmap_array_size += BITMAP_SIZE_IN_BYTES(NUM_PAGES(reg_size));
```

danielRep: Fixed b37408658924a4f2c90e58d82c9f5fbc07d1355b

src/lib/inc/bitmap.h
```diff
@@ -20,7 +20,7 @@ static const bitmap_granule_t ONE = 1;
 #define BITMAP_GRANULE_MASK(O, L)           BIT32_MASK((O), (L))
 
 #define BITMAP_SIZE(SIZE)                   (((SIZE) / BITMAP_GRANULE_LEN) + ((SIZE) % BITMAP_GRANULE_LEN ? 1 : 0))
-
+#define BITMAP_SIZE_IN_BYTES(NUM_BITS)      (((NUM_BITS) / 8) + (((NUM_BITS) % 8) > 0 ? 1 : 0))
```
josecm:
```suggestion
#define BITMAP_SIZE(NUM_BITS, BITS_PER_UNIT)                   (((NUM_BITS) / (BITS_PER_UNIT)) + (((NUM_BITS) % (BITS_PER_UNIT)) ? 1 : 0))
#define BITMAP_SIZE_IN_GRANULE(NUM_BITS)    (BITMAP_SIZE((NUM_BITS), (BITMAP_GRANULE_LEN)))
#define BITMAP_SIZE_IN_BYTES(NUM_BITS)      (BITMAP_SIZE((NUM_BITS), 8))
```

And change where we used `BITMAP_SIZE` by `BITMAP_SIZE_IN_GRANULE`

danielRep: Done. 1b089d52bd3c1e2eca0bfb209bdd3fe2442b68f1

src/core/objpool.c
```diff
@@ -9,7 +9,7 @@
 void objpool_init(struct objpool* objpool)
 {
     memset(objpool->pool, 0, objpool->objsize * objpool->num);
-    memset(objpool->bitmap, 0, BITMAP_SIZE(objpool->num));
+    memset(objpool->bitmap, 0, BITMAP_SIZE_IN_GRANULE(objpool->num));
```
josecm: I think the original code had a bug. Since we are using memset should be:

```suggestion
    memset(objpool->bitmap, 0, BITMAP_SIZE_IN_BYTES(objpool->num));
```

danielRep: Fix 76536e9e

danielRep: Fix 76536e9e

josecm (approved):

PR #241, opened 2025-08-26
fix(mem): fix root pool bitmap size

malejo97:
This PR fixes the calculation of the root pool's bitmap size.

The expression `root_pool->size` in the `root_pool_set_up_bitmap` function was being mistakenly interpreted as the size of the root pool in **bytes**, when it is actually the size of the root pool in **number of pages**.

To avoid further misunderstandings, this PR also changes the name of the `struct page_pool.size` field to `num_pages`

### DIFF
--- a/src/core/mem.c
+++ b/src/core/mem.c
@@ -51,7 +51,7 @@ bool pp_alloc(struct page_pool* pool, size_t num_pages, bool aligned, struct ppa
      */
     for (size_t i = 0; i < 2 && !ok; i++) {
         while (pool->free != 0) {
-            ssize_t bit = bitmap_find_consec(pool->bitmap, pool->size, curr, num_pages, false);
+            ssize_t bit = bitmap_find_consec(pool->bitmap, pool->num_pages, curr, num_pages, false);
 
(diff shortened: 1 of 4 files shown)


src/core/mem.c
```diff
@@ -51,7 +51,7 @@ bool pp_alloc(struct page_pool* pool, size_t num_pages, bool aligned, struct ppa
      */
     for (size_t i = 0; i < 2 && !ok; i++) {
         while (pool->free != 0) {
-            ssize_t bit = bitmap_find_consec(pool->bitmap, pool->size, curr, num_pages, false);
+            ssize_t bit = bitmap_find_consec(pool->bitmap, pool->num_pages, curr, num_pages, false);
```
miguelafsilva5, line 54: Shouldn't this be `pool->num_page * PAGE_SIZE`? Or something along these lines? Or is it actually the number of pages we want?

malejo97: Actually I think here we want the number of pages. Within the bitmap, each page is represented by a bit, so the size/length of the bitmap in number of bits is the number of pages of the pool.

PR #243, opened 2025-08-29
feat(plat): add platform initialization functions

danielRep:
This PR introduces two new platform initialization functions to support platforms lacking firmware-provided initialization.

- platform_default_init: performs the minimal setup required for Bao to run (e.g., UART pinout).
- platform_config_init: allows users to extend the initialization with additional platform-specific configuration.

In addition, we added a generic MMIO helper header to provide a uniform and straightforward way of accessing platform configuration registers.

### DIFF
--- a/src/core/inc/platform.h
+++ b/src/core/inc/platform.h
@@ -32,4 +32,8 @@ struct platform {
 
 extern struct platform platform;
 
+void platform_init(void);
+void platform_default_init(void);
+void platform_config_init(void);
+
 #endif /* __PLATFORM_H__ */
--- a/src/core/init.c
+++ b/src/core/init.c
@@ -24,6 +24,8 @@ void init(cpuid_t cpu_id, paddr_t load_addr)
 
     /* -------------------------------------------------------------- */
 
+    platform_init();
+
     console_init();
 
     if (cpu_is_master()) {
--- a/src/core/objects.mk
+++ b/src/core/objects.mk
@@ -15,3 +15,4 @@ core-objs-y+=objpool.o
 core-objs-y+=hypercall.o
 core-objs-y+=shmem.o
 core-objs-y+=remio.o
+core-objs-y+=platform.o
--- a/src/core/platform.c
+++ b/src/core/platform.c
@@ -1,0 +1,16 @@
+/**
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) Bao Project and Contributors. All rights reserved.
+ */
+
+#include <platform.h>
+
+__attribute__((weak)) void platform_default_init(void) { }
+
+__attribute__((weak)) void platform_config_init(void) { }
+
+void platform_init(void)
+{
+    platform_default_init();
+    platform_config_init();
+}


src/lib/inc/mmio.h
josecm, line 1: Why are we adding this file?

danielRep: As a way to uniform the accesses to mmio during the platform_default_init. I think we should avoid structs  defining peripherals such as clock trees for size reasons and have macros to registers that we want to access.

danielRep: Something like this:

`src/platform/<plat>/inc/plat_regs.h`
```
/* Clock Generation Module */
#define MC_CGM_5_BASE           (0x42830000UL)
#define MC_CGM_5_TOP            (0x4283FFFFUL)
#define MC_CGM_5_SIZE           (MC_CGM_5_TOP - MC_CGM_5_BASE + 1)
#define MC_CGM_5_MUX_2_CSC      (MC_CGM_5_BASE + 0x380UL)
#define MC_CGM_5_MUX_2_CSS      (MC_CGM_5_BASE + 0x384UL)
#define MC_CGM_5_MUX_2_DC_0     (MC_CGM_5_BASE + 0x388UL)

#define MC_CGM_5_MUX_2_CSC_SAFE (1UL << 3UL)

#define MC_CGM_5_MUX_2_CSS_SWIP (1UL << 16UL)

#define MC_CGM_5_MUX_2_DC_0_DE  (1UL << 31UL)
```

`src/platform/<plat>/platform.c`
```
static void plat_clock(void)
{
    uint32_t reg_val;
    /* Configure clock settings for console */
    /* Map MC_CGM_5 region */
    mem_alloc_map_dev(&cpu()->as, SEC_HYP_GLOBAL, (uint32_t)INVALID_VA, (uint32_t)MC_CGM_5_BASE,
        NUM_PAGES(MC_CGM_5_SIZE));

    /* Configure clock source for console (FIRC_CLK)*/
    mmio_write(MC_CGM_5_MUX_2_CSC, MC_CGM_5_MUX_2_CSC_SAFE);

    do {
        reg_val = mmio_read(MC_CGM_5_MUX_2_CSS);
    } while ((reg_val & MC_CGM_5_MUX_2_CSS_SWIP) != 0);

    /* Enable divider */
    mmio_write(MC_CGM_5_MUX_2_DC_0, MC_CGM_5_MUX_2_DC_0_DE);

    /* Unmap MC_CGM_5 region */
    mem_unmap(&cpu()->as, MC_CGM_5_BASE, NUM_PAGES(MC_CGM_5_SIZE), false);
}
```

josecm: I don't love it. I'd prefer we'd stick to our idiom of having the full struct defined for peripherals unless there is a strong technical reason against it.

> I think we should avoid structs defining peripherals such as clock trees for size reasons and have macros to registers that we want to access.

What do you mean by size? Code size? I don't think that is an huge issue.

josecm: @bao-project/core-team anyone wants to weigh on this?

miguelafsilva5: I think I like the idea of having mmio_write and mmio_read functions. Make the code more readable.
What would be the alternative to a simple write to a single mmio register?
` *(unsigned long*)(MC_CGM_5_MUX_2_DC_0) = MC_CGM_5_MUX_2_DC_0_DE;`
as opposed to:
`    mmio_write(MC_CGM_5_MUX_2_DC_0, MC_CGM_5_MUX_2_DC_0_DE);`

I see these functions similar to the ones in bit.h or SFR acessors. Improvements on code readability. Maybe they don't need their one file. But I am also not opposed to having structs for all peripherals.

danielRep: > What do you mean by size? Code size? I don't think that is an huge issue.

In my mind, I wanted to avoid to have a "huge" header file defining those big structs while we typically would be accessing 2,3 fields of registers per peripheral struct. But maybe that's not an obvious issue.

josecm: > What would be the alternative to a simple write to a single mmio register?
> ` *(unsigned long*)(MC_CGM_5_MUX_2_DC_0) = MC_CGM_5_MUX_2_DC_0_DE;`
> as opposed to:
> ` mmio_write(MC_CGM_5_MUX_2_DC_0, MC_CGM_5_MUX_2_DC_0_DE);`

The alternative is access hardware as we do in all other parts of the code. Define a volatile structure with the hardware register layout, and use a pointer for that struct to access the hardware.

> In my mind, I wanted to avoid to have a "huge" header file defining those big structs while we typically would be accessing 2,3 fields of registers per peripheral struct. But maybe that's not an obvious issue.

My issue is having two diferent ways of accessing MMIO registers. Are you expecting that so many large structures in this kind of platform management MMIO?

josecm: @bao-project/core-team What do you think? Should we have two different ways of accessing hardware? Or stay consistent? I see benefits in the suggested approach, but I'm not keen in adding to much variability for the same kind of operations in the code.

danielRep: Just to give some numbers: on the S32Z, the MMIO peripherals that Bao needs to access require structures representing the following registers:

- MC_ME: 73 registers
- MC_RGM: 15 registers
- SIUL2: 35 registers
- MC_CGM: 19 registers

At first glance, I actually expected there to be more registers. However, I noticed that the largest block (SIUL2) can benefit from using arrays to represent contiguous pin configuration registers. Initially, I assumed there would be more gaps between them. So, I believe your solution is a good comprise and we don't need to add this new type of operation.

josecm: I'd just prefer to keep the MMIO access idiom as is, unless there is strong technical reasons to add a new API for it.

josecm: @miguelafsilva5 @malejo97 do you see any strong reasons for this in tc4 or rh850 platforms?

miguelafsilva5: After discussing it further with @danielRep, I believe @josecm is correct. It's better to keep the same "idom" as you said.

danielRep: Should I remove this then @josecm ?

josecm: Please do. If in the future we find a stronger need for this API for a specific platform we can quickly add it back

danielRep: Done @josecm @miguelafsilva5

miguelafsilva5 (approved):

PR #247, opened 2025-09-16
fix/datanocopy: add section to hold boot related variables

danielRep:
This PR replaces the `pad_load` output section for a `datanocopy` section that has two goals:
- Hold data to be initialized and used during early stages of the boot (e.g., barriers, img_addr) that can be rewritten in non-unified platforms, since data is typically copied from a non-volatile memory to a data RAM
- Force the alignment in the end of the section to ensure the loadable portion of the image is page-aligned to avoid gaps. These gaps have seen to create issues on platforms such as Tricore.


src/arch/armv8/aarch32/boot.S
```diff
@@ -154,6 +154,19 @@ _set_master_cpu:
...
+    /* Copy data from RX to RWX */
+    ldr r7, =_data_lma_start        // LMA start
+    ldr r11, =_data_vma_start       // VMA start
+    ldr r12, =_image_load_end       // LMA end
+    bl copy_data
+
+    ldr r1, =_data_vma_start
+    ldr r3, =data_addr
+    str r1, [r3] // store image load address in img_addr
+#endif
```
josecm: Why is this added in this PR? Shoudn't the copy of data be unaltered in this commit?

danielRep: It was missing in the PR #218

josecm: Should we have a separate PR for that fix?

danielRep: Split. Extra changes committed in https://github.com/bao-project/bao-hypervisor/pull/250

danielRep:
> @danielRep are we sure aarch64 and riscv boot does not require this either?

I think we should create a PR adding those changes as soon as we have platforms to test it, no?

josecm (approved):

PR #252, opened 2025-09-25
feat(phys-irqs): add support to physical interrupts

miguelafsilva5:
## PR Description

This PR introduces support for physical IRQs. 
Architectures that use hardware IRQs need to have the IPC interrupts assigned to the correct VM, which wasn't being done.

This PR was tested on TC4 that uses hw irqs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -204,6 +204,9 @@ endif
 ifeq ($(plat_mem),non_unified)
 	build_macros+=-DMEM_NON_UNIFIED
 endif
+ifeq ($(phys_irqs_only),y)
+	build_macros+=-DPHYS_IRQS_ONLY
+endif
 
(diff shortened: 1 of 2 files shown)


josecm:
> ut couldn't the same macro be used for direct injection on arm? Or do you intend to add yet another macro? @josecm

I'm thinking with direct injection we'll try to do it on a VM by VM basis. So that will need to be a field on the vm struct

miguelafsilva5:
> > ut couldn't the same macro be used for direct injection on arm? Or do you intend to add yet another macro? @josecm
> 
> I'm thinking with direct injection we'll try to do it on a VM by VM basis. So that will need to be a field on the vm struct

I've pushed your macro suggestion

danielRep (approved):

PR #260, opened 2025-10-17
feat(slave side prot): Add support for slave side mmio protection

miguelafsilva5:
## PR description

This PR introduces support for slave-side MMIO protection. This is needed on MCU platforms that otherwise would require each device to be mapped in the MPU. With this PR, the entire MMIO region(s) is mapped on the MPU and each architecture/platform needs to implement a functions that enables and controls the access to each device. This function is called during vm device initialization.

It is important to note that, when installing an emul handler for a device inside the mmio region, if slave side mmio protection is enabled, we have to unmap the device memory range.

This PR was only tested for TC4, as it is the current platform that needs this feature. Before accepting the PR we have to make sure all other platforms are running as expected.

### DIFF
--- a/src/core/vm.c
+++ b/src/core/vm.c
@@ -20,8 +20,6 @@ static void vm_master_init(struct vm* vm, const struct vm_config* vm_config, vmi
     vm->lock = SPINLOCK_INITVAL;
 
     cpu_sync_init(&vm->sync, vm->cpu_num);
-
-    vm_mem_prot_init(vm, vm_config);
 }
 
 static void vm_cpu_init(struct vm* vm)
@@ -317,6 +315,12 @@ struct vm* vm_init(struct vm_allocation* vm_alloc, const struct vm_config* vm_co
      *  Initialize each virtual core.
      */
     vm_vcpu_init(vm, vm_config);
+
+    cpu_sync_barrier(&vm->sync);
+
+    if (master) {
+        vm_mem_prot_init(vm, vm_config);
+    }
 
     cpu_sync_barrier(&vm->sync);
 


Makefile
```diff
@@ -204,8 +204,17 @@ endif
...
+# Could be mmio_prot -> slave_side, mpu, etc
+ifeq ($(mmio_slave_side_prot),y)
+	build_macros+=-DMMIO_SLAVE_SIDE_PROT
+endif
+
+# Consistency check
+ifeq ($(mmio_slave_side_prot),y)
+  ifneq ($(arch_mem_prot),mpu)
+    $(error mmio_slave_side_prot=y requires arch_mem_prot=mpu)
+  endif
```
josecm: We can put this inside the previous if.

src/core/vm.c
```diff
@@ -212,8 +210,12 @@ static void vm_init_dev(struct vm* vm, const struct vm_config* vm_config)
 
         size_t n = ALIGN(dev->size, PAGE_SIZE) / PAGE_SIZE;
```
josecm: This should be inside the else case below since it is not used in the first case.

src/core/vm.c
```diff
@@ -212,8 +210,12 @@ static void vm_init_dev(struct vm* vm, const struct vm_config* vm_config)
...
 
-        if (dev->va != INVALID_VA) {
-            mem_alloc_map_dev(&vm->as, SEC_VM_ANY, (vaddr_t)dev->va, dev->pa, n);
+        if (DEFINED(MMIO_SLAVE_SIDE_PROT)) {
+            vm_arch_allow_mmio_access(dev);
+        } else {
+            if (dev->va != INVALID_VA) {
```
josecm:
```suggestion
        } else if (dev->va != INVALID_VA) {
```

josecm (changes requested):

Makefile
miguelafsilva5: Done

src/core/vm.c
miguelafsilva5: Good catch. Done

miguelafsilva5:
> @miguelafsilva5 please resolve conflicts again. Ping me as soon as it is done so it doesn't happen again with other PRs accepted to main.

@josecm should be fixed

Makefile
```diff
@@ -204,8 +204,13 @@ endif
...
 endif
-ifeq ($(phys_irqs_only),y)
-	build_macros+=-DPHYS_IRQS_ONLY
+
+ifeq ($(mmio_slave_side_prot),y)
+	build_macros+=-DMMIO_SLAVE_SIDE_PROT
+
+	ifneq ($(arch_mem_prot),mpu)
+		$(error mmio_slave_side_prot=y requires arch_mem_prot=mpu)
+	endif
```
DavidMCerdeira: deleted by mistake?
```
ifeq ($(phys_irqs_only),y)
	build_macros+=-DPHYS_IRQS_ONLY
```

src/core/vm.c
```diff
@@ -20,8 +20,6 @@ static void vm_master_init(struct vm* vm, const struct vm_config* vm_config, vmi
     vm->lock = SPINLOCK_INITVAL;
 
     cpu_sync_init(&vm->sync, vm->cpu_num);
-
-    vm_mem_prot_init(vm, vm_config);
```
josecm, line 24: Also deleted by mistake? Why are we removing this?

src/core/vm.c
```diff
@@ -319,6 +318,12 @@ struct vm* vm_init(struct vm_allocation* vm_alloc, const struct vm_config* vm_co
 
     cpu_sync_barrier(&vm->sync);
 
+    if (master) {
+        vm_mem_prot_init(vm, vm_config);
+    }
+
+    cpu_sync_barrier(&vm->sync);
+
```
danielRep, line 326: I don't understand why is moved from `vm_master_init` to here, at least in this PR

Makefile
miguelafsilva5: This was a mistake in one of the rebases. Thanks for catch. It is fixed in the force push.

src/core/vm.c
miguelafsilva5: We didn't remove, we moved it. I split that commit into 2. The new one, with this change provides an explanation for this change. 27b39473cb407a6e0335b5a45d05320a620bb0a7

josecm: Oh ok got it.

josecm (approved):

PR #261, opened 2025-10-24
feat(unique-asid): Make address spaces IDs unique

miguelafsilva5:
## PR Description

This PR makes the IDs fo address spaces unique. The only exception is the hypervisor address space that always has ID = 0.
This change is needed because the MPU driver only has access to the address space, and not vm/hypervisor structs. Some platforms like TC4, use the as->id to distinguish between vms. Prior to this change, the first VM (vm->id = 0) would have it's address space ID equal to 0, which would collide with the hypervisor address space ID.

This PR still has to be tested on all platforms. It compiles, but hasn't been tested.

### DIFF
--- a/src/core/mmu/mem.c
+++ b/src/core/mmu/mem.c
@@ -764,7 +764,7 @@ void mem_color_hypervisor(const paddr_t load_addr, struct mem_region* root_regio
      * the new CPU region is created, cleaned, prepared and finally mapped.
      */
     cpu_new = copy_space((void*)BAO_CPU_BASE, sizeof(struct cpu), &p_cpu);
-    as_init(&cpu_new->as, AS_HYP_CPY, HYP_ASID, NULL, colors);
+    as_init(&cpu_new->as, AS_HYP_CPY, NULL, colors);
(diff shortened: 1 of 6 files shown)


src/core/mmu/mem.c
```diff
@@ -900,13 +900,28 @@ void mem_color_hypervisor(const paddr_t load_addr, struct mem_region* root_regio
...
+    spin_lock(&as_id_alloc_lock);
+    if (as->type == AS_HYP) {
+        as->id = 0;
+    } else {
+        as->id = asid_counter;
+        asid_counter++;
+    }
+    spin_unlock(&as_id_alloc_lock);
+}
+
```
josecm: It would make more sense for this just return the allocarted asid and set as->id separately with the return value

josecm (changes requested):
The macro HYP_ASID doesnt seem to be used anymore and should be deleted.

src/core/mmu/mem.c
miguelafsilva5: Fixed. Also removed the HYp_ASID.

PR #265, opened 2025-11-05
feat(armv8r): handle different timer init cases

josecm:
On armv8r there is no firmware and the system timer must be initialized by the hypevisor. There are multiple cases from where to get the frequecy: (i) it might be fixed already in the hardware, or (ii) it can be retrieved from the plaform level timer if such timer is defined, or (iii) the hypervisor must know by default which is it (which must be defined by platform descrption from the `PLAT_GENERIC_TIMER_FREQ_HZ` macro). This PR handles all this cases.

### DIFF
--- a/src/arch/armv8/armv8-r/vmm.c
+++ b/src/arch/armv8/armv8-r/vmm.c
@@ -15,24 +15,35 @@ static uint32_t timer_freq = 0;
 void vmm_arch_profile_init()
 {
     if (cpu_is_master()) {
-        /**
-         * Since there is no firmware in cortex-r platforms, we need to initialize the system
-         * counter.
-         */
-        volatile struct generic_timer_cntctrl* timer_ctl;
-        timer_ctl = (struct generic_timer_cntctrl*)mem_alloc_map_dev(&cpu()->as, SEC_HYP_PRIVATE,
-            platform.arch.generic_timer.base_addr, platform.arch.generic_timer.base_addr,
-            sizeof(struct generic_timer_cntctrl));
+        unsigned long cur_cntfrq = sysreg_cntfrq_el0_read();
+        if (cur_cntfrq != 0UL) {
+            timer_freq = (uint32_t)cur_cntfrq;
+        } else if (platform.arch.generic_timer.fixed_freq != 0) {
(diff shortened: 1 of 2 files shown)


danielRep (dismissed):

josecm:
> be reliant on platform defines. Shouldn't we try to use the inf

What do you mean? The only thing I can't think it might make sense is having the frequency define part of the platform description instead of only being a macro (possibly generate the macro using the scripts).

src/arch/armv8/armv8-r/vmm.c
```diff
@@ -15,24 +15,38 @@ static uint32_t timer_freq = 0;
...
+
+                timer_ctl->CNTCR |= GENERIC_TIMER_CNTCTL_CNTCR_EN;
+                fence_ord_write();
+
+                timer_freq = (uint32_t)timer_ctl->CNTDIF0;
+
+                mem_unmap(&cpu()->as, (vaddr_t)timer_ctl, sizeof(struct generic_timer_cntctrl),
+                    false);
+            }
+#endif
```
DavidMCerdeira: This code segment would benefit from an `if(DEFINED(<>))`. This would remove need for ifdef'ed code, which we have been avoid successfully in the rest of the hypervisor code, and make it easier to read IMO.
```c
if (cur_cntfrq != 0UL) {
...
} else if (DEFINED(PLAT_GENERIC_TIMER_FREQ_HZ)) {
...
} else if (platform.arch.generic_timer.base_addr == 0) {
...
} else {
...
}
```

DavidMCerdeira: As you mentioned before we could also extend the platform description. Specifically, adding a `fixed_freq` field and checking if it is set. In this case this should replace the `PLAT_GENERIC_TIMER_FREQ_HZ` macro IMO.
```c
...
} else if (platform.arch.generic_timer.fixed_freq != 0) {
...
```

josecm: @DavidMCerdeira followed this suggestion in the latest forced push.

I kept the `platform.arch.generic_timer.base_addr == 0` check inside the last else case, as this is essentially a check that we have the address to go forward with the last of the three options.

DavidMCerdeira (approved):

PR #266, opened 2025-11-05
Add mps3 plat support

josecm:
This PR adds support for the cortex-r52 based MPS3 support. See https://documentation-service.arm.com/static/615dbd31ac265639eac54439?token=.

We only tested this support using the QEMU version of the platform (https://www.qemu.org/docs/master/system/arm/mps2.html)

This PR is dependent on #265.

### DIFF
--- a/src/platform/drivers/cmsdk_uart/cmsdk_uart.c
+++ b/src/platform/drivers/cmsdk_uart/cmsdk_uart.c
@@ -1,0 +1,34 @@
+/**
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) Bao Project and Contributors. All rights reserved.
+ */
+
+#include <bao.h>
+#include <drivers/cmsdk_uart.h>
+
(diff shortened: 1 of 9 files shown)


josecm:
Done @danielRep

src/platform/drivers/cmsdk_uart/cmsdk_uart.c
```diff
@@ -0,0 +1,24 @@
...
+}
+
+void uart_enable(volatile struct cmsdk_uart_hw* ptr_uart)
+{
+    ptr_uart->ctrl = 0x1;
+}
+
+void uart_putc(volatile struct cmsdk_uart_hw* ptr_uart, int8_t c)
+{
+    while (ptr_uart->state & 0x1) { }
```
danielRep: Can we use macros instead of magic numbers?

josecm: Fixed in force push

danielRep (approved):

danielRep (dismissed):

PR #271, opened 2025-11-06
ref: move shmem lock init

josecm:


danielRep (approved):

miguelafsilva5 (approved):

DavidMCerdeira (commented):
I'm not understanding the goal here to be honest.
Isn't the end result the same?
I feel like it might be related to separate allocation from initialization, but I'm not sure.
If anything I'd consider moving the `.cpu_masters = 0` to the alloc function to avoid iterating the array again.

danielRep (approved):
LGTM

PR #283, opened 2025-11-24
Fix/riscv csr macros width

josecm:


danielRep (changes requested):
@josecm can you just correct the typos on the commit message?

`fix(riscv): make csr leteral macros unsigned long to adapt o SXLEN`

danielRep (approved):

malejo97 (approved):

PR #293, opened 2025-12-18
use circular ring for IPI messages

giacomogroppi:
Static allocation of messages via a circular queue still allows message delivery in FIFO order and reduces latency, since it does not use dynamic memory. 

To enable this functionality, you need to set the `IPI_STATIC_MEM` define


DavidMCerdeira:
Greetings @giacomogroppi!

Thank you for the update, but the build is still failing and unfortunately we don't have the bandwidth to dig into this right now. 
We'd welcome a fix from your side if you manage to track it down. Happy to review whenever you're ready.

giacomogroppi:
Hi, I think I've managed to fix the build. I'm not too keen on having to pass a pointer to the macro just to check whether the operation was successful or not, so if you have any other ideas, let me know.

DavidMCerdeira:
Looks like the builds are fixed! Great!

> I'm not too keen on having to pass a pointer to the macro just to check whether the operation was successful or not, so if you have any other ideas, let me know.

Instead of macros it's better to use `static inline` functions, this way we can easily check the return value.

giacomogroppi:
ok, i'll update the code in the next days, thank you

giacomogroppi:
Hi! I've made the changes we discussed by converting some macros into static functions. I look forward to hearing back from you.

DavidMCerdeira:
Hi @giacomogroppi I finally got the chance to work on this.

For the sake time I've made a few modifications to this PR directly:
- Rebased this PR with the main branch
- Fixed a bug where the code silently ignored IPI messages when a CPU's queue fills up.
- Cleaned up leftover objpool declaration.
- Split the commit into two (add the queue, then switch cpu.c to use it).
- Added the SPDX to the `circular_queue.h` file
- Provided small description of the advantages of this approach in the commit message

tested platforms:
- qemu-virt-aarch64
- qemu-virt-riscv - there seems to be an issue with running bao-demos on the bao main branch but I cherry-picked this PR on top of v2.0.0 and it execute succesfully

Please sign-off these commits, by doing so you acknowledge the bao project's DCO. For more information check [documentation](https://bao-project.readthedocs.io/en/latest/development/contributing.html)

src/lib/inc/circular_queue.h
```diff
@@ -0,0 +1,113 @@
...
+ * Takes a pointer to the struct that contains the queue (and its buffer) and the
+ * bare field name so the macro can resolve __cq_buf_<name> automatically.
+ *
+ * Usage:  cq_init(container_ptr, field_name)
+ * e.g.:   cq_init(cpu()->interface, msgs)
+ */
+#define cq_init(container, name)                                          \
+    circular_queue_init(&(container)->name, (container)->__cq_buf_##name, \
+        sizeof((container)->__cq_buf_##name[0]),                          \
+        sizeof((container)->__cq_buf_##name) / sizeof((container)->__cq_buf_##name[0]))
```
josecm: Macros should be uppercase:


```suggestion
#define CQ_DEFINE(type, name, N) \
    type __cq_buf_##name[(N)];   \
    struct circular_queue name

/**
 * Initializes a circular_queue field, transparently binding it to its static buffer.
 *
 * Takes a pointer to the struct that contains the queue (and its buffer) and the
 * bare field name so the macro can resolve __cq_buf_<name> automatically.
 *
 * Usage:  cq_init(container_ptr, field_name)
 * e.g.:   cq_init(cpu()->interface, msgs)
 */
#define CQ_INIT(container, name)                                          \
    circular_queue_init(&(container)->name, (container)->__cq_buf_##name, \
        sizeof((container)->__cq_buf_##name[0]),                          \
        sizeof((container)->__cq_buf_##name) / sizeof((container)->__cq_buf_##name[0]))
```

src/core/cpu.c
```diff
@@ -56,25 +43,17 @@ void cpu_init(cpuid_t cpu_id)
 
 void cpu_send_msg(cpuid_t trgtcpu, struct cpu_msg* msg)
 {
-    struct cpu_msg_node* node = objpool_alloc(&msg_pool);
-    if (node == NULL) {
-        ERROR("cant allocate msg node\n");
+    bool ok = circular_queue_push(&cpu_if(trgtcpu)->msgs, msg);
+    if (!ok) {
+        ERROR("Can't add message to target cpu (%d) interface\n", trgtcpu);
```
josecm: This should only emit a warning. If a core stalls and stops draining its input queue, another core attempting to send it a message should not stall as well. The send should fail and the message should be dropped.

josecm (changes requested):

josecm:
@giacomogroppi sorry about the delay on reviweing this! This patch is very much appreciated. Let me know if you still have the time to work on it. Otherwise, I'll address my suggestions directly.

src/core/cpu.c
giacomogroppi: Hi, no problem, I can dedicate some time to the changes. If it's not possible to add the message to the queue, should I still send the IPI to the other processor, or should I just return from the function?

josecm: It would make more sense to just return :thinking:

josecm:
@giacomogroppi Thanks for addressing this so quickly. I think you missed updating the `cq_define` and `cq_init` calls to use uppercase names as well.

Could you also clean up the commit history before I proceed with the final merge?

josecm (dismissed):

josecm (approved):

DavidMCerdeira (approved):

josecm:
Thanks again for this contribution @giacomogroppi!

PR #294, opened 2025-12-19
fix(rpi): psci error on powerdown and standy

DavidMCerdeira:
Bao invokes PSCI_SUSPEND to achieve powerdown and standby, however RPi's TF-A PSCI does not implement a psci_suspend handler.


josecm (approved):
It is weird the PSCI implementation doesn't simply return `PSCI_E_NOT_SUPPORTED` so we could automatically recover and not use these macros to avoid calling the function. Neverthless, as we discussed offline, it seems it does not for some platforms, including the RPI4. So, everything looks good to me.

PR #296, opened 2026-01-06
Add support for NXP S32Z270 platform

danielRep:
This draft PR introduces the initial support for the NXP S32Z270 platform in Bao.

The branch has been tested with different VM setups, from a 1x to 4x core baremetal guest side-by-side with the Zephyr OS.

We still consider as a draft PR since we need to add core features:
~~[ ] Add xRDC support~~ -> Postponed for a future PR
~~[ ] Add platform clock initialization~~ -> This should be done per guest-specific platform initialization. As such, Bao runs with the default system clock source, i.e., FIRC = 48MHz

Other related tasks:

- [x] Create PR with the baremetal guest support for S32Z -> https://github.com/bao-project/bao-baremetal-guest/pull/20
- [x] Add baremetal to bao-demos -> https://github.com/bao-project/bao-demos/pull/76
- [x] Add zephyr tro bao-demos -> https://github.com/bao-project/bao-demos/pull/76

### DIFF
--- a/src/platform/s32z270/inc/plat/plat_regs.h
+++ b/src/platform/s32z270/inc/plat/plat_regs.h
@@ -1,0 +1,326 @@
+/**
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) Bao Project and Contributors. All rights reserved.
+ */
+
+#ifndef __PLAT_REGS_H__
+#define __PLAT_REGS_H__
+
+#include <bao.h>
+
+#define RTU0_CORE0 0
+#define RTU0_CORE1 1
+#define RTU0_CORE2 2
+#define RTU0_CORE3 3
+
+/* Mode Entry Module (MC_ME) */
+#define MC_ME_BASE (0x41900000UL)
+
+struct mc_me_hw {
+    uint32_t CTL_KEY;
+    uint32_t MODE_CONF;
+    uint32_t MODE_UPD;
+    uint8_t RESERVED_0[244];
+    uint32_t PRTN0_PCONF;
+    uint32_t PRTN0_PUPD;
+    uint32_t PRTN0_STAT;
+    uint8_t RESERVED_1[52];
+    struct {
+        uint32_t PCONF;
+        uint32_t PUPD;
+        uint32_t STAT;
+        uint32_t ADDR;
+        uint8_t RESERVED_2[16];
+    } PRTN0_COREX[8];
+    uint8_t RESERVED_3[208 - 16];
+    uint32_t PRTN1_PCONF;
+    uint32_t PRTN1_PUPD;
+    uint32_t PRTN1_STAT;
+    uint8_t RESERVED_4[52];
+    struct {
+        uint32_t PCONF;
+        uint32_t PUPD;
+        uint32_t STAT;
+        uint32_t ADDR;
+        uint8_t RESERVED_5[16];
+    } PRTN1_COREX[4];
+    uint8_t RESERVED_14[336 - 16];
+    uint32_t PRTN2_PCONF;
+    uint32_t PRTN2_PUPD;
+    uint32_t PRTN2_STAT;
+    uint8_t RESERVED_6[52];
+    struct {
+        uint32_t PCONF;
+        uint32_t PUPD;
+        uint32_t STAT;
+        uint32_t ADDR;
+        uint8_t RESERVED_7[16];
+    } PRTN2_COREX[3];
+    uint32_t PRTN2_CORE3_PCONF;
+    uint32_t PRTN2_CORE3_PUPD;
+    uint32_t PRTN2_CORE3_STAT;
+    uint32_t PRTN2_CORE3_ADDR;
+};
+
+#define MC_ME_CTL_KEY                (MC_ME_BASE + 0x0UL)
+
+#define MC_ME_CTL_KEY_KEY0           (0x5AF0)
+#define MC_ME_CTL_KEY_KEY1           (0xA50F)
+
+// RTU0 - S32Z Partition1
+#define MC_ME_PRTN1_CORE0_PCONF      (MC_ME_BASE + 0x340UL)
+#define MC_ME_PRTN1_CORE0_PUPD       (MC_ME_BASE + 0x344UL)
+#define MC_ME_PRTN1_CORE0_STAT       (MC_ME_BASE + 0x348UL)
+#define MC_ME_PRTN1_CORE0_ADDR       (MC_ME_BASE + 0x34CUL)
+
+#define MC_ME_PRTN1_COREX_OFFSET     (0x20UL)
+
+#define MC_ME_PRTN1_COREX_STAT_CCS   (1UL << 0)
+#define MC_ME_PRTN1_COREX_PCONF_CCE  (1UL << 0)
+#define MC_ME_PRTN1_COREX_PUPD_CCUPD (1UL << 0)
+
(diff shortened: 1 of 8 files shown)


src/arch/armv8/aarch32/inc/arch/subarch/sysregs.h
```diff
@@ -91,6 +91,9 @@ SYSREG_GEN_ACCESSORS_MERGE(mair_el2, hmair0, hmair1)
 SYSREG_GEN_ACCESSORS(hcr, 4, c1, c1, 0)
 SYSREG_GEN_ACCESSORS(hcr2, 4, c6, c0, 0)
 SYSREG_GEN_ACCESSORS_MERGE(hcr_el2, hcr, hcr2)
+SYSREG_GEN_ACCESSORS(hactlr, 4, c1, c0, 1)
+SYSREG_GEN_ACCESSORS(hactlr2, 4, c1, c0, 3)
+SYSREG_GEN_ACCESSORS_MERGE(actlr_el2, hactlr, hactlr2) // hactlr
```
josecm: I'm confused as why this register was defined since I don't see it being used anywhere in the PR.

src/arch/armv8/armv8-r/mpu.c
```diff
@@ -171,6 +171,7 @@ void mpu_enable(void)
     sysreg_sctlr_el2_write(reg_val);
 
     ISB();
+    DSB(ish);
```
josecm, line 200: I'd prefer this fix to be addressed in a dedicated PR or at least commit.

danielRep: Fixed here: https://github.com/bao-project/bao-hypervisor/pull/300

josecm:
> The sysreg definition is related with the enable of private peripheral access in the `actlr` register which is implementation-defined per core (e.g., zephyr requires to access LLPP bus). So, we need an infrastructure for those type of impl-defined initialization as I don't see any ideal place.

But shouldn't the use of it be addressed in this PR then?

danielRep:
> > The sysreg definition is related with the enable of private peripheral access in the `actlr` register which is implementation-defined per core (e.g., zephyr requires to access LLPP bus). So, we need an infrastructure for those type of impl-defined initialization as I don't see any ideal place.
> 
> But shouldn't the use of it be addressed in this PR then?

Since this is particular to Cortex-R52 I was going to introduce it here as an individual commit. Still didn't had it because I wanted to discuss the best approach to introduce those type of impl-defined configurations

josecm:
> Since this is particular to Cortex-R52

But is the operation we need to carry out on this register common to all Cortex-R52 instances in all platforms? Or just to cortex-r52 on s32z?

> I was going to introduce it here as an individual commit

Please do.

> I wanted to discuss the best approach to introduce those type of impl-defined configurations

Yes, we should discuss  this in more depth, in case we need to add more support for this auxiliary registers in other cores.

danielRep:
> Yes, we should discuss this in more depth, in case we need to add more support for this auxiliary registers in other cores.

Yes, already left the reasoning in our bi-weekly meeting agenda.

josecm:
> > Yes, we should discuss this in more depth, in case we need to add more support for this auxiliary registers in other cores.
> 
> Yes, already left the reasoning in our bi-weekly meeting agenda.

Nevertheless, regarding this PR specifically, shouldn't somewhere be an access to hactlr?

src/platform/s32z270/inc/plat/psci.h
```diff
@@ -0,0 +1,16 @@
...
+
+#ifndef __PLAT_PSCI_H__
+#define __PLAT_PSCI_H__
+
+#define PSCI_POWER_STATE_LVL_0    0x0000000   // TBD
+#define PSCI_POWER_STATE_LVL_1    0x1000000   // TBD
+#define PSCI_POWER_STATE_LVL_2    0x2000000   // TBD
+#define PSCI_STATE_TYPE_STANDBY   0x00000     // TBD
+#define PSCI_STATE_TYPE_POWERDOWN (0UL << 30) // TBD
+#define PSCI_STATE_TYPE_BIT       (0UL << 30) // TBD
```
DavidMCerdeira: What's TBD here?
It would be nice to have a comment explaining
The major point I see is that since platforms don't have EL3 Bao will never issue PSCI calls. But is it expected for guest to do so?

src/platform/s32z270/s32z270_desc.c
```diff
@@ -0,0 +1,54 @@
...
+
+        .gic = {
+            .gicd_addr = 0x47800000,
+            .gicr_addr = 0x47900000,
+            .maintenance_id = 25
+        },
+
+        .generic_timer = {
+            .base_addr = 0x44820000,
+            .fixed_freq = 8000000       // 8 MHz (FXOSC/RTU.GPR.CFG_CNTDV = 40 MHz / 4+1)
```
DavidMCerdeira: This is depend on the BSP init, no?
Maybe add a comment that it could potentially be redefined during runtime

src/platform/s32z270/inc/plat/plat_regs.h
```diff
@@ -0,0 +1,705 @@
...
+    uint16_t PGPDI4;
+    uint16_t PGPDI7;
+    uint16_t PGPDI6;
+    uint16_t PGPDI9;
+    uint16_t PGPDI8;
+    uint8_t RESERVED_19[2];
+    uint16_t PGPDI10;
+    uint8_t RESERVED_20[40];
+    uint32_t MPGPDO[11];
+};
```
DavidMCerdeira, line 170: consider replacing large contiguous unused fields as byte arrays
We only use a few fields yet this is ~400 LoC long

src/platform/s32z270/s32z270_desc.c
```diff
@@ -0,0 +1,54 @@
...
+
+#include <platform.h>
+
+struct platform platform = {
+
+    .cpu_num = 4,
+    .cpu_master_fixed = true,
+    .cpu_master = 0,
+    .region_num = 2,
+    .regions =  (struct mem_region[]) {
```
DavidMCerdeira, line 14: as we talked about in the meeting today, please add the remaining platform memories so that they can be tracked by the hypervisor

src/platform/s32z270/s32z270_desc.c
danielRep: Fixed here: 55f71eb1

src/platform/s32z270/inc/plat/psci.h
danielRep: This is a dependency that should be solved in another PR I believe. Add comment here: 6f0f5e3a

src/platform/s32z270/inc/plat/plat_regs.h
danielRep: Fixed here: e7159dfe

src/platform/s32z270/s32z270_desc.c
danielRep: Fixed here. ebc4480d

DavidMCerdeira: the `region_num` must be updated as well

DavidMCerdeira (dismissed):

src/platform/drivers/linflexd_uart/sources.mk
josecm, line 1: This doesn't seem necessary

DavidMCerdeira (approved):

Issue #290, opened 2025-12-11
[BUG] Entry mask should be per vCPU nor per VM


josecm:
https://github.com/bao-project/bao-hypervisor/blob/41a0bea0cd99421d2f9013fa904246671e0b20f9/src/arch/armv8/armv8-r/inc/arch/mem.h#L40

At the moment the `mpu_entry_mask` used to enable guest regions upon entry vcpu entry is kept in `struct addr_space_arch`, which is a structure instantiated per VM. However, this mask refers to the entries in the physical MPU and each physical MPU is managed locally. This means that the mask might be different for different vCPUs running the same VM. Therefore, one mask should be kept and managed per vCPU.

The common case (and ATM, maybe the only case) is for the mask to be the same for all vCPUs running within the same VM. For this reason, no issue was observed. However this might not be true in the future.

miguelafsilva5:
> > What do you think of moving the entry mask to the arch_regs struct? And make it part of the vcpu context?
> 
> This would work since this refers to a hypervisor register (prenr_el2) that is part of the vcpu's context. And similar registers are located in this struct. However, this struct is dependent on subarch and we would need to define the field two times (for aarch32 and aarch64). I'd prefer it to be part of `struct vcpu_arch` since it would need to be added only once.

I was working on this. The thing is, we modify the `mpu_entry_mask` on the mpu functions (e.g. `mpu_map`), and in these functions we only have access to the addr_space. That is why we initially placed this mask inside the addr_space struct.

Do you see any way to address this?

josecm:
My preferred solution would be a "find_local_mpu_struct_by_asid" function that would search for the target struct based on the asid. Essentially it would only need to check if the asid corresponded to the cpu's asid, or the the local vcpu's, failing neither.

PR #297, opened 2026-01-06
fix(mpu_mask): move mpu_entry_mask from vm struct to vcpu

miguelafsilva5:
## PR Description

This PR addresses issue https://github.com/bao-project/bao-hypervisor/issues/290 

The entry mask being in the vm struct made so that all cpus would use the same mask and change it for all other cpus.
This PR fixes this by moving the mask to the vcpu struct.

### DIFF
--- a/src/arch/armv8/armv8-r/inc/arch/mem.h
+++ b/src/arch/armv8/armv8-r/inc/arch/mem.h
@@ -36,10 +36,6 @@ typedef union {
     };
 } mem_flags_t;
 
-struct addr_space_arch {
-    unsigned long mpu_entry_mask;
-};
-
 #define PTE_FLAGS(_prbar, _prlar) \
     ((mem_flags_t){               \
         .prbar = (_prbar),        \
--- a/src/arch/armv8/armv8-r/mpu.c
+++ b/src/arch/armv8/armv8-r/mpu.c
@@ -95,6 +95,17 @@ static bool mpu_entry_clear(mpid_t mpid)
     return true;
 }
 
+static unsigned long* find_mpu_mask_by_asid(asid_t id)
+{
+    unsigned long* mpu_mask = NULL;
+    if (cpu()->as.id == id) {
+        mpu_mask = &cpu()->arch.profile.mpu.mpu_entry_mask;
+    } else if (cpu()->vcpu->vm->as.id == id) {
+        mpu_mask = &cpu()->vcpu->arch.mpu_entry_mask;
+    }
+    return mpu_mask;
+}
+
 bool mpu_map(struct addr_space* as, struct mp_region* mpr, bool locked)
 {
     mpid_t mpid = INVALID_MPID;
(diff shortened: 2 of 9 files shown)


src/arch/armv8/armv8-r/inc/arch/mem.h
```diff
@@ -37,7 +37,7 @@ typedef union {
 } mem_flags_t;
 
 struct addr_space_arch {
-    unsigned long mpu_entry_mask;
+    EMPTY_STRUCT_FIELDS
 };
```
josecm: I think we can fully remove this struct now.

src/arch/armv8/armv8-r/mpu.c
```diff
@@ -95,6 +95,19 @@ static bool mpu_entry_clear(mpid_t mpid)
...
+{
+    if (cpu()->as.id == id) {
+        return &cpu()->arch.profile.mpu.mpu_entry_mask;
+    }
+
+    if (cpu()->vcpu->vm->as.id == id) {
+        return &cpu()->vcpu->arch.mpu_entry_mask;
+    }
+
+    return NULL;
```
josecm:
```suggestion
    unsigned long *mpu_mask = NULL;

    if (cpu()->as.id == id) {
        mpu_mask = &cpu()->arch.profile.mpu.mpu_entry_mask;
    } else if (cpu()->vcpu->vm->as.id == id) {
        mpu_mask = &cpu()->vcpu->arch.mpu_entry_mask;
    }

    return mpu_mask;
```

It is good pratice to maintain a single return point, especially in simple functions.

src/arch/armv8/armv8-r/mpu.c
```diff
@@ -115,7 +128,16 @@ bool mpu_map(struct addr_space* as, struct mp_region* mpr, bool locked)
             if (locked) {
                 mpu_entry_lock(mpid);
             }
-            bitmap_set((bitmap_t*)&as->arch.mpu_entry_mask, mpid);
+
+            unsigned long* mpu_mask = find_mpu_mask_by_asid(as->id);
+
+            if (mpu_mask == NULL) {
+                ERROR("Couldn't find an mpu_mask for as_id %d\n");
```
josecm: I am not sure if we need an error here. Just return false should suffice. Calling functions should check the success of the mapping through the return value. Nevertheless, if this if case is executed, something has gone wrong indeed - maybe a warning?

If we do keep the error, I don't think we need to return.

src/arch/armv8/armv8-r/mpu.c
```diff
@@ -139,7 +160,14 @@ bool mpu_unmap(struct addr_space* as, struct mp_region* mpr)
 
         mpu_entry_clear(mpid);
 
-        bitmap_clear((bitmap_t*)&as->arch.mpu_entry_mask, mpid);
+        unsigned long* mpu_mask = find_mpu_mask_by_asid(as->id);
+
+        if (mpu_mask == NULL) {
+            ERROR("Couldn't find an mpu_mask for as_id %d\n");
+            return false;
```
josecm: Same as above.

src/arch/armv8/armv8-r/mpu.c
miguelafsilva5: My bad. Fixed in the amended commit.

src/arch/armv8/armv8-r/inc/arch/mem.h
miguelafsilva5: Removed in the amended commit

src/arch/armv8/armv8-r/mpu.c
miguelafsilva5: I was debating on whether to use ERROR or WARNING. I also don't think that the calling function will pass an incorrect address space. For that to happen, something really really wrong must have happened. Changed to a warning in the amended commit either way

PR #304, opened 2026-01-13
fix(riscv/aia): imsic offset depends on number of guest files

josecm:
Previously the implementation assumed a single guest interrupt file per IMSIC. With this all interrupt files were contiguous in memory and the offset between each hart's interrupt files' base was two pages (s- and vs- level respectively). However, when the number of guest interrupt files is larger than 1, this is not true. This patch removes this assumption and calculates the size of each IMSIC memory address range (and therfore, offset between each hart's IMSIC base address) according to the layout mandated in the AIA spec version 1, which is depentent on GEILEN.


josecm:
> `IMSIC_NUM_FILES` seems to be left unused. I think we should remove it

Fixed. And rebased on main.

PR #305, opened 2026-01-13
fix(riscv/sbi): add missing unlock in hsm suspend handler

josecm:


danielRep (approved):
Don't forget to add the description to the PR @josecm

PR #308, opened 2026-01-27
feat(e3650): add platform e3650

leon6002:
This PR added support for Semidrive's E3650 platform.

The related PR:
https://github.com/bao-project/bao-demos/pull/75

### DIFF
--- a/src/arch/armv8/armv8-r/cpus/core_impl.c
+++ b/src/arch/armv8/armv8-r/cpus/core_impl.c
@@ -6,7 +6,8 @@
 #include <arch/sysregs.h>
 #include <arch/core_impl.h>
 
-#define ARMV8R_R52_PARTNUMBER 0xD13
+#define ARMV8R_R52_PARTNUMBER  0xD13
+#define ARMV8R_R52P_PARTNUMBER 0xD16
 
 /* Weak references to CPU implementation init functions */
 extern void cortex_r52_impl_init(void) __attribute__((weak));
@@ -18,7 +19,8 @@ void cpu_arch_core_impl_init(void)
     uint32_t part_num = MIDR_EL1_PARTNUM(midr);
 
     switch (part_num) {
-        case ARMV8R_R52_PARTNUMBER: /* Cortex-R52 */
+        case ARMV8R_R52_PARTNUMBER:  /* Cortex-R52 */
+        case ARMV8R_R52P_PARTNUMBER: /* Cortex-R52+ */
             if (cortex_r52_impl_init != NULL) {
                 cortex_r52_impl_init();
             }
--- a/src/platform/e3650/e3650_desc.c
+++ b/src/platform/e3650/e3650_desc.c
@@ -1,0 +1,45 @@
+/**
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) Bao Project and Contributors. All rights reserved.
+ */
+
+#include <platform.h>
+
+struct platform platform = {
+
+    .cpu_num = 4,
+    .cpu_master_fixed = true,
+    .cpu_master = 0,
+
+    .region_num = 1,
+    .regions =  (struct mem_region[]) {
+        { // IRAM (3MB total, 256KB reserved for Bootloader at start)
+            .base = 0x00A40000, 
+            .size = 0x002C0000, 
+        }
+    },
+
+
+
+    .console = {
+        .base = E3650_UART_BASE, 
+    },
+
+    .arch = {
+        .gic = {
+            .gicd_addr = E3650_GICD_BASE,
+            .gicr_addr = E3650_GICR_BASE,
+            .maintenance_id = 25,
+        },
+
+        .generic_timer = {
+            .fixed_freq = 600000000,
+        },
+
+        .clusters = {
+            .num = 1,
+            .core_num =  (size_t[]) { 4 },
+        },
+    },
+
+};
--- a/src/platform/e3650/inc/plat/psci.h
+++ b/src/platform/e3650/inc/plat/psci.h
@@ -1,0 +1,19 @@
+/**
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) Bao Project and Contributors. All rights reserved.
+ */
+
+#ifndef __PLAT_PSCI_H__
+#define __PLAT_PSCI_H__
+
+/* On E3650 platform PSCI is not used to interact with the firmware. However, our baremetal guest
+uses a PSCI call to wake its secondary cores. In the future we may remove support for PSCI on Arm
+MPU platforms entirely. */
+#define PSCI_POWER_STATE_LVL_0    0x0000000
+#define PSCI_POWER_STATE_LVL_1    0x1000000
+#define PSCI_POWER_STATE_LVL_2    0x2000000
+#define PSCI_STATE_TYPE_STANDBY   0x00000
+#define PSCI_STATE_TYPE_POWERDOWN (0UL << 30)
+#define PSCI_STATE_TYPE_BIT       (0UL << 30)
+
+#endif /* __PLAT_PSCI_H__ */
--- a/src/platform/e3650/objects.mk
+++ b/src/platform/e3650/objects.mk
@@ -1,0 +1,5 @@
+## SPDX-License-Identifier: Apache-2.0
+## Copyright (c) Bao Project and Contributors. All rights reserved.
+
+boards-objs-y+=e3650_desc.o
+
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ platforms is presented below:
 **Armv8-R AArch32**
 - [x] Arm Fixed Virtual Platforms
 - [x] Arm MPS3-AN536
+- [x] Semidrive E3650
 - [x] NXP S32Z270
 - [ ] Renesas RZT2M
 
--- a/src/platform/drivers/e3650_uart/e3650_uart.c
+++ b/src/platform/drivers/e3650_uart/e3650_uart.c
@@ -1,0 +1,72 @@
+/**
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) Bao Project and Contributors. All rights reserved.
+ */
+
+#include <drivers/e3650_uart.h>
+#include <fences.h>
+
+void uart_init(volatile struct e3650_uart* uart)
+{
+    /* 0. Check if already enabled*/
+    if (uart->mcr0 & MCR0_MODEN) {
+        return;
+    }
+
+    /* 1. Disable Module */
+    uart->mcr0 = 0;
+
+    /* 2. Reset Module */
+    uart->mcr0 |= MCR0_MODRST;
+    while (uart->mcr0 & MCR0_MODRST) {
+        /* Wait for self-clearing reset */
+    }
+
+    /* 3. Configure Mode & Data Frame */
+    /* Defaults: Serial Mode, Full Duplex, 8N1 */
+    uart->pcr0 = 0;
+
+    /* 4. Clear FIFOs */
+    /* Write 1 to Bit 16 to clear TX/RX FIFOs */
+    uart->fcr0 = 0;
+    uart->fcr1 = 0;
+    uart->fcr0 |= FCR_FIFO_CLR;
+    uart->fcr1 |= FCR_FIFO_CLR;
+
+    /* 5. Configure Filters */
+    /* RXSYNCEN (Bit 17) | RXFILTCTL=3 (Bits 18-20) */
+    uint32_t filter_cfg = PCR8_RXSYNCEN | PCR8_RXFILTCTL_VAL;
+    uart->pcr8 = filter_cfg;
+    uart->pcr10 = filter_cfg;
+
+    /* 6. Set Baud Rate */
+    uart->pcr1 = UART_BAUDRATE_DIV;
+
+    /* 7. Enable TX/RX */
+    uart->pcr0 |= (PCR0_TXEN | PCR0_RXEN);
+
+    /* 8. Disable Debug Mode configuration */
+    uart->mcr0 |= MCR0_DEBUG_CTRL;
+    uart->mcr0 &= ~MCR0_DEBUG_SWEN;
+}
+
+void uart_enable(volatile struct e3650_uart* uart)
+{
+    uart->mcr0 |= MCR0_MODEN;
+}
+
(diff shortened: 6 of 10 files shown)


src/platform/e3650/e3650.c
```diff
@@ -0,0 +1,75 @@
...
+
+
+static inline void e3650_enable_imp_def(void)
+{
+    unsigned long val;
+    unsigned long lo, hi;
+
+    /* Enable Prefetch */
+    /* mrrc p15, 0, r0, r1, c15 : read AxCACHE PREFETCH CONTROL REGISTER */
+    __asm__ volatile("mrrc p15, 0, %0, %1, c15" : "=r"(lo), "=r"(hi));
```
danielRep: We use the macros SYSREG_GEN_ACCESSORS* to generate the appropriate accessors for system registers. Please modify accordingly throughout `e3650_enable_imp_def`

src/platform/e3650/e3650.c
```diff
@@ -0,0 +1,75 @@
...
+    hi &= ~3U;           /* Clear bit 0,1 */
+    lo &= ~(7U << 13);
+    /* CONFIG_DATA_PREFETCH_OT = 5 */
+    lo |= (1U << 11) | (5U << 13);
```
danielRep: Please avoid using magical numbers and switch for macros.

src/platform/e3650/e3650.c
```diff
@@ -0,0 +1,75 @@
...
+    __asm__ volatile("mcrr p15, 0, %0, %1, c15" :: "r"(lo), "r"(hi));
+
+    /* Enable Branch Prediction */
+    val = 0;
+    __asm__ volatile("mcr p15, 1, %0, c9, c1, 1" :: "r"(val));
+    __asm__ volatile("isb");
```
danielRep:
```suggestion
   ISB();
```

src/platform/e3650/e3650.c
```diff
@@ -0,0 +1,75 @@
...
+    /* Enable FLASHIFREGIONR (Flash) */
+    __asm__ volatile("mrc p15, 0, %0, c15, c0, 1" : "=r"(val));
+    val |= FLASHIFREGIONR_ENABLE;
+    __asm__ volatile("mcr p15, 0, %0, c15, c0, 1" :: "r"(val));
+
+    /* IMP_CSCTLR : AXIM and Flash port cache way configuration */
+    val = (ICACHE_FLASH_2_AXIM_2 | DCACHE_FLASH_2_AXIM_2);
+    __asm__ volatile("mcr p15, 1, %0, c9, c1, 0" :: "r"(val));
+
+    __asm__ volatile("isb");
```
danielRep:
```suggestion
    ISB();
```

src/platform/e3650/objects.mk
```diff
@@ -0,0 +1,6 @@
+## SPDX-License-Identifier: Apache-2.0
+## Copyright (c) Bao Project and Contributors. All rights reserved.
+
+boards-objs-y+=e3650_desc.o
+boards-objs-y+=e3650.o
```
danielRep: `e3650.c` should be named `platform.c` to keep consistency. We're using that naming for the source file containing the definition of `void platform_default_init(void)`

danielRep (changes requested):
The PR is already in a good state for approval. There is, however, an issue with ` e3650_enable_imp_def: these type of core implementation-defined configurations should be done in a different manner. We discussed a possible solution for these cases in our weekly meeting. We'll try to introduce the necessary infrastructure for this in the following days. I'll tag this PR as soon as that PR is introduced.

Please solve also the issues with gitlint (commit sign-off) and the code needs to be format using `make format`

src/platform/e3650/objects.mk
leon6002: Renamed e3650.c to platform.c as suggested. Thanks!

src/platform/e3650/e3650.c
leon6002: I have refactored the system register access using the SYSREG_GEN_ACCESSORS macros as suggested. I've also verified the changes on the actual hardware, and the platform behaves as expected. Please let me know if the current implementation aligns with the project's standards.

src/platform/e3650/e3650.c
leon6002: I have replaced the magic numbers with macros as suggested.

src/platform/e3650/e3650.c
leon6002: Refactored as suggested.

src/platform/e3650/e3650.c
leon6002: Refactored as suggested.

src/platform/e3650/objects.mk
leon6002: I have also formatted the code using `make format` as suggested.

src/platform/e3650/e3650_desc.c
```diff
@@ -0,0 +1,65 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) Bao Project and Contributors. All rights reserved.
+ */
+
+#include <platform.h>
+
+#define E3650_UART_BASE 0xF8D60000 // UART15 Base
+
+#define E3650_GICD_BASE 0xF4000000 // GIC Distributor
+#define E3650_GICR_BASE 0xF4100000 // GIC Redistributor (Core 0)
```
danielRep: Can you move this to the platform.h?

josecm (dismissed):

src/platform/e3650/e3650_desc.c
```diff
@@ -0,0 +1,65 @@
...
+        },
+        { // GIC Distributor
+            .base = E3650_GICD_BASE,
+            .size = 0x10000,
+        },
+        { // GIC Redistributor
+            .base = E3650_GICR_BASE,
+            .size = 0x20000,
+        },
+    },
```
danielRep: The `mmio_regions` field it's supposed to be used in conjunction with the `MMIO_SLAVE_SIDE_PROT` configuration, which we are not using in this platform. The idea is to map all mmio regions so that protection is guaranteed at the slave side and not by the MPU. This allows us to save MPU entries. As such, this is unnecessary here. The information in `.console`and `.gic` is sufficient to map those in the MPU.

src/platform/e3650/e3650_desc.c
```diff
@@ -0,0 +1,65 @@
...
+
+    .arch = {
+        .gic = {
+            .gicd_addr = E3650_GICD_BASE,
+            .gicr_addr = E3650_GICR_BASE,
+            .maintenance_id = 25,
+        },
+
+        .generic_timer = {
+            .base_addr = 0xF8C20000, // APB_SYS_CNT_BASE
```
danielRep: The R52 is a special case the generic timer is implementation-defined. By configuring in this field the based address of the timer, we are falling into the case where we read freq from memory-mapped timer control registers. Is this true in E3650? Because from past experience on porting R52-based platforms, we had to explicitly force the freq using `fixed_freq` field as no memory-mapped registers were available.

danielRep (changes requested):

src/platform/e3650/inc/plat/psci.h
```diff
@@ -0,0 +1,16 @@
...
+
+#ifndef __PLAT_PSCI_H__
+#define __PLAT_PSCI_H__
+
+#define PSCI_POWER_STATE_LVL_0    0x0000000   // TBD
+#define PSCI_POWER_STATE_LVL_1    0x1000000   // TBD
+#define PSCI_POWER_STATE_LVL_2    0x2000000   // TBD
+#define PSCI_STATE_TYPE_STANDBY   0x00000     // TBD
+#define PSCI_STATE_TYPE_POWERDOWN (0UL << 30) // TBD
+#define PSCI_STATE_TYPE_BIT       (0UL << 30) // TBD
```
danielRep:
```suggestion
/* On E3650 platform PSCI is not used to interact with the firmware. However, our baremetal guest
uses a PSCI call to wake its secondary cores. In the future we may remove support for PSCI on Arm
MPU platforms entirely. */
#define PSCI_POWER_STATE_LVL_0    0x0000000
#define PSCI_POWER_STATE_LVL_1    0x1000000
#define PSCI_POWER_STATE_LVL_2    0x2000000
#define PSCI_STATE_TYPE_STANDBY   0x00000
#define PSCI_STATE_TYPE_POWERDOWN (0UL << 30)
#define PSCI_STATE_TYPE_BIT       (0UL << 30)
```

src/platform/e3650/e3650_desc.c
```diff
@@ -0,0 +1,65 @@
...
+    .cpu_master_fixed = true,
+    .cpu_master = 0,
+
+    .region_num = 1,
+    .regions =  (struct mem_region[]) {
+        { // Start after 256KB Bootloader reservation
+            .base = 0x00A40000, 
+          // 3MB - 256KB
+            .size = 0x002C0000, 
+        }
```
danielRep: Is the memory layout of the platform? Doesn't E3650 have more memory regions?

src/platform/e3650/platform.c
```diff
@@ -0,0 +1,80 @@
...
+    /* IMP_CSCTLR : AXIM and Flash port cache way configuration */
+    val = (ICACHE_FLASH_2_AXIM_2 | DCACHE_FLASH_2_AXIM_2);
+    sysreg_imp_csctlr_write(val);
+
+    ISB();
+}
+
+void platform_default_init(void)
+{
+    e3650_enable_imp_def();
```
danielRep: In the last week, we have introduced this commit that allows to enable core-specific register to be initialized by the hypervisor. From our analysis, on Cortex-R52 these are sufficient for most guests. Is there any reason do you need the rest of your configuration? For example, from the Cortex-R52, the IMPL_BPCTLR has a reset value 0x0, meaning that the dynamic branch predictor is enabled by reset. If there isn't a special case for touching these extra registers, just remove this function since it is no longer needed after a rebase with main.

danielRep (changes requested):

src/platform/e3650/e3650_desc.c
leon6002: sure

src/platform/e3650/e3650_desc.c
leon6002: The E3650 does have the `sys_cnt` registers, but for consistency, I've updated the configuration to use `fixed_freq`. 
image

src/platform/e3650/e3650_desc.c
leon6002: Thanks for the clarification. I have removed the mmio_regions configuration.

src/platform/e3650/platform.c
leon6002: I have removed `platform.c` as requested. However, the `cortex_r52_impl_init()` function in `src/arch/armv8/armv8-r/cpus/core_impl.c` was not invoked successfully, which caused the e3650 to fail during startup. After testing, I found that forcing the execution of `cortex_r52_impl_init()` by bypassing the conditional checks allowed the system to start up successfully. This suggests the issue may be related to `ARMV8R_R52_PARTNUMBER` not matching the expected value `0xD13`. I am continuing to investigate this issue.

danielRep: You solved this by introducing `PLAT_CPUS:=cortex_r52`, right?

src/arch/armv8/aarch32/inc/arch/subarch/sysregs.h
```diff
@@ -121,6 +121,8 @@ SYSREG_GEN_ACCESSORS(sctlr_el2, 4, c1, c0, 0)
 SYSREG_GEN_ACCESSORS(hactlr, 4, c1, c0, 1)
 SYSREG_GEN_ACCESSORS(imp_periphregionr, 0, c15, c0, 0)
 SYSREG_GEN_ACCESSORS(flashifregionr, 0, c15, c0, 1)
+SYSREG_GEN_ACCESSORS(imp_csctlr, 1, c9, c1, 0)
+SYSREG_GEN_ACCESSORS_64(imp_prefetch_ctrl, 0, c15)
```
danielRep: I don't see this being used. I believe they are not needed.

danielRep (changes requested):

src/arch/armv8/aarch32/inc/arch/subarch/sysregs.h
leon6002: Yes, I have just removed it.

src/platform/e3650/e3650_desc.c
leon6002: Yes, the E3650 does have multiple memory regions (like CRAM, ROM, Flash, etc.).

However, in this snippet, I grouped the continuous IRAM regions together to serve as the main system memory. According to the TRM, IRAM1, IRAM2, and IRAM3 are physically contiguous, spanning from 0x00A0_0000 to 0x00CF_FFFF (3MB in total).

Since we are reserving the first 256KB (0x40000) for the bootloader, the usable base address becomes 0x00A40000, and the remaining size is 3MB - 256KB (0x002C0000).

I only defined this continuous RAM block here as it acts as our main memory. Do you think we also need to map the CRAM or other specific memory regions in this descriptor for our current use case?
image

src/platform/e3650/platform.c
leon6002: Yes, I added `PLAT_CPUS:=cortex_r52` into `src/platform/e3650/platform.mk` so that `src/arch/armv8/armv8-r/cpus/cortex_r52.c` could be compiled.

leon6002: I think I found the issue, e3650's cpu is actually cortex-r52p, and the partnumber for r52+ is 0xD16, as showd in the log below, should we add a cortex_r52p file?:
image

src/arch/armv8/armv8-r/cpus/core_impl.c
leon6002, line 1: @danielRep please check if this approach is appropriate. Or should we add a seperate file for cortex-r52p?

src/platform/e3650/e3650_desc.c
danielRep: @leon6002 could you answer this last comment? The remaining ones I already closed since you already addressed them.

leon6002: Sorry for the delay, my previous response  was stuck in "Pending" status and I didn't realize it wasn't visible to you. Now you should be able to see it.

danielRep (approved):

josecm (approved):

PR #310, opened 2026-02-05
ref(riscv): refactor vimsic interrupt injection

josecm:
This PR refactors APLIC MSI injection in IMSIC guest interrupt files and fixes two issues:

1. The previous implementation assumed that MSI injection into a guest interrupt file always targeted the local vCPU. If the target vCPU was remote, it sent a CPU message to the target core so the MSI could be injected locally. However, the IMSIC MMIO interface allows a CPU to inject an MSI directly into a *remote* (i.e., different core) guest IMSIC via its MMIO interface. Supporting this requires:

     - Mapping guest interrupt files in the hypervisor.
     - Introducing a function that injects an MSI by writing to the appropriate guest interrupt file MMIO interface based on the target CPU and its assigned guest interrupt file.
  
     Alternatively, local vCPU injection could still use the AIA CSRs (as before). To simplify the code, this optimization is intentionally avoided for now. Future work may evaluate whether MMIO-based injection is sufficiently costly to justify reintroducing this optimization.

2. On updates to APLIC registers, the previous implementation iterated over all APLIC interrupts to determine whether an MSI should be forwarded to a vCPU. In most cases (e.g., enabling a single interrupt), only one interrupt needs to be considered. This PR addresses the issue by adding a third argument to `vaplic_update_hart`, indicating whether the call was triggered by a single interrupt state update. If the interrupt ID is invalid (e.g., when triggered by a global vAPLIC control update), the logic falls back to scanning all interrupts for forwarding.

### DIFF
--- a/src/arch/riscv/irqc/aia/inc/imsic.h
+++ b/src/arch/riscv/irqc/aia/inc/imsic.h
@@ -25,7 +25,10 @@ struct imsic_intp_file_hw {
 
 struct imsic_global_hw {
     struct imsic_intp_file_hw s_file;
+    struct imsic_intp_file_hw guest_file;
 } __attribute__((__packed__, aligned(0x1000ULL)));
+
+extern volatile struct imsic_global_hw* imsic[PLAT_CPU_NUM];
 
 /**
  * @brief Initializes the IMSIC
@@ -68,16 +71,28 @@ void imsic_set_enbl(irqid_t intp_id);
  *        Only little endian is supported.
  *
  * @param target_cpu The ID of the target CPU
+ * @param msi_id The MSI ID to be sent
  */
-void imsic_send_msi(cpuid_t target_cpu);
+static inline void imsic_send_msi(cpuid_t target_cpu, irqid_t msi_id)
+{
+    imsic[target_cpu]->s_file.seteipnum_le = msi_id;
+}
 
 /**
- * @brief Inject an interrupt into a guest.
+ * @brief Sends an MSI to the guest file of specified CPU with the specified IPI ID.
  *
- * @param guest_file Guest interrupt file ID
- * @param intp_id Interrupt ID
+ *        The function sends an MSI to the target CPU's guest file by setting the seteipnum_le
+ *        register in the IMSIC. The seteipnum_le register is used to specify the ID of the
+ *        interrupt being sent. Only little endian is supported. This function assume a single guest
+ *        interrupt file per hart.
+ *
+ * @param target_cpu The ID of the target CPU
+ * @param msi_id The MSI ID to be injected in the guest interrupt file
  */
-void imsic_inject_pend(size_t guest_file, irqid_t intp_id);
+static inline void imsic_send_guest_msi(cpuid_t target_cpu, irqid_t msi_id)
+{
+    imsic[target_cpu]->guest_file.seteipnum_le = msi_id;
+}
 
 /**
  * @brief Handles interrupts in the IMSIC.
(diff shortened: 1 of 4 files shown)


src/arch/riscv/irqc/aia/inc/imsic.h
```diff
@@ -72,15 +75,25 @@ void imsic_set_enbl(irqid_t intp_id);
  *
  * @param target_cpu The ID of the target CPU
```
DavidMCerdeira, line 73: missing second param documentation

src/arch/riscv/irqc/aia/inc/imsic.h
```diff
@@ -72,15 +75,25 @@ void imsic_set_enbl(irqid_t intp_id);
...
+    imsic[target_cpu]->s_file.seteipnum_le = msi_id;
+}
 
 /**
- * @brief Inject an interrupt into a guest.
+ * @brief Sends an MSI to the guest file specified CPU with the specified IPI ID.
+ *
+ *        The function sends an MSI to the guest file specified CPU by setting the seteipnum_le
+ *        register in the IMSIC. The seteipnum_le register is used to specify the ID of the
+ *        interrupt being sent. Only little endian is supported. This function assume a single guest
```
DavidMCerdeira, line 86: missing second param documentation

src/arch/riscv/irqc/aia/inc/imsic.h
```diff
@@ -72,15 +75,25 @@ void imsic_set_enbl(irqid_t intp_id);
...
+ *        interrupt file per hart.
  *
- * @param guest_file Guest interrupt file ID
- * @param intp_id Interrupt ID
+ * @param target_cpu The ID of the target CPU
  */
```
DavidMCerdeira:
```suggestion
 * @brief Sends an MSI to the guest file specified CPU with the specified IPI ID.
 *
 *        The function sends an MSI to the target CPU's guest file by setting the seteipnum_le
 *        register in the IMSIC. The seteipnum_le register is used to specify the ID of the
 *        interrupt being sent. Only little endian is supported. This function assumes a single guest
 *        interrupt file per hart.
 *
 * @param target_cpu The ID of the target CPU
 */
```

src/arch/riscv/irqc/aia/vaplic.c
```diff
@@ -526,91 +544,89 @@ static bool vaplic_idc_emul_handler(struct emul_access* acc)
...
 /**
- * @brief Processes an incoming event.
+ * @brief Triggers the hart/harts interrupt line update.
  *
- * @param event the event id
- * @param data
+ * @param vcpu virtual cpu
+ * @param vhart_index virtual hart to update the interrupt line. If UPDATE_ALL_HARTS were passed,
+ *        this function will trigger the interrupt line update to all virtual harts running in this
+ *        vm.
```
DavidMCerdeira: Is this description stale?

src/arch/riscv/irqc/aia/vaplic.c
```diff
@@ -526,91 +544,89 @@ static bool vaplic_idc_emul_handler(struct emul_access* acc)
...
+        if (irq_id == INVALID_IRQID) {
+            for (irqid_t i = 1; i < APLIC_MAX_INTERRUPTS; i++) {
+                vaplic_forward_by_msi(vcpu, i);
+            }
+        } else {
+            vaplic_forward_by_msi(vcpu, irq_id);
+        }
     }
 }
+#endif
```
DavidMCerdeira:
```suggestion
#endif /* (IRQC == AIA) */
```

PR #313, opened 2026-02-12
Add support for including BSP platform initialization

DavidMCerdeira:
This modifies config support to support two cases:
1. CONFIG_REPO/CONFIG points to a single file i.e., CONFIG_REPO/CONFIG.c
2. CONFIG_REPO/CONFIG points to a directory in which case the directory
   can contain multiple sources files related to the single config, and
   a config.c file

This should make it more practical to leverage generated BSPs, e.g.,
from and IDE, with clock, port, and other platform initializations

Upon further discussion, we concluded it is best to separate BSP initalization source files, from the config. It is up to the user to bind the two if needed.

We will develop this in this PR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,12 +100,18 @@ ifneq ($(MAKECMDGOALS), clean)
  core_mem_prot_dir:=$(core_dir)/$(arch_mem_prot)
 endif
 
-# Check configuration exists and set configurtion sources based on it
-config_dir:=$(CONFIG_REPO)
-config_src:=$(wildcard $(config_dir)/$(CONFIG).c)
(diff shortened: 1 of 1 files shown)


josecm:
Here is my proposal: https://github.com/josecm/bao-hypervisor/commit/35552e551d665458430982e6fd072db36449c396

It allows for:
- configs to add arbitrary code via the objects.mk
- configs to modify build parameters through config.mk
- in a folder config to modify the name of the confic C file (no longer mandatory to be config.c, but only the default)

I didn't address the utility automatic addition of source files in a given directory. For me this is a separate issue.

PR #319, opened 2026-02-20
fix(rh850/interrupts): implement interrupts_arch_irq_is_forwardable

DavidMCerdeira:


src/arch/rh850/inc/arch/interrupts.h
```diff
@@ -14,4 +14,14 @@
...
 #define MAX_INTERRUPT_HANDLERS PLAT_MAX_INTERRUPTS
 
+static inline bool interrupts_arch_irq_is_forwardable(irqid_t int_id)
+{
+    UNUSED_ARG(int_id);
+
+    /**
+     * In RH850 the hypervisor no interrupts needs to be forwarded to guests.
+     */
+    return true;
```
josecm: So shouldn't it return false?

DavidMCerdeira: yes

PR #320, opened 2026-02-20
feat(plat): add support for BeagleBone AI-64

puranikvinit:
This PR introduces support for the BeagleBone AI-64 platform, featuring TI's TDA4VM SoC with the K3 (keystone3) architecture.


josecm:
@puranikvinit this is looking amazing! Let us know when it is at a point we can test it here.

Also, if you could in parallel start a PR for bao-demos, so you could follow it step-by-step.

puranikvinit:
Hey @josecm,
I am still working on the platform hooks and, in the process, fixing shortcomings in the implemented drivers. It will take a few more iterations for me to get to a point where we can start testing. I will keep you updated!

For `bao-demos`, I need to see how TI ships their binaries for this SoC. The SPL setup for this platform is quite involved, as I've seen with the J722S. I will start working on that as well once I figure out things better.

josecm:
Looking forward!

puranikvinit:
Hey @josecm,
I have an initial version of platform code for the board implemented now. I guess this is a good starting point to see if we can get the hypervisor running and have a driver sanity check as well. In the meanwhile, I will get the interrupt management code implemented.

puranikvinit:
Well, after you brought up the point of TF-A setting things up, I went into more details of the boot flow... What I saw is:
- TF-A has a TISCI driver, for implementing PSCI interface only
- u-boot implements a TISCI driver for boot-time board config. This part of u-boot, compiled as `tiboot3`, is executed on the bootloader processor, and, while it is nowhere mentioned in their public documentation for TISCI, I found it buried under an SDK code, that it is a one-time issued, boot time API. No other application, at any point of time in the execution can issue this API post that. So any changes that needs to be made to the board config, needs to have uboot source modified.
- The guests running, have their own TISCI drivers, which will communicate with the DMSC with their own IDs, for all resource/interrupt requests. At GIC level, Bao already manages isolation, so no work required on that end.

Taking all these points into consideration, we won't have to do anything wrt interrupts at all for the hosts other than what is already implemented. The only usecases where Bao needs this driver, is to register for interrupts which Bao will rely on, for example, SMMU interrupts. I should have studied implementations in more detail, before jumping in to code the drivers... What do we do? The docs still don't seem to explicitly mention these intricacies, hence the confusion. If Bao relies on SMMU interrupts, then yeah, we can have it registered in boot time, else we can strip out these drivers completely... Good learning though!

josecm:
> Taking all these points into consideration, we won't have to do anything wrt interrupts at all for the hosts other than what is already implemented. The only usecases where Bao needs this driver, is to register for interrupts which Bao will rely on, for example, SMMU interrupts. I should have studied implementations in more detail, before jumping in to code the drivers... What do we do? The docs still don't seem to explicitly mention these intricacies, hence the confusion. If Bao relies on SMMU interrupts, then yeah, we can have it registered in boot time, else we can strip out these drivers completely... Good learning though!

I wouldn't say it's wasted work. As you said, it seems we will require it for initializing the SMMU interrupt. And I am guessing we will actually require these drivers even if we do not include them now.

> * u-boot implements a TISCI driver for boot-time board config. This part of u-boot, compiled as `tiboot3`, is executed on the bootloader processor, and, while it is nowhere mentioned in their public documentation for TISCI, I found it buried under an SDK code, that it is a one-time issued, boot time API. No other application, at any point of time in the execution can issue this API post that. So any changes that needs to be made to the board config, needs to have uboot source modified.

Does u-boot setup the interrupt routing at this stage? If so (and this is the bulk of the board config you mention) we can even do this in bao directly so we don't have depend on u-boot. But not as a priority btw. If we do this, I'd leave it to after all other stuff is done.

> * The guests running, have their own TISCI drivers, which will communicate with the DMSC with their own IDs, for all resource/interrupt requests. At GIC level, Bao already manages isolation, so no work required on that end.

This is another point where we might need these drivers. If guest do assume they can commicate with the DMSC, Bao needs to intercept those calls, filter them, and forward them to the DMSC. So these drivers will be useful for this functionality also.

puranikvinit:
Hey @josecm,

Apologies for the long silence; I was dealing with a personal health issue and had to step away from everything for a while. I'm back now and wanted to share a proper update.

Bao **boots successfully** on the BeagleBone AI-64, with a baremetal guest running! I've also pushed the corresponding board support to `bao-demos` and `bao-baremetal-guest`, so the full flow is testable end to end now.

### Highlights:

- Boot success: The hypervisor initializes, brings up secondary cores via PSCI, and drops into the baremetal guest.
- TISCI/Secure Proxy: Bidirectional Secure Proxy communication between Bao and SYSFW is working, confirmed by successfully probing the secure proxy channels, and pinging the DMSC sysfw during platform init.
- Multicore handling: dual core boot verified; guest pinned to a single core, secondary core idled correctly.

### Caveats & Fixes

1. TISCI Secure Proxy `hw_host` check removed

There's no register in the K3 architecture that exposes the owner of a given Secure Proxy channel, so the hw_host ownership check in `k3_sec_proxy.c` was unimplementable as written. Removed it permanently.

2. TF-A `PSCI_CPU_OFF` not supported

When an unassigned CPU correctly fell through to `cpu_powerdown()`, UART output got scrambled. Turns out TI K3's TF-A doesn't support `psci_power_down()` (returns -1). Bao's `cpu_arch_profile_powerdown` was catching this and hitting the `ERROR()` macro, which drops into an infinite loop and floods the console. Patched this to fall back to `wfi` instead, so the core idles gracefully without taking down the UART. This is a temporary workaround though, I haven't dug into whether this is a TF-A config issue on our end or genuinely unsupported on this platform/TF-A version. Will investigate further.

### Open Questions:

1. Slave core powerdown - is `wfi` fallback the right long-term approach, or should we chase down the TF-A rejection?

See caveat 2 above. Open to thoughts on whether this is worth root-causing on the TF-A side, or whether falling back to `wfi` whenever `PSCI_CPU_OFF` isn't supported is a reasonable general pattern for Bao to adopt.

2. Runtime firewall gating for guest VMs - checking my understanding before I commit to an approach

Picking up our discussion from a few months back on resource/firewall management; I went back into the TISCI docs to try to nail down something concrete, and I think I found a relevant mechanism, but I want to lay out exactly what the docs say versus what I'm inferring, since I don't want to commit to an approach based on a misread.

The concrete question: say a UART is owned by Bao-EL2 per the static board config. Can Bao toggle, at runtime, which guest's priv-ID is allowed to access it (e.g. allow guest0, deny guest1)?

What the docs state directly:

> Each firewall in the system is assigned an owner. Ownership is assigned using Host ID. Only a owner of a firewall can: Configure the firewall permissions, Query the firewall permissions, Transfer the firewall ownership to another owner(host)
`TISCI_MSG_SET_FWL_REGION` (0x9000U) — Message to set a firewall region configuration
The request takes fwl_id, region, control, and permissions (described as "Contents of the firewall PERMISSION register to set"), and:
Response for configuring the firewall permissions... ACK response indicates that firewall configuration is successful.

My reading: if Bao owns a firewall (per static board config), it can call `SET_FWL_REGION` to set/change the permission bits, which are priv-ID based, without needing to touch the RM board config payload that's locked at boot. This seems different from the RM resource ownership we discussed before, which is the part that's genuinely locked after the first board config message.

Where I'm not confident, and want your read:

- The docs don't say anywhere that this message is callable more than once, or that changes take effect live without some reinit step. I'm inferring "repeatable" mainly from the absence of "one-time" language (which the RM board config docs do use explicitly).
- I don't yet know how many distinct priv-IDs are available for guests on the J721E, i.e. whether each guest can reliably get its own priv-ID to gate against, or whether some guests would share one if running on certain cores.

I'd rather have you sanity check this before I build anything on top of it. Let me know if I'm misreading the API, or if there's a cleaner way to think about this that I'm missing.

puranikvinit:
One more thing I wanted to flag separately from the above, since it's more of a general observation about the hypervisor than something specific to this platform: While bringing this port up, I ran into two issues that took far longer to diagnose than they should have, purely because of limited visibility into hypervisor/CPU state when things went wrong (made worse by the fact that I don't have the 10-pin pogo connector needed for JTAG on the BeagleBone AI-64, so a hardware debugger wasn't an option for me here...):

1. The CPU 0 powerdown race (caveat in my comment above) - a hard hang with zero console output, and no way to attach a debugger to inspect state. I ended up sprinkling in my own temporary printfs at various points in `vmm_assign_vcpu` and the powerdown path to narrow down where things were going wrong, which is what eventually let me spot the race.

2. A guest entry address misconfiguration on my end - the guest hit a Stage-2 translation fault with no register dump at the fault point. Without JTAG to inspect live state, I again added ad hoc prints around the fault handler and VM init path, and cross-referenced those against the linker script, `plat.h`, and VM config to spot the mismatch. A built-in register dump (PC/ELR, ESR, FAR, SPSR) at the fault point would've gotten me there in minutes instead of hours of manual instrumentation.

Both were resolvable eventually, but only because I kept adding my own logging by hand as I went; without a debugger available, that manual instrumentation was really my only lever, and it shouldn't have to be a one-off thing I build each time.

Would it be worth adding more verbose diagnostics at key failure points, like the `ERROR()` handlers, fault/exception entry, powerdown paths, with register state dumps (PC/ELR, ESR, FAR, SPSR, etc.) built in by default? Maybe gated behind a debug/verbose build flag (`CONFIG_BAO_VERBOSE_FAULTS` or similar), so it's available when bringing up a new platform or debugging without a debugger attached.

Happy to put together a small proposal or even a first pass at this if it's something the team would find valuable, figured I'd raise it here first since it came directly out of this bring-up experience. Can also move this to its own issue if that's a better place for it.

src/core/init.c
```diff
@@ -24,10 +24,10 @@ void init(cpuid_t cpu_id)
 
     /* -------------------------------------------------------------- */
 
-    platform_init();
-
     console_init();
 
+    platform_init();
```
josecm: This reorder breaks tc4dx, s32z270 and rh850-u2a16: their platform_default_init sets up the pinmux and clocks the console needs (s32z270 derives the baud rate from clocks configured there). I get why it moved, console_write spins until the console is ready, so the INFO prints the driver does on every read and write would hang the boot with the original order. I'd keep the order and drop those prints instead.

Which raises the question of what the probe plus ping buy us at boot. Nothing uses TISCI at runtime yet. I'd be fine mapping the proxy here and leaving the first real transaction to the SMMU interrupt route work, which is where this gets exercised for real.

src/platform/drivers/firmware/inc/drivers/tisci.h
```diff
@@ -0,0 +1,327 @@
...
+#define TISCI_MSG_REQ_FLAG_REQ_NOTFWD2DM                 (1 << TISCI_MSG_REQ_FLAG_REQ_NOTFWD2DM_IDX)
+
+/* msg response flags */
+#define TISCI_MSG_RSP_FLAG_ACK_IDX                       (1)
+
+#define TISCI_MSG_RSP_FLAG_ACK                           (1 << TISCI_MSG_RSP_FLAG_ACK_IDX)
+#define TISCI_MSG_RSP_FLAG_NAK                           (0 << TISCI_MSG_RSP_FLAG_ACK_IDX)
+
+#define TISCI_MSG_MAX_SIZE                               (56U)
+#define TISCI_MSG_TXN_TIMEOUT                            (1000) /* 1000ms */
```
josecm: This ends up in read_timeout_us, so the response timeout is 1 ms, not the 1000 ms the comment says. Linux waits 1 s for sysfw. Fix the unit in the name and the value?

src/platform/drivers/firmware/tisci.c
```diff
@@ -0,0 +1,276 @@
...
+    if (rx_len > TISCI_MSG_MAX_SIZE) {
+        ERROR("channel_%d expected msg len too long\n", chan_id);
+        return TISCI_STATUS_CODE_INVALID_MSG_LEN;
+    }
+
+    unsigned int retry_iterator = TISCI_MSG_NUM_RETRIES;
+    for (; retry_iterator > 0; retry_iterator--) {
+        int32_t recv_status = ctx->ops.recv(chan_id, rx_buf, rx_len);
+        if (TISCI_STATUS_CODE_NO_ERROR != recv_status) {
+            ERROR("channel_%d tisci msg rcv failed\n", chan_id);
```
josecm: ERROR in bao prints and spins forever, so nothing after it ever runs: not the returns, not the status codes, and not the retry loop here (the seq_id mismatch case a few lines down ERRORs as well, so it never retries either). Either the drivers ERROR themselves, and all the status plumbing can go, or they return codes and the platform decides. Given these only run at init and a failure means the platform is unusable, I'd go for the former, it removes a good chunk of code.

src/platform/drivers/firmware/tisci.c
```diff
@@ -0,0 +1,276 @@
...
+
+    if (TISCI_STATUS_CODE_NO_ERROR != rev_status) {
+        ERROR("channel_%d tisci txn failed\n", tx_chan_id);
+        return rev_status;
+    }
+
+    return TISCI_STATUS_CODE_NO_ERROR;
+}
+
+int32_t tisci_rm_irq_set(tisci_ctx* ctx, uint8_t host_id, uint8_t tx_chan_id, uint8_t rx_chan_id,
```
josecm: Nothing calls this or the release counterpart, same for mbox_k3_sec_proxy_clear, the secure header struct, the VALID_ bit defines and the irq_line fields in the thread descriptor. I'd drop them until the SMMU interrupt route brings the first user, easier to review next to it.

src/platform/drivers/mailbox/k3_sec_proxy.c
```diff
@@ -0,0 +1,322 @@
...
+            ERROR("secure_proxy_thread_%d cannot READ on WRITE thread\n", thread_id);
+            return MBOX_K3_SEC_PROXY_STATUS_CODE_INCORRECT_DRXN;
+        } else {
+            ERROR("secure_proxy_thread_%d cannot WRITE on READ thread\n", thread_id);
+            return MBOX_K3_SEC_PROXY_STATUS_CODE_INCORRECT_DRXN;
+        }
+    }
+
+    /* check if msg queue has entries before txn attempt */
+    if (MBOX_K3_SEC_PROXY_MSG_DRXN_READ == msg_drxn) {
```
josecm: u-boot polls cur_cnt for both directions, for a write thread it counts free slots. Without it a write into a full queue sets the error bit and only the next verify notices. Worth checking in both cases?

src/platform/drivers/mailbox/inc/drivers/k3_sec_proxy.h
```diff
@@ -0,0 +1,260 @@
...
+typedef struct {
+    mbox_k3_sec_proxy_instance thread_inst; /**< Secure Proxy instance configuration */
+    uint32_t read_timeout_us; /**< timeout before read operation fails due to no data in the mbox
+                                 queue */
+    mbox_k3_sec_proxy_thread_desc* sec_proxy_thread_desc; /**< Array of thread descriptors */
+    size_t num_threads; /**< Number of entries in the sec_proxy_thread_desc array */
+} mbox_k3_sec_proxy_desc;
+
+// clang-format off
+#define MBOX_K3_SEC_PROXY_THREAD_DESC_ENTRY(                                   \
```
josecm: `make format-check` fails on this header: clang-format sees the bracketed designated initializers in this macro, guesses Objective-C for the .h, then refuses our Cpp style ("Unsuitable") and CI diffs the whole file. Moving the table macro out of the header fixes it, or write the table out in the platform description, it is thirty short entries. The only other diffs are the one line read_reg/write_reg bodies and a for continuation in the description.

src/platform/beaglebone-ai64/bb-ai64_desc.c
```diff
@@ -0,0 +1,130 @@
...
+            .recv = platform_tisci_recv,
+        },
+    .seq_id_lock = {0, 0},
+    .seq_id = 0U,
+};
+
+void platform_default_init(void)
+{
+    if (cpu_is_master()) {
+        cpu_sync_init(&cpu_glb_sync, platform.cpu_num);
```
josecm: cpu_glb_sync is already initialized in cpu_init and the barrier is reusable. Re-initializing it here also races: a secondary that reaches the barrier below before the master gets here has its increment wiped and both sides wait forever. Just drop the init.

src/platform/beaglebone-ai64/bb-ai64_desc.c
```diff
@@ -0,0 +1,130 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) Bao Project and Contributors. All rights reserved.
+ */
+
+#include <platform.h>
+
+struct platform platform = {
+    .cpu_num = 2,
+    .region_num = 2,
+    .regions =
```
josecm: Region 0 spans the OP-TEE secure DDR at 0x9e800000 (24 MB, no-map in the device tree and firewalled). The pool allocates bottom up, so we get there once guests ask for around 490 MB, which a Linux guest will. Can we carve it out, one region below and one above? The R5F IPC pools right after it are plain DRAM as long as no R5 firmware runs, those I'd leave in.

src/platform/beaglebone-ai64/inc/plat/psci.h
```diff
@@ -0,0 +1,20 @@
...
+#define PSCI_POWER_STATE_LVL_0    (0x0U)
+
+#define PSCI_STATE_TYPE_STANDBY   (0x0U)
+#define PSCI_STATE_TYPE_BIT       (1UL << 16)
+
+#define PSCI_STATE_ID_RUN         (0x0U)
+#define PSCI_STATE_ID_RETENTION   (0x1U)
+#define PSCI_STATE_ID_POWERDOWN   (0x2U)
+
+#define PSCI_STATE_TYPE_POWERDOWN PSCI_STATE_TYPE_BIT | PSCI_STATE_ID_POWERDOWN
```
josecm: On the power down question from the thread: bao already has a knob for this, rpi4, imx8mp and qemu define PLAT_PSCI_POWERDOWN_NOT_SUPPORTED and the core falls back to a standby suspend, which k3 TF-A implements as wfi. Defining it here is all we need, no core changes and no wfi patch. As pushed, an unassigned core ends in the ERROR loop. The bit 16 state type layout is right for this TF-A btw, k3 doesn't select the extended state id format.

```suggestion
#define PSCI_STATE_TYPE_POWERDOWN (PSCI_STATE_TYPE_BIT | PSCI_STATE_ID_POWERDOWN)

#define PLAT_PSCI_POWERDOWN_NOT_SUPPORTED 1
```

src/platform/beaglebone-ai64/bb-ai64_desc.c
```diff
@@ -0,0 +1,130 @@
...
+void platform_default_init(void)
+{
+    if (cpu_is_master()) {
+        cpu_sync_init(&cpu_glb_sync, platform.cpu_num);
+    }
+
+    mbox_k3_sec_proxy_init(&k3_sec_proxy);
+    cpu_sync_barrier(&cpu_glb_sync);
+
+    uint8_t cur_host_id = cpu_is_master() ? J721E_HOST_ID_MAIN_A72_CONTEXT_2_NON_SECURE :
```
josecm: Do the two A72 cores present different privids to the secure proxy firewall? I'd expect the cluster to be a single initiator, in which case host 12 and host 13 are a software convention rather than isolation. Not for this PR, but we should know before we design the guest side.

src/platform/beaglebone-ai64/inc/plat/platform.h
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+#ifndef __PLAT_PLATFORM_H__
+#define __PLAT_PLATFORM_H__
+
+#define UART8250_REG_WIDTH (4U)
+
+#ifndef GENERATING_DEFS
+#include <drivers/8250_uart.h>
+#include <drivers/k3_sec_proxy.h>
```
josecm: Smaller things, grouped:

- Only the description needs the driver headers, and pulling k3_sec_proxy.h here drags cpu.h and sysregs.h into every platform.h include. I'd include them in the description.
- Functions take the file's module prefix in the tree, so k3_sec_proxy_ rather than mbox_k3_sec_proxy_, and we don't use leading underscore statics anywhere. Also, "k3" now clashes with the SpacemiT K3 platform we just added, ti_sci and ti_sec_proxy would be less ambiguous.
- The typedef'd anonymous structs and uppercase enum typedefs are not how the rest of the tree is written, struct tags please.
- The doxygen blocks are far denser than anything else in src. I'd trim to what the code can't say by itself.
- The "needs to be checked" note on GIC_VERSION can go, J721E has a GIC-500, which is v3.

src/core/init.c
josecm: This goes with the drivers. It breaks tc4dx, s32z270 and rh850-u2a16, whose platform_default_init sets up the pinmux and clocks the console needs. I get why it moved, console_write spins until the console is ready, so the INFO prints the driver does on every read and write would hang the boot in the original order. With the drivers out, the reason goes too.

src/platform/beaglebone-ai64/bb-ai64_desc.c
```diff
@@ -0,0 +1,130 @@
...
+    .ops =
+        {
+            .send = platform_tisci_send,
+            .recv = platform_tisci_recv,
+        },
+    .seq_id_lock = {0, 0},
+    .seq_id = 0U,
+};
+
+void platform_default_init(void)
```
josecm: The whole block from the thread table down leaves with the drivers, same for the driver includes in platform.h. Once TISCI has a real user we'll know what needs to happen here at init, if anything.

src/platform/beaglebone-ai64/bb-ai64_desc.c
josecm: Region 0 spans the OP-TEE secure DDR at 0x9e800000 (24 MB, no-map in the device tree and firewalled). The pool allocates bottom up, so we get there once guests ask for around 490 MB, which a Linux guest will. Can we carve it out, one region below and one above? The R5F IPC pools right after it are plain DRAM as long as no R5 firmware runs, those I'd leave in.

src/platform/beaglebone-ai64/inc/plat/psci.h
josecm: On the power down question from the thread: bao already has a knob for this, rpi4, imx8mp and qemu define PLAT_PSCI_POWERDOWN_NOT_SUPPORTED and the core falls back to a standby suspend, which k3 TF-A implements as wfi. Defining it here is all we need, no core changes and no wfi patch. As pushed, an unassigned core ends in the ERROR loop. The bit 16 state type layout is right for this TF-A btw, k3 doesn't select the extended state id format.

```suggestion
#define PSCI_STATE_TYPE_POWERDOWN (PSCI_STATE_TYPE_BIT | PSCI_STATE_ID_POWERDOWN)

#define PLAT_PSCI_POWERDOWN_NOT_SUPPORTED 1
```

src/platform/beaglebone-ai64/platform.mk
```diff
@@ -0,0 +1,18 @@
+## SPDX-License-Identifier: Apache-2.0
+## Copyright (c) Bao Project and Contributors. All rights reserved.
+
+# Architecture definition
+ARCH:=armv8
+# CPU definition
+CPU:=cortex-a72
+
+#needs to be checked
```
josecm: J721E has a GIC-500, so v3 is right and the note can go.

PR #322, opened 2026-02-24
feat(cores): add cpu-impl regs initialization infrastructure

danielRep:
This pull request introduces a CPU implementation-specific initialization framework, with a concrete implementation for the Cortex-R52 core. It adds a new abstraction layer for per-core initialization, updates the build system to support core-specific files, and extends system register acessors to support new registers required for implementation-specific setup. To use the infrastructure each platform needs to explicitly mention which cores are feature through `PLAT_CPUS` make variable.

* Introduced the `cpu_arch_core_impl_init()` function, allowing per-core implementation-specific initialization to be invoked (MIDR is used to read the partnumber of the running core)
* Added a Cortex-R52-specific initialization routine (`cortex_r52_impl_init`) that configures HACTLR, peripheral region, and flash region system registers

### DIFF
--- a/src/arch/armv8/aarch32/inc/arch/subarch/sysregs.h
+++ b/src/arch/armv8/aarch32/inc/arch/subarch/sysregs.h
@@ -118,6 +118,10 @@ SYSREG_GEN_ACCESSORS_64(icc_sgi1r_el1, 0, c12)
 SYSREG_GEN_ACCESSORS(vsctlr_el2, 4, c2, c0, 0)
 SYSREG_GEN_ACCESSORS(sctlr_el2, 4, c1, c0, 0)
 
+SYSREG_GEN_ACCESSORS(hactlr, 4, c1, c0, 1)
+SYSREG_GEN_ACCESSORS(imp_periphregionr, 0, c15, c0, 0)
+SYSREG_GEN_ACCESSORS(flashifregionr, 0, c15, c0, 1)
+
 #define SYSREG_GEN_GIC_LR(n, crn1, crn2, op2)           \
     SYSREG_GEN_ACCESSORS(ich_lr##n, 4, c12, crn1, op2)  \
     SYSREG_GEN_ACCESSORS(ich_lrc##n, 4, c12, crn2, op2) \
--- a/src/arch/armv8/armv8-r/cpus/core_impl.c
+++ b/src/arch/armv8/armv8-r/cpus/core_impl.c
@@ -6,6 +6,11 @@
 #include <arch/sysregs.h>
 #include <arch/core_impl.h>
 
+#define ARMV8R_R52_PARTNUMBER 0xD13
+
+/* Weak references to CPU implementation init functions */
+extern void cortex_r52_impl_init(void) __attribute__((weak));
+
 void cpu_arch_core_impl_init(void)
 {
     /* Read MIDR to identify core implementation */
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -127,7 +130,7 @@ build_dir:=$(cur_dir)/build/$(PLATFORM)/$(CONFIG)
 bin_dir:=$(cur_dir)/bin/$(PLATFORM)/$(CONFIG)
 directories:=$(build_dir) $(bin_dir)
 
-src_dirs+=$(cpu_arch_dir) $(lib_dir) $(core_dir) $(core_mem_prot_dir) \
+src_dirs+=$(cpu_arch_dir) $(cores_dir) $(lib_dir) $(core_dir) $(core_mem_prot_dir) \
```
josecm: I think this should not be visible from top level Makefile. This should be included in the arch.mk file for armv8.

src/arch/armv8/inc/arch/core_impl.h
```diff
@@ -0,0 +1,11 @@
+/**
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) Bao Project and Contributors. All rights reserved.
+ */
+
+#ifndef ARMV8_CORE_IMPL_H
+#define ARMV8_CORE_IMPL_H
```
josecm:
```suggestion
#ifndef ARCH_CORE_IMPL_H
#define ARCH_CORE_IMPL_H
```

src/arch/armv8/armv8-r/cpus/objects.mk
```diff
@@ -0,0 +1,9 @@
+## SPDX-License-Identifier: Apache-2.0
+## Copyright (c) Bao Project and Contributors. All rights reserved.
+
+cpu-objs-y+=$(ARCH_PROFILE)/cpus/core_impl.o
+
+# CPU-specific implementations based on platform configuration
+ifeq ($(filter cortex-r52,$(PLAT_CPUS)),cortex-r52)
+cpu-objs-y+=$(ARCH_PROFILE)/cpus/cortex_r52.o
+endif
```
josecm: We can't selectively chose the cores we are adding, otherwise we need conditional compilation in `cpu_arch_core_impl_init` switch case.

I like the idea of only compiling the code for the cores we really need. But if so, we need an alternative.

josecm: Also, we should do domething like this so we don't have to change the filter each time we add the initialization for a new core:

```suggestion
cpu-objs-y+=$(foreach cpu, $(PLAT_CPUS), $(ARCH_PROFILE)/cpus/$(cpu).o)
```

src/arch/armv8/armv8-r/cpu.c
```diff
@@ -6,9 +6,12 @@
 #include <cpu.h>
 #include <platform.h>
 #include <arch/gic.h>
+#include <arch/core_impl.h>
 
 void cpu_arch_profile_init(cpuid_t cpuid, paddr_t load_addr)
 {
+    cpu_arch_core_impl_init();
```
josecm: Can't we move this to `cpu_arch_init` so we don't have to replicate this call for the -A and -R profiles?

src/arch/armv8/inc/arch/core_impl.h
danielRep: Fixed here f65b5f1a

Makefile
danielRep: Fixed here 6fb87e74

src/arch/armv8/armv8-r/cpus/objects.mk
danielRep: Fixed here 1ee0eb79

src/arch/armv8/armv8-r/cpu.c
danielRep: Cleaner :+1: Fixed here 6ee6d2dd

src/arch/armv8/armv8-r/cpus/objects.mk
danielRep: Fixed here cfb6933c

PR #325, opened 2026-03-01
fix: correctly check if memory was correctly reserved

josecm:
Previously `mem_reserve_ppool_ppages` returned true if the physical pages to be returned were not part of the pool. This was interpreted by some calling functions as if the memory was succefully reserved. This patch fixes the issue, by changing the return value of this function to false is the pages were not part of the pool and hencefort not allocated and creates explicit functions to keep track if memory was correctly reserved.

### DIFF
--- a/src/core/mem.c
+++ b/src/core/mem.c
@@ -156,22 +156,16 @@ static bool mem_are_ppages_reserved_in_pool(struct page_pool* ppool, struct ppag
 
 static bool mem_reserve_ppool_ppages(struct page_pool* pool, struct ppages* ppages)
 {
+    bool reserved = false;
     bool is_in_rgn = mem_ppages_in_pool(pool, ppages);
-    if (!is_in_rgn) {
-        return false;
+    if (is_in_rgn && !mem_are_ppages_reserved_in_pool(pool, ppages)) {
+        size_t pageoff = NUM_PAGES(ppages->base - pool->base);
+        bitmap_set_consecutive(pool->bitmap, pageoff, ppages->num_pages);
+        pool->free -= ppages->num_pages;
+        reserved = true;
     }
 
-    size_t pageoff = NUM_PAGES(ppages->base - pool->base);
-
-    bool was_free = true;
-    if (mem_are_ppages_reserved_in_pool(pool, ppages)) {
-        was_free = false;
-    }
-
-    bitmap_set_consecutive(pool->bitmap, pageoff, ppages->num_pages);
-    pool->free -= ppages->num_pages;
-
-    return is_in_rgn && was_free;
+    return reserved;
 }
 
 void* mem_alloc_page(size_t num_pages, as_sec_t sec, bool phys_aligned)


src/core/mem.c
```diff
@@ -342,8 +419,8 @@ static bool mem_reserve_physical_memory(struct page_pool* pool)
             continue;
         }
 
-        if (!mem_reserve_ppool_ppages(pool, &ppages)) {
-            return false;
+        if (mem_reserve_ppool_ppages(pool, &ppages)) {
```
DavidMCerdeira:
```suggestion
        if (mem_reserve_ppool_ppages(pool, &ppages) && !vm_cfg->image.reserved) {
```
Since this function is called multiple times, we should skip trying to reserve if reserved already
the same applies for the other reserve operations

src/core/mem.c
```diff
@@ -356,8 +433,8 @@ static bool mem_reserve_physical_memory(struct page_pool* pool)
             if (reg->place_phys) {
```
DavidMCerdeira:
```suggestion
            if (reg->place_phys && !reg->reserved) {
```

src/core/mem.c
```diff
@@ -368,14 +445,12 @@ static bool mem_reserve_physical_memory(struct page_pool* pool)
         if (shmem->place_phys) {
```
DavidMCerdeira:
```suggestion
        if (shmem->place_phys && !shmem->reserved) {
```

src/core/mem.c
miguelafsilva5: I was testing this PR, and the codes returns false here. I have a single VM with a single memory region. At this point, the memory is not reserved. I'm assuming that the memory will only be reserved later when the VM image is installed.

src/core/mem.c
```diff
@@ -316,18 +320,91 @@ static bool mem_vm_img_in_phys_rgn(struct vm_config* vm_config)
...
+    if (DEFINED(MEM_NON_UNIFIED)) {
+        mem_hyp_image_no_load_reserved = false;
     }
 
+    for (size_t i = 0; i < config.vmlist_size; i++) {
+        struct vm_config* vm_cfg = &config.vmlist[i];
+
+        // If the vm image is part of a statically allocated region of the same vm, we defer the
+        // reservation to the image itself and marked the image reserved.
+        if (mem_vm_img_in_phys_rgn(vm_cfg)) {
```
DavidMCerdeira, line 335:
```suggestion
        if (mem_vm_img_in_phys_rgn(vm_cfg) || !vm_cfg->image.separately_loaded) {
```
If the vm image is builtin it is reserved as part of hypervisor data/image. The comment above should also address this
I believe this addresses @miguelafsilva5 error

miguelafsilva5: It should be `!vm_cfg->image.separately_loaded`. But that fixes the problem

josecm: I don't understand what's the point of this. It doesn't matter if it was separately loaded or not. If the image is inside one of the regions, we can considered it as reserved and defer the reservation to the region itself. Am I missing something?

src/core/mem.c
josecm: Agree. I'll switch the order of the conditions to short-circuit the call to `mem_reserve_ppool_ppages` if the image was already reserved.

src/core/mem.c
josecm: @miguelafsilva5 what do you mean returns false? If you are referring to an MPU platform place_phys should always be true for all regions, as marked per `config_mem_prot_init`

src/core/mem.c
josecm: Addressed in latest force push

src/core/mem.c
josecm: Addressed in latest force push.

src/core/mem.c
DavidMCerdeira: Won't this return false if the memory, in this case the vm image, is already reserved?
```c
    bool was_free = true;
    if (mem_are_ppages_reserved_in_pool(pool, ppages)) {
        was_free = false;
    }

    bitmap_set_consecutive(pool->bitmap, pageoff, ppages->num_pages);
    pool->free -= ppages->num_pages;

    return is_in_rgn && was_free;
```

src/core/mem.c
DavidMCerdeira: We'll continue this dicussion in https://github.com/bao-project/bao-hypervisor/pull/325#discussion_r2872070693

src/core/mem.c
josecm: I am fixing this next. The function itself is wrong because we proceed to reserve the pages with `bitmap_set_consecutive` even if some of these bits were previously reserved (i.e., was_free is false). Do you agree?

See #327

josecm: @miguelafsilva5 sorry, I now understand what you meant. If the image was not separately loaded (i.e., its builtin in the hypervisor) it is already reserved when we mark the hypervisor image as reserved. Fixed.

PR #326, opened 2026-03-01
ref: explicitly choose section on mem_map_cpy

josecm:


miguelafsilva5:
> > @josecm we can remove the `typedef enum AS_SEC` since we are no longer using it.
> 
> You mean like this?
> 
> ```
> typedef enum {
>     .....
> } as_sec_t;
> ```

Yeah. In the file `types.h`

PR #327, opened 2026-03-02
Fix page-pool reservation so failure has no side effects

josecm:
This PR fixes a bug in mem_reserve_ppool_ppages() where the page pool state could be mutated even when the reservation failed.

Depends on #325


src/core/mem.c
```diff
@@ -127,7 +127,7 @@ bool pp_alloc(struct page_pool* pool, size_t num_pages, bool aligned, struct ppa
 
     return ok;
 }
-
+a
```
DavidMCerdeira: extra

josecm: Sorry :sweat_smile:. Added that on a rebase. Fixed

PR #329, opened 2026-03-03
fix(psci): per-plat smccc version support

DavidMCerdeira:
With S32G3 platform we misinterpreted the VENDOR_START PSCI_FEATURES
argument. The correct flag is SMCC_VERSION. This flag seems to be needed
for vendor specific calls that depend on a high-enough smccc version.
Addtionally, we make this flag per-platform, since currently supported
platforms except for S32G3 don't need this feature.


DavidMCerdeira:
Working on RPi4

src/arch/armv8/psci.c
```diff
@@ -170,9 +171,12 @@ static int32_t psci_features_handler(uint32_t feature_id)
         case PSCI_AFFINITY_INFO_SMC32:
         case PSCI_AFFINITY_INFO_SMC64:
         case PSCI_FEATURES:
-        case PSCI_1_1_RESET_TYPE_VENDOR_START:
+#ifdef PLAT_PSCI_SUPPORT_SMCCC_VERSION
+        case SMCCC_VERSION:
```
josecm, line 175: It seems that SMCC_VERSION FID is part of "ARM Architecture Calls" so I don't understand why we are handling it in the PSCI handler.

DavidMCerdeira: It surprised me as well, but it seems that's the flow Linux uses.
Snippet for context
```c
static void __init psci_init_smccc(void)
{
	u32 ver = ARM_SMCCC_VERSION_1_0;
	int feature;

	feature = psci_features(ARM_SMCCC_VERSION_FUNC_ID); // <----- Here

	if (feature != PSCI_RET_NOT_SUPPORTED) {
		u32 ret;
		ret = invoke_psci_fn(ARM_SMCCC_VERSION_FUNC_ID, 0, 0, 0);
		if (ret >= ARM_SMCCC_VERSION_1_1) {
			arm_smccc_version_init(ret, psci_conduit);
			ver = ret;
		}
	}

	/*
	 * Conveniently, the SMCCC and PSCI versions are encoded the
	 * same way. No, this isn't accidental.
	 */
	pr_info("SMC Calling Convention v%d.%d\n",
		PSCI_VERSION_MAJOR(ver), PSCI_VERSION_MINOR(ver));

}

```

josecm: It's weird for Linux testing for a PSCI feature which is not part of PSCI. Neverthless, it does seem it is required.

Issue #162, opened 2024-07-22
Linux RT stalls after boot


josecm:
@Sunt-git and @LSVulnerable detected issues (#46 and #161) in which Linux with the RT patch stalls after initialization.

@LSVulnerable noted that "when a pl011 interrupt inject, maintenance interrupt occur and then guestos have no more response".

PR #334, opened 2026-03-19
Fix issues with vGIC LR management

josecm:
This PR introduces a set of fixes related to issues with vGIC List Register management. Among others, it addresses issues #162.

### DIFF
--- a/src/arch/armv8/vgic.c
+++ b/src/arch/armv8/vgic.c
@@ -180,7 +180,12 @@ static void vgic_route(struct vcpu* vcpu, struct vgic_int* interrupt)
         vgic_add_lr(vcpu, interrupt);
     }
 
-    if (!interrupt->in_lr && vgic_int_has_other_target(vcpu, interrupt)) {
+    /**
+     * An active interrupt cannot be forwarded to another CPU — ownership
+     * cannot be yielded while the interrupt is active, so any IPI would
+     * be silently ignored by the recipient.
+     */
+    if (!interrupt->in_lr && !(interrupt->state & ACT) && vgic_int_has_other_target(vcpu, interrupt)) {
         struct cpu_msg msg = {
             (uint32_t)VGIC_IPI_ID,
             VGIC_ROUTE,
--- a/src/arch/armv8/vgic.c
+++ b/src/arch/armv8/vgic.c
@@ -146,7 +146,16 @@ void vgic_send_sgi_msg(struct vcpu* vcpu, cpumap_t pcpu_mask, irqid_t int_id)
 
 static void vgic_route(struct vcpu* vcpu, struct vgic_int* interrupt)
 {
-    if ((interrupt->state == INV) || !interrupt->enabled) {
+    if (interrupt->state == INV) {
+        return;
+    }
+
+    /**
+     * A disabled interrupt that is still active must be placed in an LR so
+     * the guest can EOI it directly, avoiding an unnecessary LRENP maintenance
+     * interrupt trap.
+     */
+    if (!interrupt->enabled && !(interrupt->state & ACT)) {
         return;
     }
 
@@ -315,7 +324,7 @@ bool vgic_add_lr(struct vcpu* vcpu, struct vgic_int* interrupt)
 {
     bool ret = false;
 
-    if (!interrupt->enabled || interrupt->in_lr) {
+    if ((!interrupt->enabled && !(interrupt->state & ACT)) || interrupt->in_lr) {
         return ret;
     }
 
--- a/src/arch/armv8/vgic.c
+++ b/src/arch/armv8/vgic.c
@@ -402,7 +402,7 @@ bool vgic_add_lr(struct vcpu* vcpu, struct vgic_int* interrupt)
             }
         }
 
-        if (pend_found > 1) {
+        if (pend_found > 1 && pend_ind >= 0) {
             lr_ind = pend_ind;
         } else {
             lr_ind = act_ind;
--- a/src/arch/armv8/vgic.c
+++ b/src/arch/armv8/vgic.c
@@ -318,6 +318,16 @@ bool vgic_remove_lr(struct vcpu* vcpu, struct vgic_int* interrupt)
     return ret;
 }
 
+/* Must be called holding vgic_spilled_lock */
+static void vgic_remove_spilled(struct vcpu* vcpu, struct vgic_int* interrupt)
+{
+    struct list* spilled_list = gic_is_priv(interrupt->id)
+        ? &vcpu->arch.vgic_spilled
+        : &vcpu->vm->arch.vgic_spilled;
+    list_rm(spilled_list, &interrupt->node);
+    interrupt->in_spilled = false;
+}
+
 static void vgic_spill_lr(struct vcpu* vcpu, size_t lr_ind)
 {
     gic_lr_t lr = (gic_lr_t)gich_read_lr(lr_ind);
@@ -392,6 +402,11 @@ bool vgic_add_lr(struct vcpu* vcpu, struct vgic_int* interrupt)
     }
 
     if (lr_ind >= 0) {
+        if (interrupt->in_spilled) {
+            spin_lock(&vcpu->vm->arch.vgic_spilled_lock);
+            vgic_remove_spilled(vcpu, interrupt);
+            spin_unlock(&vcpu->vm->arch.vgic_spilled_lock);
+        }
         vgic_write_lr(vcpu, interrupt, (size_t)lr_ind);
         ret = true;
     } else {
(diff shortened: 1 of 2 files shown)


irenebus (commented):
Hi, @josecm. This successfully fixes the issue described in #371 . Thanks for pointing me to it!

josecm:
> Hi, @josecm. This successfully fixes the issue described in #371 . Thanks for pointing me to it!

Thanks for your feedback @irenebus. Glad your issue is solved!

rabara:
Hi @josecm, 

Any plan to merge this PR? 
On Agilex5 I have same problem as described in #371 and these change are needed to make UART work with Linux.

josecm:
Hey @rabara! Having you aknowledge this PR fixes issues on your side also is already another push to merging this. Last week I asked @DavidMCerdeira that he makes this a high priority. So I am hoping we merge at most in a couple weeks. Hopefully, less.

src/arch/armv8/vgic.c
```diff
@@ -1118,19 +1163,28 @@ static void vgic_refill_lrs(struct vcpu* vcpu, bool npie)
...
+        }
+    }
+    spin_unlock(&vcpu->vm->arch.vgic_spilled_lock);
+
+    if (interrupt != NULL) {
+        interrupt->state &= (uint8_t)~ACT;
+        if (vgic_int_is_hw(interrupt)) {
+            gic_set_act(interrupt->id, false);
+        } else {
+            if (interrupt->state & PEND) {
```
DavidMCerdeira: Given that `interrupt->state` is updated in line 1182:
```c
interrupt->state &= (uint8_t)~ACT;
```

When will this condition be true?

src/arch/armv8/vgic.c
```diff
@@ -1118,19 +1163,28 @@ static void vgic_refill_lrs(struct vcpu* vcpu, bool npie)
...
+            vgic_remove_spilled(vcpu, interrupt);
+        } else {
+            spin_unlock(&interrupt->lock);
+            interrupt = NULL;
+        }
+    }
+    spin_unlock(&vcpu->vm->arch.vgic_spilled_lock);
+
+    if (interrupt != NULL) {
+        interrupt->state &= (uint8_t)~ACT;
```
DavidMCerdeira: Shouldn't this be done with a lock on the interrupt?

src/arch/armv8/vgic.c
```diff
@@ -168,6 +199,15 @@ static void vgic_route(struct vcpu* vcpu, struct vgic_int* interrupt)
             }
         }
     }
+
+    /**
+     * If the interrupt was not placed in an LR or forwarded to another CPU,
+     * track it in the spilled list so it gets injected when an LR slot
+     * becomes available.
+     */
+    if (!interrupt->in_lr && !interrupt->in_spilled) {
```
DavidMCerdeira: This check is already performed in `vgic_add_spilled`, why check it outside as well?
If it's for performance reasons a comment should state that.

The comment explaining the current check should be moved to `vgic_add_spilled`.

src/arch/armv8/vgic.c
```diff
@@ -283,18 +330,14 @@ bool vgic_remove_lr(struct vcpu* vcpu, struct vgic_int* interrupt)
...
-    if (gic_is_priv(interrupt->id)) {
-        spilled_list = &vcpu->arch.vgic_spilled;
-    } else {
-        spilled_list = &vcpu->vm->arch.vgic_spilled;
-    }
-    list_push(spilled_list, (node_t*)interrupt);
-    spin_unlock(&vcpu->vm->arch.vgic_spilled_lock);
-    gich_set_hcr(gich_get_hcr() | GICH_HCR_NPIE_BIT);
+    struct list* spilled_list = gic_is_priv(interrupt->id)
+        ? &vcpu->arch.vgic_spilled
```
DavidMCerdeira: There's something I'd like to make clear here: there's only a lock for the SPI spilled list, `vcpu.vm.arch.vgic_spilled`, but no lock for the private interrupts list, `vcpu.arch.gic_spilled`

It might be worth explaining this in a comment

DavidMCerdeira (commented):
In commit 2cc772bcb7ccf5a20c5f32197ae335f4bd3c6423:
> HW PENDACT is impossible (the physical GIC prevents re-assertion
while active), so state & PEND is always 0 for HW.

I don't think this is true, can you explain a bit more?

In commit db718b874a5832116312b185ea9d6583c96133d2
> When all LRs are full and pend_found > 1 but all pending LRs have higher
    priority than the new interrupt (pend_ind == -1), the spill selection set
    lr_ind = -1 and skipped eviction entirely, even if a lower-priority active
    LR was available to spill.
    Add pend_ind >= 0 to the condition so the selection falls through to
    act_ind when no eligible pending candidate exists.

I think it should be:
> When all LRs are full, the spill selection set lr_ind = -1 and skipped eviction entirely, even if a lower-priority active LR was available to spill.
    Add pend_ind >= 0 to the condition so the selection falls through to
    act_ind when no eligible pending candidate exists.

DavidMCerdeira:
I finally got some time to review this.

In addition to my other comments, @josecm please update the branch and fix the issues identified by the gitlint and the format check.

src/arch/armv8/vgic.c
josecm: I admit the lock/unlock sequence is a bit convoluted. But if you look carefully into it you can see it is: we lock the interrupt, if we get ownership of the irq we don't unlock it, otherwise we unlock it and set interrupt to null. We only get in this code in the former case, and unlock it at the end.

src/arch/armv8/vgic.c
josecm: The two checks have different purposes. The one in vgic_route is a routing decision: the interrupt was not placed in an LR and was not handed to another CPU, so this core keeps tracking it in the spilled list. The one inside vgic_add_spilled is a correctness guard so the function is safe to call from any context, since an interrupt that is in an LR or already listed must never be pushed again.

That said, while going through this, I noticed the routing condition did not actually match its comment: the forwarded case was never checked, so an interrupt forwarded to another CPU was still added to the spilled list after ownership was yielded, and a refill on a non-target vcpu could then pick it up. Fixed in 5259930e: vgic_route now skips the fallback for forwarded interrupts, the comment moved into vgic_add_spilled, and NPIE is only armed when an interrupt is actually queued.

src/arch/armv8/vgic.c
josecm: There is actually no need for a lock on the private list: it is only ever touched by the cpu hosting the vcpu those interrupts are private to. Accesses from other cores to another vcpu's private interrupts go through the vgic IPI, so they also end up running on the hosting cpu.

josecm:
> In commit 2cc772bcb7ccf5a20c5f32197ae335f4bd3c6423:
> > HW PENDACT is impossible (the physical GIC prevents re-assertion while active), so state & PEND is always 0 for HW.
>
> I don't think this is true, can you explain a bit more?

You are right, this is not true in general. It only holds for the physical delivery path: since the hypervisor only does the priority drop and leaves deactivation to the guest, the physical interrupt stays active while the guest handles it, and the GIC does not signal an interrupt that is still active. A re-assertion just parks as physical pending, so the hypervisor never gets to inject pend on top of a virtual act.

But the guest can still produce pend and act on a hw interrupt through the virtual distributor: an ISPENDR write while the interrupt is active, or an ISACTIVER write while it is pending, goes through vgic_int_update_pend/vgic_int_update_act and ends up with both bits set. That is actually why vgic_write_lr already had the special case writing only ACT for a PENDACT hw interrupt, as an LR with the HW bit set cannot hold the pending and active state.

The code change itself does not depend on the wrong claim: the hw branch needs no strip because the LR is written with only ACT and the pending state is tracked by the physical GIC, which will not re-deliver it while the interrupt is disabled given the physical enable mirrors the virtual one. I will reword the commit message accordingly when I update the branch.

josecm:
> I think it should be:
> > When all LRs are full, the spill selection set lr_ind = -1 and skipped eviction entirely, even if a lower-priority active LR was available to spill.

That would overstate the bug. With all LRs full the old code only skipped eviction in one specific case: more than one LR holding a pending interrupt, but none of them an eligible victim. If an eligible pending LR existed it was spilled, and with at most one pending LR the selection already fell through to the active candidate. So the "more than one pending, none eligible" condition is what the bug actually requires and I would rather keep it in the message.

I agree the current phrasing leans too much on code internals though. When I update the branch I will reword it without the variable names, to something like: "When all LRs are full and more than one of them holds a pending interrupt, but none of those is an eligible victim (none has lower priority than the new interrupt), the spill selection skipped eviction entirely, even if a lower priority active LR was available to spill."

josecm:
While building a follow-up fix on top of this branch I found a lock order inversion that can deadlock the whole VM: the spilled list walkers take the VM wide vgic_spilled_lock and then the candidate interrupt lock, while the emulation paths hold the interrupt lock and take the spilled lock inside vgic_add_lr and vgic_add_spilled. One core in vgic_refill_lrs holding the spilled lock and spinning on an interrupt lock, and another core writing a register of that same spilled interrupt holding its lock and spinning on the spilled lock, wait on each other forever.

The inverted order in vgic_refill_lrs actually predates this branch, but the LRENP commit in this series extended it to vgic_eoir_highest_spilled_active, so I fixed both here: 430410b8 restructures the walkers to pick the candidate under the spilled lock, release it, and revalidate under both locks taken in the canonical order.

DavidMCerdeira (dismissed):

PR #338, opened 2026-03-27
armv8: add MPIDR table support and extend affinity handling for newer CPUs

joaopeixoto13:
This patch introduces support for newer Arm CPU topologies (e.g. `Cortex-A55`, `A76`, `A78AE`) where CPU identification is not strictly encoded in Aff0 (`MPIDR_EL1[7:0]`) but may instead rely on higher affinity levels such as `Aff1` (`MPIDR_EL1[15:8]`). On these platforms, `Aff0` typically represents thread ID (often 0x0 for single-threaded cores like Cortex-A55), making previous assumptions about CPU ID encoding invalid.

Key changes:
- Add `mpidr_table` support to allow explicit CPU ID <-> MPIDR mappings. This is required for SoCs with non-linear CPU layouts (e.g. Jetson Orin Nano), where CPU IDs do not map sequentially within clusters.
- Introduce `mpidr_to_cpu_id()` in early boot to derive logical CPU IDs from MPIDR values using the platform-provided table when available.
- Extend MPIDR affinity mask handling from 2 to 3 levels (Aff0–Aff2).
- Update GICv3 SGI routing to include Aff2, enabling correct interrupt targeting on multi-level affinity systems.
- Refactor `platform_arch_cpuid_to_mpidr`() to:
  * Prefer mpidr_table when present
  * Fall back to cluster topology
  * Preserve architectural MPIDR bits (RES1, U bit)
- Add required structure fields and offset definitions for the new mpidr_table.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -221,6 +221,9 @@ ifeq ($(arch_mem_prot),mpu)
 	build_macros+=-DMEM_PROT_MPU
 endif
 ifeq ($(plat_mem),non_unified)
+	ifeq ($(ARCH),aarch64)
+		$(error AArch64 with non_unified memory is not supported)
+	endif
 	build_macros+=-DMEM_NON_UNIFIED
 endif
 ifeq ($(phys_irqs_only),y)
--- a/src/arch/armv8/aarch64/boot.S
+++ b/src/arch/armv8/aarch64/boot.S
@@ -238,6 +238,7 @@ set_loop:
  * x5 - current entry
  * x6 - index in mpidr_table->table[]
  * x7 - original MPIDR (saved for MT bit check)
+ * x8 - temporary scratch register
  */
 .global mpidr_to_cpu_id
 mpidr_to_cpu_id:
--- a/src/arch/armv8/aarch64/boot.S
+++ b/src/arch/armv8/aarch64/boot.S
@@ -47,6 +47,7 @@ _reset_handler:
 
 	mrs  x0, MPIDR_EL1
 	adrp x1, _image_start
+    bl mpidr_to_cpu_id
 
 	adr x2, img_addr
     str x1, [x2] // store image load address in img_addr
@@ -66,7 +67,9 @@ _reset_handler:
 	 */
 
 	lsr x3, x0, #8
-	and x3, x3, 0xff
+    and x0, x0, #0xFF
+	and x3, x3, #0xFF
+
 	adr x4, platform
 	ldr x4, [x4, PLAT_ARCH_OFF+PLAT_ARCH_CLUSTERS_OFF+PLAT_CLUSTERS_CORES_NUM_OFF]
 	ldr x5, =BAO_VAS_BASE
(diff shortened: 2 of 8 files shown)


src/arch/armv8/aarch64/boot.S
```diff
@@ -224,3 +227,71 @@ set_loop:
 	cmp x5, x3 // last way reached yet?
 	ble way_loop // if not, iterate way_loop
 	ret
+
+/*
+ * Registers
```
DavidMCerdeira, line 232: x8 is used but not documented

src/arch/armv8/aarch64/boot.S
```diff
@@ -224,3 +227,71 @@ set_loop:
...
+    mov x7, x0                        /* save original MPIDR for MT check */
+    adr x2, platform
+    ldr x2, [x2, PLAT_ARCH_OFF+PLAT_ARCH_MPIDR_TABLE_OFF+PLAT_MPIDR_TABLE_OFF]
+    cbz x2, no_table
+    /* Relocate pointer */
+    ldr x5, =BAO_VAS_BASE
+    sub x2, x2, x5
+    add x2, x2, x1
+    mov x4, x2
+    adr x3, platform
```
DavidMCerdeira, line 254: For non-unified platforms this will not work correctly as the data is not yet loaded into the final address
In RH850 I had a similar issue and I've calculated the table in the binary before data load

src/arch/armv8/aarch64/boot.S
```diff
@@ -66,7 +67,9 @@ _reset_handler:
 	 */
 
 	lsr x3, x0, #8
-	and x3, x3, 0xff
+    and x0, x0, #0xFF
+	and x3, x3, #0xFF
```
DavidMCerdeira, line 71: can we safely discard aff0?

src/arch/armv8/platform.c
```diff
@@ -8,37 +8,71 @@
...
+        mpidr = (cpuid << 8);
     } else {
-        /**
-         * No cluster information in configuration. Assume a singl cluster.
-         */
         mpidr = cpuid;
     }
 
+done:
+
```
DavidMCerdeira: extra line

src/arch/armv8/aarch64/boot.S
joaopeixoto13: > can we safely discard aff0?

In both `MT=0` and `MT=1` cases, we always construct the final output in `x0` (check `mpidr_to_cpu_id`) as:
- `[15:8] `= cluster ID 
- `[7:0]` = core ID

The difference is only in where we source the values from in the MPIDR:
- `MT=0`: core comes from `aff0`, cluster from `aff1`
- `MT=1`: core comes from `aff1`, cluster from `aff2`

Regarding `aff0` when `MT=1` is set, `aff0` is not meaningful since our CPUs do not support multithreading (`aff0` identifies individual threads within a multithreaded core, but in our case this level is always zero). As a result, it can be safely ignored in this case.

src/arch/armv8/platform.c
joaopeixoto13: Fixed [here](https://github.com/bao-project/bao-hypervisor/pull/338/changes/dfbf6e0230de19e2f147c29794b9a4ed255d3bd4)

src/arch/armv8/aarch64/boot.S
joaopeixoto13: > x8 is used but not documented

Fixed [here](https://github.com/bao-project/bao-hypervisor/pull/338/changes/af644f235028bda56dd2ce7165b795a5c10cb3b5)

src/arch/armv8/aarch64/boot.S
joaopeixoto13: > For non-unified platforms this will not work correctly as the data is not yet loaded into the final address In RH850 I had a similar issue and I've calculated the table in the binary before data load

Fixed [here](https://github.com/bao-project/bao-hypervisor/pull/338/changes/a6fa3f1425e81d92de37b69dfd7c688aa4e51394).

**Note**: Since there are no known platforms combining AArch64 with a non-unified memory model, and this PR introduces an explicit limitation in that regard, I added a `#error` directive to prevent compilation for this unsupported configuration. Let me know what you think.

src/arch/armv8/aarch64/boot.S
DavidMCerdeira: I understand know
thanks!

DavidMCerdeira (dismissed):

src/arch/armv8/platform.c
```diff
@@ -8,37 +8,70 @@
...
+    if (has_mt) {
+        mpidr = (cpuid << 8);
     } else {
-        /**
-         * No cluster information in configuration. Assume a singl cluster.
-         */
         mpidr = cpuid;
     }
 
+done:
```
josecm: @joaopeixoto13 our coding sytle does not allow for goto labels. Please re-factor to avoid this.

Makefile
```diff
@@ -221,6 +221,9 @@ ifeq ($(arch_mem_prot),mpu)
 	build_macros+=-DMEM_PROT_MPU
 endif
 ifeq ($(plat_mem),non_unified)
+	ifeq ($(ARCH),aarch64)
+		$(error AArch64 with non_unified memory is not supported)
+	endif
```
josecm, line 226: This makes sense, but I  don't see why it is part of this PR. Can you explain?

josecm (changes requested):

Makefile
joaopeixoto13: Because the MPIDR table handling performs manual address relocation:
```
    sub x2, x2, x5
    add x2, x2, x1
    mov x4, x2
    adr x3, platform
```

This logic assumes that the data referenced by the table pointer is
already loaded into its final address space. On non_unified platforms,
this is not guaranteed, as data may not yet be relocated or available
at the expected runtime address during early boot.

As a result, the relocation math becomes invalid and the table access
may point to incorrect memory.

src/arch/armv8/platform.c
joaopeixoto13: Fixed [here](https://github.com/bao-project/bao-hypervisor/pull/338/changes/960b3417dc9d31f80355ae2b19c75695ccb94288)

joaopeixoto13:
@josecm I just cleaned up the git history and rebased the branch onto main.
If everything looks good on your side, it's ready to be merged into the main branch.

Makefile
josecm: I understand the issue. However, in the future if you find ourselves supporting platforms featuring for example Cortex-R82 this may be an issue.

josecm (approved):

DavidMCerdeira (approved):

DavidMCerdeira (approved):

PR #349, opened 2026-04-07
ref: remove unused struct addr_space_arch from MPU arch headers

josecm:
`struct addr_space_arch` was defined in `src/arch/tricore/inc/arch/mem.h`
and `src/arch/rh850/inc/arch/mem.h` but is never referenced anywhere in
the codebase. Remove both dead definitions.


danielRep (approved):

DavidMCerdeira (approved):
Please rebase @josecm

danielRep (approved):

PR #351, opened 2026-04-09
RISC-V Vector Extension Support

ninolomata:
Adds optional support for exposing the RISC-V vector extension to guests.

This change:
- adds `sstatus.VS` field definitions
- marks guest vector state as enabled during vCPU reset by setting `VS=Dirty`
  in both `sstatus` and `vsstatus`
- adds a `CPU_EXT_V` platform configuration macro to control vector extension exposure.


danielRep (approved):
LGTM. Spec for anyone reviewing too. https://docs.riscv.org/reference/isa/extensions/vector/_attachments/riscv-v-spec.pdf

Can you rebase @ninolomata ?

src/platform/qemu-riscv64-virt/inc/plat/platform.h
```diff
@@ -9,6 +9,7 @@
 #include <drivers/sbi_uart.h>
 
 #define CPU_EXT_SSTC 1
+#define CPU_EXT_V    0
```
josecm, line 12: no need to set the extension to 0, if it is not part of the platform

josecm (approved):

PR #352, opened 2026-04-09
Add Zicboz and Zicbom cache maintenance support

ninolomata:
## Summary

Add RISC-V cache-maintenance support based on the `Zicbom` and `Zicboz`
extensions.

This change:
- adds cache-block operation helpers for `cbo.clean`, `cbo.flush`, and
  `cbo.inval`
- implements `cache_flush_range()` using `Zicbom` cache-block flushes when the
  extension is enabled
- adds the missing `henvcfg` field definitions needed to control cache-block
  management exposure
- programs `henvcfg` during hypervisor initialization to expose `Zicboz` and
  `Zicbom` only when the platform is configured to support them
- clears `senvcfg` during vCPU reset so guest-visible environment configuration
  starts from a known state
- enables `Zicbom` in the RISC-V build flags and advertises `Zicboz`/`Zicbom`
  in the `qemu-riscv32-virt` and `qemu-riscv64-virt` platform configurations

## Implementation

The PR adds RISC-V cache-block helper wrappers in the cache header and uses
them in `cache_flush_range()` when `CPU_EXT_ZICBOM` is enabled. A weak fallback
implementation is preserved for platforms that do not provide cache maintenance.

On the virtualization side, `henvcfg` is reset during `vmm_arch_init()` and
then updated to enable:
- `HENVCFG_CBZE` for `Zicboz`
- `HENVCFG_CBCFE | HENVCFG_CBIE_FLUSH` for `Zicbom`

The configuration is verified through readback so unsupported WARL fields are
detected early.

The patch also adds the missing CSR definitions and resets `senvcfg` during
vCPU initialization to keep guest-visible state deterministic.

## Requirements to run

Running this support requires coordinated changes across Bao, the guest DT, and
QEMU.

### Bao configuration

Enable cache-maintenance support in the Bao platform configuration for:
- `qemu-riscv32-virt`
- `qemu-riscv64-virt`

Set:
- `CPU_EXT_ZICBOM = 1`
- `CPU_EXT_ZICBOZ = 1`

### Linux / Device Tree

The guest DT must expose cache-block management support by adding:
- `riscv,cbom-block-size = <64>;`
- `riscv,cboz-block-size = <64>;`

It must also advertise the extensions in `riscv,isa`.

For RV32:
- `riscv,isa = "rv32imafdch_zicbom_zicboz_zicntr_zicsr_zifencei_sstc";`

For RV64:
[...]

PR #354, opened 2026-04-09
feat(ssnpm): Add support for pointer masking

ninolomata:
## Summary

Add initial RISC-V `Ssnpm` support for programming the guest-visible PMM mode.

This change:
- adds the `HSTATUS_HUPMM` field definitions
- programs the guest-visible PMM mode in `henvcfg` during hypervisor init
  when `CPU_EXT_SSNPM` is enabled
- adds platform configuration knobs for:
  - `CPU_EXT_SSNPM`
  - `CPU_EXT_SSNPM_PMM_MODE`

## Dependency

This PR depends on:
- https://github.com/bao-project/bao-hypervisor/pull/352

It is intended to be reviewed and merged on top of that series, since it builds
on the related RISC-V extension plumbing already introduced there.

## Implementation

The patch:
- adds `HSTATUS_HUPMM` field definitions in the RISC-V CSR header
- extends `vmm_arch_init()` to configure PMM mode when `CPU_EXT_SSNPM` is enabled
- validates the selected mode at build time and rejects unsupported cases
- adds platform macros to control whether `Ssnpm` is enabled and which PMM mode
  Bao should advertise

The following PMM modes are supported:
- `HENVCFG_PMM_DISABLED`
- `HENVCFG_PMM_PMLEN_7`
- `HENVCFG_PMM_PMLEN_16`

On RV32, the build fails explicitly since `Ssnpm` is not available there.

## Requirements to run

Running this support requires coordinated changes across Bao, the guest DT, and
QEMU.

### Bao configuration

Enable cache-maintenance support in the Bao platform configuration for:
- `qemu-riscv32-virt`
- `qemu-riscv64-virt`

Set:
- `CPU_EXT_SSNPM = 1`
- set the mode (.e.g, `CPU_EXT_SSNPM_PMM_M = HENVCFG_PMM_PMLEN_7`)

### Linux / Device Tree

It must also advertise the extensions in `riscv,isa`.

For RV32:
- `riscv,isa = "rv32imafdch_ssnpm_zicbom_zicboz_zicbop_zicntr_zicsr_zifencei_sstc";`

For RV64:
- `riscv,isa = "rv64imafdch_ssnpm_zicbom_zicboz_zicbop_zicntr_zicsr_zifencei_sstc";`

### QEMU

A QEMU version with `ssnpm` support is required, set `ssnpm=true` on the `cpu.

- RV32 QEMU:
  - use `https://github.com/ninolomata/qemu-bao/tree/riscv-rv32-henvcfg-stateen-fixes`
- RV64 QEMU:
  - use the same version currently used in `bao-demos` (`QEMU 10.0.2`)


src/arch/riscv/vmm.c
```diff
@@ -81,6 +81,46 @@ void vmm_arch_init()
         }
     }
 
+#if CPU_HAS_EXTENSION(CPU_EXT_SSNPM)
+#if defined(RV32)
+#error "Ssnpm extension is not available for RV32. Please disable CPU_EXT_SSNPM."
+#elif !defined(CPU_EXT_SSNPM_PMM_MODE)
+#error "Ssnpm extension PMM mode is not defined. Please define CPU_EXT_SSNPM_PMM_MODE."
+#else
```
josecm, line 89: I don't think these checks should be here. What I'd suggest is to have a private function vmm_arch_config_ssmpm and then an if (CPU_HAS_EXT(...)) vmm_arch_config_ssmpm().

src/arch/riscv/vmm.c
```diff
@@ -81,6 +81,46 @@ void vmm_arch_init()
...
+     * Program the guest-visible Ssnpm PMM mode if the hypervisor was
+     * configured to use it (via the CPU_EXT_SSNPM_PMM_MODE macro).
+     * Otherwise, henvcfg keeps the reset value written above.
+     */
+    uint64_t pmm_val = 0;
+
+    switch (CPU_EXT_SSNPM_PMM_MODE) {
+        case HENVCFG_PMM_DISABLED:
+            if (cpu_is_master()) {
+                WARNING("Ssnpm extension is enabled but PMM mode is set to disabled.\r\n");
```
josecm, line 100: Should actually be "extension is present". But I don't believe this is worth it. I don't think we should allow the configruation in this way. Either we: (i) simply disable support for the extension for now, or (ii) allow to set this in the config in a config global .arch field (i believe we have to add this) (e.g., .arch.npm_mode) or (ii) allow to set the mode per vm, in the a vm config arch field (e.g., arch.npm_mode field)

src/arch/riscv/vmm.c
```diff
@@ -81,6 +81,46 @@ void vmm_arch_init()
...
+        case HENVCFG_PMM_PMLEN_16:
+            pmm_val = HENVCFG_PMM_PMLEN_16 << HENVCFG_PMM_OFF;
+            break;
+        default:
+            if (cpu_is_master()) {
+                ERROR("Unsupported PMM mode for Ssnpm extension.\r\n");
+            }
+    }
+
+    csrs_henvcfg_set(pmm_val);
```
josecm, line 116: To make sure we do this cleanly dont we first need to csrs_henvcfg_clear(HENVCFG_PMM_MSK) ?

josecm (changes requested):

ninolomata:
I added a field per-VM to disable/enable this extension.

src/arch/riscv/vmm.c
ninolomata: fixed in the new version.

PR #359, opened 2026-05-08
fix(gpsr): fix the code to allow other GPSR ids other than 0

miguelafsilva5:
This PR fixes some problems where bao was expected to use only the GPSR 0.
The IPI_CPU_MSG was based on bao having always the GPSR id 0. Also, in the vm_ipi_init, the id 0 was hardcoded instead of using the already existing macro


danielRep (approved):
LGTM

DavidMCerdeira (approved):

PR #360, opened 2026-05-14
Fix undefined behavior in BIT_MASK() for LEN == 0

malejo97:
This PR fixes undefined behavior in `BIT_MASK(OFF, LEN)` when `LEN = 0`.

Previous implementation could evaluate `1UL << -1` resulting in an invalid negative shift and undefined behavior.

The fix:

- Safely handles LEN == 0
- Removes undefined behavior
- Preserves existing behavior for valid non-zero lengths


danielRep (approved):

malejo97:
We've preferred to maintain the existing `BIT_MASK()` convention where `LEN` values are expected to be within `[1, BITS_PER_LONG]`, leaving `LEN=0` unsupported. Callers using this macro are expected to respect those boundaries.

Closing this PR in favor of fixing the offending call site and documenting the valid `LEN` range.

Follow-up: #361

PR #154, opened 2024-06-05
feat(test-framework): Add a test-framework to bao

miguelafsilva5:
## PR Description

This PR integrates a test framework into bao-hypervisor. This is done by adding 2 submodules:

- **bao-tests** : This repo contains the python scripts responsible for building the system, connect to the board under test, and log the test results. It also contains the C code required to declare and implement tests.
- **bao-nix** : This repo is comprised of several nix packages used to build the whole system stack (toolchains, firmware, bao, guests, etc)

Furthermore, this PR includes a directory dedicated to the implementation of test setups (recipes).
More info on how to use is available in tests/README.md.

This PR is currently working for qemu-aarch64-virt and qemu-riscv64-virt

PR #363, opened 2026-05-15
feat(kao): add support for bao kao testing framework

danielRep:
# Overview

This PR integrates Bao Kao (abbreviated as bkao, "to put to the test"), an end-to-end testing (and benchmarking) framework for the Bao Hypervisor. Bao Kao automates the full pipeline from source build to on-target execution: it fetches toolchains, builds guest workloads, generates Bao configuration files based on targeted setup from YAML descriptors, assembles a bootable image (with optional firmware), launches it on the target platform (emulated or physical), and captures serial output to determine pass/fail.

Two new submodules are introduced:
- tests/bkao: submodule containing the Bao Kao framework 
- tests/benchs: OSYX close-source benchmarks

Platforms supported: 
- Virtual:
   - qemu-aarch64-virt 
   - qemu-riscv64-virt
   - fvp-a
   - fvp-r
- Physical 
   - zcu104
   - s32z270
   - tc4dx
   - rh850 (partially)

> [!NOTE]
> This PR supersedes #154. We have refactored completely the framework, removing nix support and leveraging only python to implement bao kao. 

# Source Tree Structure

```
  tests/
  ├── bkao/                   # Bao Kao framework (submodule)
  │   └── src/
  │       ├── bkao.py         # Main runner and orchestration entry point
  │       ├── platforms/      # Per-platform build and launch logic
  │       ├── firmware/       # ATF, OpenSBI, U-Boot constructors
  │       ├── guests/         # Guest workload builders (baremetal, ...)
  │       ├── hypervisor/     # Hypervisor builders including Bao and a config generator
  │       └── utils/          # Helpers
  ├── benchs/                 # Benchmarks workload repository (submodule)
  │   ├── src/benchmarks/
  │   │   ├── ctx-switch/     # Context-switch benchmark
  │   │   └── irq-lat/        # IRQ latency benchmark
  │   └── configs/            # Per-benchmark YAML platform configs
  └── tests/
      ├── configs/
      │   ├── baremetal/      # Baremetal guest configs (YAML + platform BSPs)
      │   ├── freertos/       # FreeRTOS guest configs
      │   └── linux/          # Linux guest configs
      └── src/
          ├── inc/            # Test framework headers (testf.h, asserts, commands)
          ├── boot.c          # Boot test
          ├── irq.c           # IRQ test 
          └── bao-test.mk     # Test build integration makefile
```

# Tests Available

```
┌─────┬────────────┬──────────┬───────────┬──────────────────────────────────────────────┐
│ ID  │   Suite    │   Test   │   Setup   │                 Description                  │
├─────┼────────────┼──────────┼───────────┼──────────────────────────────────────────────┤
│ 100 │ BOOT_CHECK │ VM_BOOT  │ baremetal │ Check that baremetal guest boots             │
│     │            │          │           │ successfully                                 │
├─────┼────────────┼──────────┼───────────┼──────────────────────────────────────────────┤
│ 101 │ BOOT_CHECK │ CPU_BOOT │ baremetal │ Check that all CPUs on the baremetal guest   │
│     │            │          │           │ boot successfully                            │
├─────┼────────────┼──────────┼───────────┼──────────────────────────────────────────────┤
│ 200 │ IRQ_CHECK  │ TIMER    │ baremetal │ Check that timer interrupt is triggered and  │
│     │            │          │           │ handled successfully                         │
├─────┼────────────┼──────────┼───────────┼──────────────────────────────────────────────┤
│ 201 │ IRQ_CHECK  │ UART     │ baremetal │ Check that UART interrupt is triggered and   │
│     │            │          │           │ handled successfully                         │
└─────┴────────────┴──────────┴───────────┴──────────────────────────────────────────────┘
```

Tests are defined using the `BAO_TEST(suite, test, setup, description)` macro and discovered automatically at runtime by scanning `tests/tests/src/*.c`. Each test is tagged with a suite, a setup (which selects the VM configuration to use), and an ID used for selective execution.

IDs are computed as `(file_index × 100) + test_index_within_file`, where:

- file_index — 1-based position of the .c file in alphabetical order within
tests/tests/src/
- test_index — 0-based order of BAO_TEST(...) appearances within that file

So with the current files (boot.c, irq.c):

```
┌────────┬────────────┬─────────┬─────┐
│  File  │ file_index │ test_nr │ ID  │
├────────┼────────────┼─────────┼─────┤
│ boot.c │ 1          │ 0       │ 100 │
├────────┼────────────┼─────────┼─────┤
│ boot.c │ 1          │ 1       │ 101 │
├────────┼────────────┼─────────┼─────┤
│ irq.c  │ 2          │ 0       │ 200 │
├────────┼────────────┼─────────┼─────┤
│ irq.c  │ 2          │ 1       │ 201 │
└────────┴────────────┴─────────┴─────┘
```

Guest types supported in test configs: baremetal, freertos, linux. For now, we only target a baremetal setup.

# How to Run (TLDR)

Via bao Makefile targets:
**Run all tests for a platform:**
`make tests PLATFORM=<platform>`

**Run all benchmarks for a platform:**
`make benchs PLATFORM=<platform>`

**Other options only via calling bkao-py directly:**

**Run specific tests or benchmarks by ID:**
`python3 tests/bkao/src/bkao.py -t 100 101 -p <platform> `
`python3 tests/bkao/src/bkao.py -b 100 -p <platform>`

**Run all tests except some:**
`python3 bkao.py -t -x 200,201 -p <platform>`

**Skip firmware and toolchain rebuild (faster iteration):**
`python3 bkao.py -t -p <platform> --no-firmware-build --no-toolchain-build`

**Use custom hypervisor sources:**
`python3 bkao.py -t -p <platform> --hyp-srcs /path/to/bao`

**Set verbosity (0=final report only, 1=failures, 2=all):**
`python3 bkao.py -t -p <platform> -l 2`

Pass platform-specific args - only virtual platforms (e.g. GIC version):
`python3 bkao.py -t -p qemu-aarch64-virt --plat-virt-args "GICV3"`

# Bao Kao Process Pipeline

For each test or benchmark, bkao executes the following pipeline:

1. **Discover:** scans `tests/tests/src/*.c` for `BAO_TEST` macros, `benchs/src/benchmarks/` for benchmark dirs, and `platforms/*.py` for platform builders.
2. **Parse args:** reads CLI arguments, resolves which tests/benchmarks to run, and builds the runtime config.
3. **Platform setup:** instantiates the requested platform class and calls setup_platform() (emulator paths, serial config, etc.).
4. **Toolchain:** builds or fetches the cross-compilation toolchain; skipped with --no-toolchain-build, expecting the prefix to be in PATH.
5. **launch_tests():** for each group of tests/benchmarks sharing the same setup:
  - Renders config.c from the YAML VM descriptor via Jinja2 templates.
  - Builds each guest workload (baremetal, FreeRTOS, Linux, or benchmark binary).
  - Builds the Bao hypervisor against the generated config and guest images.
  - Builds firmware (ATF/OpenSBI/U-Boot); skipped with --no-firmware-build.
  - Launches the platform and captures serial output, parsing `[TESTF-C]` tokens for pass/fail.
6. **Final cleanup:** removes wrkdir/ build artifacts.


tests/tests/src/irq.c
```diff
@@ -0,0 +1,68 @@
...
+
+BAO_TEST(IRQ_CHECK, TIMER, BAREMETAL, "Check that timer interrupt is triggered and handled successfully")
+{
+    if(cpu_is_master()) {
+        COMMAND_SET_TIMEOUT(TEST_TIMEOUT);
+
+        irq_set_handler(TIMER_IRQ_ID, test_interrupt_timer_callback);
+        timer_set(TIMER_INTERVAL);
+        irq_enable(TIMER_IRQ_ID);
+        irq_set_prio(TIMER_IRQ_ID, IRQ_MAX_PRIO);
```
miguelafsilva5:
```suggestion
        irq_set_prio(TIMER_IRQ_ID, TIMER_IRQ_PRIO);
```

should follow the baremetal nomenclature

tests/tests/src/irq.c
```diff
@@ -0,0 +1,68 @@
...
+
+BAO_TEST(IRQ_CHECK, UART, BAREMETAL, "Check that UART interrupt is triggered and handled successfully")
+{
+    if(cpu_is_master()) {
+        COMMAND_SET_TIMEOUT(TEST_TIMEOUT);
+
+        irq_set_handler(UART_IRQ_ID, uart_rx_handler);
+        uart_enable_rxirq();
+        irq_enable(UART_IRQ_ID);
+        irq_set_prio(UART_IRQ_ID, IRQ_MAX_PRIO);
```
miguelafsilva5:
```suggestion
        irq_set_prio(UART_IRQ_ID, UART_IRQ_PRIO);
```

should follow the baremetal nomenclature

tests/tests/configs/baremetal/tc4dx.yaml
```diff
@@ -0,0 +1,37 @@
...
+    build_options:
+    config:
+      image:
+        base_addr: 0x90100000
+      entry: 0x90100000
+      platform:
+        cpu_num: 6
+        regions:
+          - base: 0x90100000
+            size: 0x100000
```
miguelafsilva5:
```suggestion
            size: 0x80000
```

correct memory region size

tests/tests/configs/baremetal/tc4dx.yaml
```diff
@@ -0,0 +1,37 @@
...
+          - pa: 0xF003D800
+            va: 0xF003D800
+            size: 0x400
+          - pa: 0xF46C0000
+            va: 0xF46C0000
+            size: 0x200
+            interrupts:
+              - 173
+        arch:
+          gspr_num: 1
```
miguelafsilva5:
```suggestion
          gpsr_num: 1
```

typo

tests/tests/configs/baremetal/tc4dx.yaml
```diff
@@ -0,0 +1,37 @@
...
+          gspr_groups:
```
miguelafsilva5, line 36:
```suggestion
          gpsr_groups:
```

typo. 
Also, the config generator needs to generate a cast. the output should be 
```                   
 .arch = {
              .gpsr_num = 1,
              .gpsr_groups = (unsigned long int[]){1},
          }
```

tests/tests/src/irq.c
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/be72d52d07e75cdaa95c587c3d5d4a70d8402d51.

tests/tests/src/irq.c
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/be72d52d07e75cdaa95c587c3d5d4a70d8402d51.

tests/tests/configs/baremetal/tc4dx.yaml
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/74e225e281ddbe93fdbef04f5a50085e0dde01fc.

tests/tests/configs/baremetal/tc4dx.yaml
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/74e225e281ddbe93fdbef04f5a50085e0dde01fc.

tests/tests/configs/baremetal/tc4dx.yaml
Diogo21Costa: That's a good point. The config generator was missing this detail, and it was addressed in https://github.com/bao-project/bao-kao/pull/40.

miguelafsilva5: Tested and working

Makefile
```diff
@@ -437,3 +439,17 @@ $(call ci, format, $(all_c_files))
...
 
 endif
+
+-include $(tests_dir)/tests.mk
+
+.PHONY: tests
+
+tests:
+	@echo "Running bao-kao tests for $(PLATFORM)..."
+	@python3 $(kao_dir)/kao.py -t -p $(PLATFORM)
```
josecm: We're not passing `--hyp-srcs`, so kao clones bao-hypervisor at tag v2.0.0 and tests that instead of the branch. You can see it in the last CI run: `[INFO] Fetching hypervisor sources...`, which only prints when it actually clones. So the green ticks here are on the release, not on this PR.

The flag already exists and is wired up, so I think it's just:

```suggestion
	@python3 $(kao_dir)/kao.py -t -p $(PLATFORM) --hyp-srcs $(cur_dir)
```

I'd also make the no-arg case not silently clone, otherwise anyone calling kao.py directly hits the same thing.

Note this has to go together with removing the `make clean` kao does on the hypervisor sources, otherwise every `make tests` will wipe your build/ and bin/.

Makefile
```diff
@@ -437,3 +439,17 @@ $(call ci, format, $(all_c_files))
 ci: license-check format-check
 
 endif
+
+-include $(tests_dir)/tests.mk
+
+.PHONY: tests
+
+tests:
```
josecm: There's no way to pass anything else to kao from here (test ids, `-l`, `--no-firmware-build`). Can we forward these?

The old tests.mk did this already:

```make
LOG_LEVEL?=2
ECHO?=tf
SETUP?=baremetal
TEST?=boot
```

and passed them all through, so you could do `make ... TEST=irq LOG_LEVEL=1`. I'd keep that, either the same way or with a single `KAO_ARGS`.

Also if `tests/kao` isn't initialized this fails with a bare `python3: can't open file`.

Makefile
```diff
@@ -437,3 +439,17 @@ $(call ci, format, $(all_c_files))
 ci: license-check format-check
 
 endif
+
+-include $(tests_dir)/tests.mk
```
josecm: This includes a file that isn't in the tree and then we define the targets inline anyway. It's a leftover: `tests/tests.mk` existed on feat/test-framework (67b6c7fa) and was dropped in the rewrite, but the include stayed.

Either remove the include, or put the targets back in tests/tests.mk where they were. I'd lean to the second. This is bao's build system and these are orchestrating an external runner, and `-include` also degrades nicely when the submodule isn't initialized. If you do that the include has to move up near line 73, because `non_build_targets` is consumed at line 82.

As it is now there's also a trap: if someone re-adds tests/tests.mk with a `tests:` recipe, make will warn about overriding and the inline one silently wins.

Makefile
```diff
@@ -437,3 +439,17 @@ $(call ci, format, $(all_c_files))
...
+
+tests:
+	@echo "Running bao-kao tests for $(PLATFORM)..."
+	@python3 $(kao_dir)/kao.py -t -p $(PLATFORM)
+
+.PHONY: benchs
+
+benchs:
+	@echo "Running bao-kao benchmarks for $(PLATFORM)..."
+	@python3 $(kao_dir)/kao.py -b -p $(PLATFORM)
```
josecm: If the benchmarks submodule goes away from this repo (see my comment in .gitmodules), this target should probably go with it.

Right now with no benchs checked out kao finds no benchmarks and exits 0 without doing anything, so `make benchs` looks like it worked.

.github/workflows/test-bkao.yaml
```diff
@@ -0,0 +1,41 @@
...
+    branches: [ main ]
+  pull_request:
+  workflow_dispatch:
+
+
+jobs:
+
+  run-tests:
+    runs-on: ubuntu-latest
+    container:
```
josecm: Every other workflow goes through `./.github/actions/bao-container-run` now. It was added in a97c3022 because the image is too big to pull under `container:` on the free runners (bao-ci#94). This one predates it, so it just needs to move over. The branch is 35 commits behind main anyway, so it needs a rebase first.

After that `--user root` and the safe.directory step below aren't needed, the action does both.

.github/workflows/test-bkao.yaml
```diff
@@ -0,0 +1,41 @@
...
+        ]
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Mark workspace as safe for Git
+        run: git config --global --add safe.directory $GITHUB_WORKSPACE
+
+      - name: Checkout private benchmarks submodule
```
josecm: This should go together with the submodule (see my comment in .gitmodules). It doesn't fetch anything anyway: with `update = none`, `git submodule update --init tests/benchs` just prints `Skipping submodule` and exits 0. The CI log confirms it, kao says `No runnable benchmarks discovered.`

One thing to be careful about before deleting it. The `insteadOf` line is what makes the guest build work: bao-baremetal-guest has a nested `ci` submodule with an ssh url, so when kao clones the guest it needs that rewrite. I removed the step while running this locally and the clone failed with `Host key verification failed`. So either keep that line in a step of its own, or better, change the url in bao-baremetal-guest to https.

.github/workflows/test-bkao.yaml
```diff
@@ -0,0 +1,41 @@
...
+        env:
+          TOKEN: ${{ secrets.BENCHMARKS_TOKEN }}
+        run: |
+          git config --global url."https://x-access-token:${TOKEN}@github.com/".insteadOf "git@github.com:"
+          git submodule update --init tests/benchs
+
+      - name: Run bao kao tests
+        run: make tests PLATFORM=${{ matrix.platform }}
```
josecm: We're not passing `--no-toolchain-build` or `--no-firmware-build`, so every job downloads a toolchain and rebuilds U-Boot and TF-A. From the last run the kao step is 101s on aarch64 and 119s on fvp-a, against 39s on fvp-r which doesn't build firmware, so it's around 80s a job.

`--no-toolchain-build` is safe to add today. `--no-firmware-build` isn't yet, aarch64 passes `-bios flash.bin` unconditionally so the firmware has to exist first.

.github/workflows/test-bkao.yaml
```diff
@@ -0,0 +1,41 @@
...
+    runs-on: ubuntu-latest
+    container:
+      image: baoproject/bao:latest
+      options: --user root
+    strategy:
+      matrix:
+        platform: [
+          "qemu-aarch64-virt",
+          "qemu-riscv64-virt",
+          "fvp-a",
```
josecm, line 19: These two download the FVP model (~101MB) on every run. Can we cache it with actions/cache keyed on the version?

Just so it's on record, we can't put it in the CI image. I read the EULA inside the tarball and it only grants use, there's no redistribution right (§1.1, §2.5). Downloading it like we do here is fine.

.gitmodules
```diff
@@ -1,3 +1,10 @@
 [submodule "ci"]
 	path = ci
 	url = git@github.com:bao-project/bao-ci.git
+[submodule "tests/kao"]
+	path = tests/kao
+	url = git@github.com:bao-project/bao-kao.git
+[submodule "tests/benchs"]
```
josecm: Can we drop this from the public repo altogether and keep the benchmarks submodule only in the private fork that actually has them?

As it is it never works here anyway: `update = none` means the workflow step that's supposed to fetch it just skips and exits 0, so benchs is never checked out in CI. So we're carrying a private url in a public repo for nothing. The fork can add the submodule and the matching workflow step on top.

.gitignore
```diff
@@ -1,5 +1,6 @@
 bin/*
 build/*
+**/wrkdir/
```
josecm: The reason this has to be a global glob is that kao creates wrkdir in the cwd, so it lands in the repo root next to src/ and configs/.

Can we get a `--wrkdir` option and decide from the makefile where it goes, e.g. `tests/wrkdir`? Then this is a single path. If we add it, `cleanup()` should refuse to rmtree a directory it didn't create.

tests/tests/src/inc/testf_assert.h
```diff
@@ -0,0 +1,62 @@
...
+
+#define TESTF_ASSERT_OP(x, y, op) \
+    do {                          \
+        if (!(x op y)) {          \
+            (*failures)++;        \
+            LOG_FAILURE();        \
+        }                         \
+    } while (0)
+
+#define TESTF_PASS(message) printf("    Message: %s\n", message);
```
josecm: This one isn't guarded by `TESTF_LOG_LEVEL` like the rest of the pass-side output, so it prints even at level 0. I ran with level 1 and everything went quiet except this line.

tests/tests/configs/baremetal/qemu-aarch64-virt.yaml
```diff
@@ -0,0 +1,30 @@
+
+
+# Bao Hypervisor VM Configuration
+# This YAML describes the fields needed to generate a C config file (config.c)
+
+vms:
```
josecm, line 6: Question, not a change request: what do we get from the yaml + jinja over just writing config.c?

One yaml per (setup, platform) is one config.c per (setup, platform), so it's not fewer files, and nothing in the output is actually dynamic. Every value comes from the yaml, and `BAO_WRKDIR_IMGS` is resolved by the preprocessor, not the template. It's also missing things we'd want to test: cpu_affinity, colors, ipcs, remio_devs, mmu. And s32z270 already ships a hand-written plat.c next to its yaml.

Is the idea (a) to describe VMs separately and compose them for multi-guest setups, or (b) to be able to generate configs for another hypervisor? `args[2].split("+")` already builds a guest list so (a) looks intended, but the config lookup is still per-combination. If it is (a), does it also do the allocation (non-overlapping regions, cpu partitioning)? Just concatenating two VM configs doesn't really save anything.

Either way it should be in the README, because from outside it isn't obvious what it's for.

tests/tests/configs/freertos/qemu-aarch64-virt.yaml
```diff
@@ -0,0 +1,28 @@
+# Bao Hypervisor VM Configuration
+# This YAML describes the fields needed to generate a C config file (config.c)
+
+vms:
+  - vm1:
+    name: "baremetal"
```
josecm: This is a byte-for-byte copy of the baremetal one, including `name: "baremetal"`, so the freertos config declares a baremetal VM. Same for the linux one. Nothing can select them either, `build_guests` only knows baremetal.

Can we drop both directories from this PR and bring them back with the guest builders behind them?

tests/tests/configs/linux/qemu-aarch64-virt.yaml
```diff
@@ -0,0 +1,28 @@
+# Bao Hypervisor VM Configuration
+# This YAML describes the fields needed to generate a C config file (config.c)
+
+vms:
+  - vm1:
+    name: "baremetal"
```
josecm: Same as the freertos one.

tests/tests/configs/baremetal/s32z270/bsp/Clock_Ip.c
```diff
@@ -0,0 +1,1241 @@
+/*
+ * Copyright 2021-2024 NXP
```
josecm: This BSP is 79 files and 33.5k lines, which is around 63% of the PR, for a board that isn't in the CI matrix and needs a T32 probe anyway. It's also a different license from ours and tests/ is outside license-check's scope, so nothing verifies it.

Can we leave s32z270 for a follow-up, or move the BSP to its own repo as a submodule like benchs? It would make this PR much easier to review.

josecm (changes requested):

tests/tests/src/irq.c
```diff
@@ -0,0 +1,68 @@
...
+    if(cpu_is_master()) {
+        COMMAND_SET_TIMEOUT(TEST_TIMEOUT);
+
+        irq_set_handler(UART_IRQ_ID, uart_rx_handler);
+        uart_enable_rxirq();
+        irq_enable(UART_IRQ_ID);
+        irq_set_prio(UART_IRQ_ID, UART_IRQ_PRIO);
+        COMMAND_SEND_CHAR("a");
+
+        timer_wait(TEST_TIME_WAIT);
```
josecm: The TIMER test above spins on the flag with the watchdog as backstop, but this one sleeps 200ms and asserts. The injected char goes through the logger thread and a pty, so on a loaded runner 200ms can be missed and the test goes red on timing alone. Can we make it spin on `irq_en_uart` like TIMER does? The watchdog already bounds it.

josecm:
@Diogo21Costa can you go comment by comment and say which points of the review are addressed by the latest push, and if relevant, provide futher information. For those not addressed and we which you don't plan to address, please explain we don't think they should be.

Makefile
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/ff5c6e2b826e1f5771ee82484e24942c018625cc.

Aditionally, this required changes in bao-kao (https://github.com/bao-project/bao-kao/commit/322e337863f7235b81dfa483d6b542af5b5f9b96)

Makefile
Diogo21Costa: fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/f69f2ce4f4ee99e8cf41dbe31f5efd42b4c4c9ef and https://github.com/bao-project/bao-hypervisor/pull/363/commits/1a690a52a91b32fd9a67558bfdf1c7682f740efb.

Makefile
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/d47ac0fc6c2a5f8b66bcbe2a87e26489edd88d37.

Makefile
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/664d043b33a18bc71b54d026261dd74445260ade.

.github/workflows/test-bkao.yaml
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/3ec913a9edfdc8ecb13579c3eb20519ea2a01cc9.

.github/workflows/test-bkao.yaml
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/664d043b33a18bc71b54d026261dd74445260ade.

.github/workflows/test-bkao.yaml
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/902ebe470a8dba71da508efdfc1db99ed35c9afe.

.github/workflows/test-bkao.yaml
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/37690113e64e61b03a51416f73ebede57e65b43e.

.gitmodules
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/664d043b33a18bc71b54d026261dd74445260ade.

.gitignore
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/707e9e4f97c4387c6801fdc6af646343b9522842. Additionally, it required changes in bao-kao (https://github.com/bao-project/bao-kao/commit/f3dd67f6e8bf0e97bbc30daae58c335adbceceb7).

tests/tests/src/inc/testf_assert.h
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/54bfa1299cbef545ae540ad5f75b08898bb5a7df.

tests/tests/configs/baremetal/qemu-aarch64-virt.yaml
Diogo21Costa: This was a design choice; its main purpose is to make a test scenario a single, declarative unit: the Bao configuration together with the guest/build parameters needed to reproduce that scenario. In that sense, a YAML file describes more than the contents of config.c. It is intended to be the input consumed by the test/build tooling, while the generated config.c is one target-specific artifact of that description. This also gives us a path to emit equivalent scenario configurations for other hypervisors, rather than making the benchmark/test definitions inherently Bao-specific.

That said, the current implementation does not yet fully realize the compositional model you describe:
- `args[2].split("+")` was introduced with multi-guest scenarios in mind, but configuration selection is still effectively per (setup, platform) combination.
- It does not currently compose independently defined VM descriptions and perform resource allocation or validation (for example, CPU partitioning, non-overlapping memory regions, cache colors, IPCs, remote-I/O devices, or MMU settings).

So the intended answer is closer to (a) as a longer-term scenario-description direction, plus (b) as an eventual portability benefit, rather than a claim that the current YAML files already provide full VM composition or allocation. Today, they mainly centralize the scenario metadata/configuration used by the test infrastructure and generate a Bao-specific configuration from it.

tests/tests/configs/freertos/qemu-aarch64-virt.yaml
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/ab070b9f9c69e8a7f40061b1331ab502b2601906.

tests/tests/configs/linux/qemu-aarch64-virt.yaml
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/ab070b9f9c69e8a7f40061b1331ab502b2601906.

tests/tests/configs/baremetal/s32z270/bsp/Clock_Ip.c
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/f2f822f4ce17831b92937653eb47273ab945a7ff.

tests/tests/src/irq.c
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/10aeae72b85c1a1d918a7f3dd7857b4e97d2d33a.

Diogo21Costa:
> @Diogo21Costa can you go comment by comment and say which points of the review are addressed by the latest push, and if relevant, provide futher information. For those not addressed and we which you don't plan to address, please explain we don't think they should be.

@josecm Matched each of the comments with the specific fixes. While validating the review changes, make tests PLATFORM=qemu-riscv64-virt exposed additional integration problems. These were addressed by: 
cbcf574 fix(riscv): update test configuration
0f69f50 update(kao): add boot timeout
0fcda4b update(kao): use supported RISC-V QEMU
e17c3ab update(kao): report setup failures
80c2f98 update(kao): enable RISC-V Svpbmt
dea359b fix(riscv): build AIA test guest

with corresponding Bao Kao commits:
e2f3659 fix(riscv): enable AIA in QEMU
1334fec fix(logger): time out stalled boots
71d0579 fix(riscv): use supported QEMU version
f72572a fix(setup): report command failures
ad50039 fix(riscv): enable Svpbmt in QEMU

.github/workflows/test-bkao.yaml
```diff
@@ -0,0 +1,42 @@
...
+        with:
+          submodules: recursive
+
+      - name: Cache FVP model
+        if: startsWith(matrix.platform, 'fvp-')
+        uses: actions/cache@v4
+        with:
+          path: |
+            tests/wrkdir/platforms/fvp_a/FVP_Model_A
+            tests/wrkdir/platforms/fvp_r/FVP_Model_R
```
josecm: This breaks on the first cache hit. actions/cache restores only the two model directories, so `tests/wrkdir` comes back without the `.kao-wrkdir` marker and `prepare_wrkdir` refuses it. The Aug 31 run was a miss (`Cache not found for input keys`) and saved the cache at the end, so the next push to this branch is the one that hits. I reproduced it locally by creating the model directory by hand and running kao: `RuntimeError: Refusing to use unmarked working directory: .../tests/wrkdir`.

Simplest fix is to cache the marker too:
```suggestion
          path: |
            tests/wrkdir/.kao-wrkdir
            tests/wrkdir/platforms/fvp_a/FVP_Model_A
            tests/wrkdir/platforms/fvp_r/FVP_Model_R
```
Though I'd rather kao tolerated a pre-existing wrkdir, or kept the model somewhere outside it, since anyone restoring a cache by hand hits the same wall.

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "ci"]
 	path = ci
 	url = git@github.com:bao-project/bao-ci.git
+[submodule "tests/kao"]
+	path = tests/kao
+	url = git@github.com:bao-project/bao-kao.git
```
josecm: The pin (5128ee4) is on the `fix/PR_review` branch of bao-kao, 9 commits ahead of main, and there's no PR open for it. So this would merge pointing at a branch that can be force-pushed or deleted. Can we get those merged on bao-kao first and repin here?

tests/tests/configs/baremetal/qemu-aarch64-virt.yaml
josecm: I would maybe be inclined for the metadata to be on the yaml and we providing a coupled separate config.c that may be pointed to by the yaml. What do you think?

.github/workflows/test-bkao.yaml
josecm, line 34: This breaks on the first cache hit. actions/cache restores only the two model directories, so `tests/wrkdir` comes back without the `.kao-wrkdir` marker and `prepare_wrkdir` refuses it. The Aug 31 run was a miss (`Cache not found for input keys`) and saved the cache at the end, so the next push to this branch is the one that hits. I reproduced it locally by creating the model directory by hand and running kao: `RuntimeError: Refusing to use unmarked working directory: .../tests/wrkdir`.

Simplest fix is to cache the marker too:
```suggestion
          path: |
            tests/wrkdir/.kao-wrkdir
            tests/wrkdir/platforms/fvp_a/FVP_Model_A
            tests/wrkdir/platforms/fvp_r/FVP_Model_R
```
Though I'd rather kao tolerated a pre-existing wrkdir, or kept the model somewhere outside it, since anyone restoring a cache by hand hits the same wall.

.gitmodules
josecm, line 6: The pin (5128ee4) is on the `fix/PR_review` branch of bao-kao, 9 commits ahead of main, and there's no PR open for it. So this would merge pointing at a branch that can be force-pushed or deleted. Can we clean up that history (it's nine fix commits on top of each other), land it on bao-kao main, and repin here to the resulting main commit? I'd rather the gitlink point at a clean main history than at a review branch.

tests/tests/src/irq.c
```diff
@@ -0,0 +1,67 @@
+#include "testf.h"
```
josecm, line 6: No SPDX header here or in boot.c, and the yaml configs have none either. That's because tests/ is outside `all_files` and `all_c_files` in the Makefile, so neither license-check nor format-check ever sees it. I ran clang-format with the ci/.clang-format style on these two files and both come back changed (`if(` without the space, the long `BAO_TEST` lines, the aligned defines). Can we add tests/tests to both lists and fix what falls out? The example configs already carry the header, so the yaml files can take a comment version of it.

tests/tests.mk
```diff
@@ -0,0 +1,23 @@
...
+check-kao:
+	@if [ ! -f $(kao_script) ]; then \
+		echo "Bao Kao is not initialized; run 'git submodule update --init --recursive tests/kao'."; \
+		exit 1; \
+	fi
+
+.PHONY: tests
+tests: check-kao
+	@echo "Running bao-kao tests for $(PLATFORM)..."
+	@python3 $(kao_script) -t $(KAO_TESTS) -p $(PLATFORM) \
```
josecm, line 26: `make tests` with no PLATFORM gets past the Makefile's own check (tests is a non build target) and dies inside kao with an argparse usage error, `argument -p/--platform: expected one argument`. Can we check it in check-kao next to the submodule test, so both misuses get a one line message?

josecm (changes requested):

.github/workflows/test-bkao.yaml
Diogo21Costa: Thanks! Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/24a0aeaef7a595b87c4d19c4fc49efe42001e53a.

tests/tests/src/irq.c
Diogo21Costa: Thanks! Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/f461852bca5803bc8ec65afca8f5652e9d9f2183.

tests/tests.mk
Diogo21Costa: Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/5b6ad2fa9932d5a40bc0ddc829bf03fcd8fbad9c

miguelafsilva5:
Using the TC4DX as an example (not sure if this is the same for other platforms). I believe kao assumes the board is always connected via ttyUSB0. I believe this is an hardcoded value. Maybe we can solve this by passing the port via command line? i.e. PORT=ttyUSB0. while leaving USB0 as default.

tests/tests/src/inc/testf_assert.h
```diff
@@ -0,0 +1,67 @@
...
+    } while (0)
+
+#define TESTF_PASS(message)                       \
+    do {                                          \
+        if (TESTF_LOG_LEVEL > 1) {                \
+            printf("    Message: %s\n", message); \
+        }                                         \
+    } while (0)
+
+#define TESTF_FAIL(message) \
```
miguelafsilva5: Any specific reason for not having this macro inside a do {} while(0), like the others?

miguelafsilva5:
Are all tests compiled and executed for all platforms? Let's say I want to develop and test a specific feature from a certain platform, how do I implement such tests? Can they be excluded from some architectures? Let's use cache coloring as an example. I might be reasonable to test if the coloring is working, but some platforms do not use caches.

miguelafsilva5:
I was testing creating tests. And commenting a test does not compile. 

Example:
```
/*
BAO_TEST(MYTESTS, BOOT, BAREMETAL,
    "Just a simple test to check that the test framework is working")
{
    if (cpu_is_master()) 
        TESTF_PASS("Test framework is working!\n");
}
*/
```

tests/tests/src/inc/testf_assert.h
```diff
@@ -0,0 +1,67 @@
...
+#define EXPECTED_TRUE(x)         TESTF_ASSERT_OP(x, 0, !=)
+#define EXPECTED_FALSE(x)        TESTF_ASSERT_OP(x, 0, ==)
+#define EXPECTED_PTR_NULL(x)     TESTF_ASSERT_OP(x, NULL, ==)
+#define EXPECTED_PTR_NOT_NULL(x) TESTF_ASSERT_OP(x, NULL, !=)
+
+#define EXPECTED_ARRAY_EQUAL(ptr1, ptr2, count, type)           \
+    do {                                                        \
+        unsigned int intra_fails = 0;                           \
+        for (int i = 0; i < (count); i = i + 1) {               \
+            if (*(type*)((ptr1) + i) != *(type*)((ptr2) + i)) { \
```
miguelafsilva5: There is a missing cast in this line. The ideal solution is:
`if (((const type*)(ptr1))[i] != ((const type*)(ptr2))[i])` 

Add `const` so the macro can be used with consts. And if we don't cast the pointer type before incrementing, it will stride with the original type. e.g.: if you have an array of ints, and you want to compare one element as an array of chars, it will compare the first element of each int in the original array. 

This is a test made by Claude to prove this assessment:

```
BAO_TEST(MYTESTS, ARRAY_EQUAL_ELEM_STRIDE, BAREMETAL,
    "EXPECTED_ARRAY_EQUAL must stride by sizeof(type), not by sizeof(*ptr)")
{
    if (cpu_is_master()) {
        static uint32_t a[8];
        static uint32_t b[8];

        for (int i = 0; i < 8; i = i + 1) {
            a[i] = 0u;
            b[i] = 0u;
        }
        a[0] = 0x11223344u;
        b[0] = 0x1122ff44u; /* differ from a[0] only in byte offset 1 */

        const unsigned char* ba = (const unsigned char*)a;
        const unsigned char* bb = (const unsigned char*)b;

        unsigned int correct_mismatches = 0;
        for (int i = 0; i < 8; i = i + 1) {
            if (ba[i] != bb[i]) {
                correct_mismatches++;
            }
        }

        unsigned char fails_before = *failures;
        EXPECTED_ARRAY_EQUAL(a, b, 8, unsigned char);
        unsigned int macro_mismatches = (unsigned int)(*failures - fails_before);
        *failures = fails_before; /* roll back the probe's own accounting */

        printf("[PROBE] ba[1]=0x%x bb[1]=0x%x\n", (unsigned)ba[1], (unsigned)bb[1]);
        printf("[PROBE] correct byte-wise mismatches = %u\n", correct_mismatches);
        printf("[PROBE] EXPECTED_ARRAY_EQUAL mismatches = %u\n", macro_mismatches);

        if (macro_mismatches == correct_mismatches) {
            TESTF_PASS("EXPECTED_ARRAY_EQUAL agreed with a correct byte-wise comparison\n");
        } else {
            TESTF_FAIL("EXPECTED_ARRAY_EQUAL strided by sizeof(*ptr) not sizeof(type): "
                       "it read byte 0 of each 32-bit word and missed the difference "
                       "at byte offset 1\n");
        }
    }
}
```

tests/tests/src/inc/testf_assert.h
```diff
@@ -0,0 +1,67 @@
...
+                }                                               \
+            }                                                   \
+        }                                                       \
+    } while (0)
+
+#define EXPECTED_ARRAY_NOT_EQUAL(ptr1, ptr2, count, type)       \
+    do {                                                        \
+        unsigned int intra_fails = 0;                           \
+        for (int i = 0; i < (count); i = i + 1) {               \
+            if (*(type*)((ptr1) + i) != *(type*)((ptr2) + i)) { \
```
miguelafsilva5: Same thing as in the previous comment

josecm:
> Using the TC4DX as an example (not sure if this is the same for other platforms). I believe kao assumes the board is always connected via ttyUSB0. I believe this is an hardcoded value. Maybe we can solve this by passing the port via command line? i.e. PORT=ttyUSB0. while leaving USB0 as default.

I think this is a very important point. Each platform driver must be behind an abstract interface (class?) and each platform implements its access details. Is this the current architecture?

Diogo21Costa:
> > Using the TC4DX as an example (not sure if this is the same for other platforms). I believe kao assumes the board is always connected via ttyUSB0. I believe this is an hardcoded value. Maybe we can solve this by passing the port via command line? i.e. PORT=ttyUSB0. while leaving USB0 as default.
> 
> I think this is a very important point. Each platform driver must be behind an abstract interface (class?) and each platform implements its access details. Is this the current architecture?

> > Using the TC4DX as an example (not sure if this is the same for other platforms). I believe kao assumes the board is always connected via ttyUSB0. I believe this is an hardcoded value. Maybe we can solve this by passing the port via command line? i.e. PORT=ttyUSB0. while leaving USB0 as default.
> 
> I think this is a very important point. Each platform driver must be behind an abstract interface (class?) and each platform implements its access details. Is this the current architecture?

Platform-specific operations are already encapsulated in individual platform classes, and the Kao runner accesses them through a common set of methods, such as `setup_platform()`, `build_firmware()`, `get_serial_ports()`, and `launch_test()`. You are correct that each physical platform currently defines a hard-coded default serial device. To address this, I introduced an override through the `--serial-port` argument. When provided, this argument overrides the UART port defined by the selected platform; when omitted, Kao continues to use that platform’s existing default port. For example:
`python3 src/kao.py -p tc4dx --serial-port /dev/ttyUSB1 ...`
Thus, TC4DX still defaults to `/dev/ttyUSB0`, while users can select a different port without changing the platform driver.


This update introduces changes in the bao-kao repo only (https://github.com/bao-project/bao-kao/pull/41/commits/15006f7b2241839aa9e3f613e7cf624a14cb0b46).

Diogo21Costa:
> Are all tests compiled and executed for all platforms? Let's say I want to develop and test a specific feature from a certain platform, how do I implement such tests? Can they be excluded from some architectures? Let's use cache coloring as an example. I might be reasonable to test if the coloring is working, but some platforms do not use caches.

A Kao invocation targets one platform only. Kao does not automatically execute the test suite across every platform; CI must invoke Kao separately for each target. To add a test, the author defines it using `BAO_TEST(...)` in a C source file and provides the corresponding setup configuration for the target platform under `tests/tests/configs/<setup>/<platform>.yaml.`
All test C source files are currently compiled into the guest, regardless of the selection. However, only the tests selected with -t are called and executed. Tests can be manually excluded from a particular invocation using --test-exclude.

There is currently no declarative mechanism for a test author to specify supported platforms, architectures, or required capabilities. Therefore, a cache-coloring test cannot currently be marked to automatically skip platforms without cache-coloring support; each platform invocation or CI matrix entry must explicitly include or exclude it.
Supporting this automatically would require introducing platform or capability requirements in the test metadata and filtering tests before building and executing them.

Diogo21Costa:
> I was testing creating tests. And commenting a test does not compile.
> 
> Example:
> 
> ```
> /*
> BAO_TEST(MYTESTS, BOOT, BAREMETAL,
>     "Just a simple test to check that the test framework is working")
> {
>     if (cpu_is_master()) 
>         TESTF_PASS("Test framework is working!\n");
> }
> */
> ```

You are correct. Kao was scanning the raw C source, so BAO_TEST definitions inside comments were still discovered and added to the generated test entry code, causing the build to fail. Fixed in https://github.com/bao-project/bao-kao/pull/41/commits/62ee65e73b4e5a3e0f056778d846fb37579e7d52

tests/tests/src/inc/testf_assert.h
Diogo21Costa: Good catch, fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/a66c1277167e4705f637310ea58202a64f8c1b5b.

tests/tests/src/inc/testf_assert.h
Diogo21Costa: Thanks! Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/1031678a0bda292638613660091bb88ce4160fb3.

tests/tests/src/inc/testf_assert.h
Diogo21Costa: Thanks! Fixed in https://github.com/bao-project/bao-hypervisor/pull/363/commits/1031678a0bda292638613660091bb88ce4160fb3.

miguelafsilva5:
> > > Using the TC4DX as an example (not sure if this is the same for other platforms). I believe kao assumes the board is always connected via ttyUSB0. I believe this is an hardcoded value. Maybe we can solve this by passing the port via command line? i.e. PORT=ttyUSB0. while leaving USB0 as default.
> > 
> > 
> > I think this is a very important point. Each platform driver must be behind an abstract interface (class?) and each platform implements its access details. Is this the current architecture?
> 
> > > Using the TC4DX as an example (not sure if this is the same for other platforms). I believe kao assumes the board is always connected via ttyUSB0. I believe this is an hardcoded value. Maybe we can solve this by passing the port via command line? i.e. PORT=ttyUSB0. while leaving USB0 as default.
> > 
> > 
> > I think this is a very important point. Each platform driver must be behind an abstract interface (class?) and each platform implements its access details. Is this the current architecture?
> 
> Platform-specific operations are already encapsulated in individual platform classes, and the Kao runner accesses them through a common set of methods, such as `setup_platform()`, `build_firmware()`, `get_serial_ports()`, and `launch_test()`. You are correct that each physical platform currently defines a hard-coded default serial device. To address this, I introduced an override through the `--serial-port` argument. When provided, this argument overrides the UART port defined by the selected platform; when omitted, Kao continues to use that platform’s existing default port. For example: `python3 src/kao.py -p tc4dx --serial-port /dev/ttyUSB1 ...` Thus, TC4DX still defaults to `/dev/ttyUSB0`, while users can select a different port without changing the platform driver.
> 
> This update introduces changes in the bao-kao repo only ([bao-project/bao-kao@15006f7](https://github.com/bao-project/bao-kao/commit/15006f7b2241839aa9e3f613e7cf624a14cb0b46)).

@Diogo21Costa is this solution only available via python? Can I not overwrite the port using make?

If the make functionality is meant to be mainly used by the CI, I believe such an override still needs to be available. Imagine you have several boards connected to server. And for some reason, a board needs to be taken out to a live demo or something. The CI should be able to update the actions script easily to adapt to such conditions.

miguelafsilva5:
> > Are all tests compiled and executed for all platforms? Let's say I want to develop and test a specific feature from a certain platform, how do I implement such tests? Can they be excluded from some architectures? Let's use cache coloring as an example. I might be reasonable to test if the coloring is working, but some platforms do not use caches.
> 
> A Kao invocation targets one platform only. Kao does not automatically execute the test suite across every platform; CI must invoke Kao separately for each target. To add a test, the author defines it using `BAO_TEST(...)` in a C source file and provides the corresponding setup configuration for the target platform under `tests/tests/configs/<setup>/<platform>.yaml.` All test C source files are currently compiled into the guest, regardless of the selection. However, only the tests selected with -t are called and executed. Tests can be manually excluded from a particular invocation using --test-exclude.
> 
> There is currently no declarative mechanism for a test author to specify supported platforms, architectures, or required capabilities. Therefore, a cache-coloring test cannot currently be marked to automatically skip platforms without cache-coloring support; each platform invocation or CI matrix entry must explicitly include or exclude it. Supporting this automatically would require introducing platform or capability requirements in the test metadata and filtering tests before building and executing them.

@Diogo21Costa my only gripe with this approach is that you are limiting a test developer from using platform-specific APIs. If I develop a test for TC4 using CSFR reads/writes, they won't compile for ARM or RISC-V. And then, you are putting a strain on the ARM and RISC-V developers to one-by-one exclude tests that don't compile from their builds.  
Am I understanding this wrong?

miguelafsilva5:
> > I was testing creating tests. And commenting a test does not compile.
> > Example:
> > ```
> > /*
> > BAO_TEST(MYTESTS, BOOT, BAREMETAL,
> >     "Just a simple test to check that the test framework is working")
> > {
> >     if (cpu_is_master()) 
> >         TESTF_PASS("Test framework is working!\n");
> > }
> > */
> > ```
> 
> You are correct. Kao was scanning the raw C source, so BAO_TEST definitions inside comments were still discovered and added to the generated test entry code, causing the build to fail. Fixed in [bao-project/bao-kao@62ee65e](https://github.com/bao-project/bao-kao/commit/62ee65e73b4e5a3e0f056778d846fb37579e7d52)

Tested and working :ok_hand:

josecm:
> @Diogo21Costa my only gripe with this approach is that you are limiting a test developer from using platform-specific APIs. If I develop a test for TC4 using CSFR reads/writes, they won't compile for ARM or RISC-V. And then, you are putting a strain on the ARM and RISC-V developers to one-by-one exclude tests that don't compile from their builds. Am I understanding this wrong?

I think this makes a lot of sense. Each test should be couple with some description of it targets the hypervisor globally, a given architecture, or a given platform. And the framework could understand "oh! I only run this test on this platforms" or "I can't run this on this platform".

But maybe we can get the PR through without such feature and add it next.

josecm:
> @Diogo21Costa is this solution only available via python? Can I not overwrite the port using make?
> 
> If the make functionality is meant to be mainly used by the CI, I believe such an override still needs to be available. Imagine you have several boards connected to server. And for some reason, a board needs to be taken out to a live demo or something. The CI should be able to update the actions script easily to adapt to such conditions.

@miguelafsilva5 can you explain exactly what do you mean by "update the actions script"?  You mean when the tty board for a given board changes name? 

I kind of agree that the ci make rule should allow one to do this.  I want to test and i need to override the tty port I write something like `make test KAO_SERIAL=/dev/ttyUSB2`, otherwise we need to call kao by hand no?

miguelafsilva5:
> > @Diogo21Costa is this solution only available via python? Can I not overwrite the port using make?
> > If the make functionality is meant to be mainly used by the CI, I believe such an override still needs to be available. Imagine you have several boards connected to server. And for some reason, a board needs to be taken out to a live demo or something. The CI should be able to update the actions script easily to adapt to such conditions.
> 
> @miguelafsilva5 can you explain exactly what do you mean by "update the actions script"? You mean when the tty board for a given board changes name?
> 
> I kind of agree that the ci make rule should allow one to do this. I want to test and i need to override the tty port I write something like `make test KAO_SERIAL=/dev/ttyUSB2`, otherwise we need to call kao by hand no?

Yap. That's what I meant and that's what I propose.

miguelafsilva5:
> > @Diogo21Costa my only gripe with this approach is that you are limiting a test developer from using platform-specific APIs. If I develop a test for TC4 using CSFR reads/writes, they won't compile for ARM or RISC-V. And then, you are putting a strain on the ARM and RISC-V developers to one-by-one exclude tests that don't compile from their builds. Am I understanding this wrong?
> 
> I think this makes a lot of sense. Each test should be couple with some description of it targets the hypervisor globally, a given architecture, or a given platform. And the framework could understand "oh! I only run this test on this platforms" or "I can't run this on this platform".
> 
> But maybe we can get the PR through without such feature and add it next.

We can opt for a simple solution in the meanwhile, like allowing folders inside tests/tests/src with platform names, and if the makefile finds a folder matching the PLATFORM=xxx name, it includes all the sources of that folder too.

But I'm also ok with moving forward with the PR and opening an issue

miguelafsilva5:
@josecm @Diogo21Costa I think this might a bit out-of-scope, but may be worth addressing in some form of Test Guidelines.
Let's say we are testing multiple IRQ latencies. Our current test implementation do not disarm the interrupts they are testing, which leaves them interrupting other tests. 
I think we have mention somewhere that each test is responsible to free the resources they use, once the test is finished.

josecm:
> @josecm @Diogo21Costa I think this might a bit out-of-scope, but may be worth addressing in some form of Test Guidelines. Let's say we are testing multiple IRQ latencies. Our current test implementation do not disarm the interrupts they are testing, which leaves them interrupting other tests. I think we have mention somewhere that each test is responsible to free the resources they use, once the test is finished.

This is a very good point :thinking: should we add some `tests_clear_state()` called by the harness before every tets, that brings the vm to a predefined state (ie, disables all interrupts, and we can think of other state it might require resetting...)

miguelafsilva5:
> > @josecm @Diogo21Costa I think this might a bit out-of-scope, but may be worth addressing in some form of Test Guidelines. Let's say we are testing multiple IRQ latencies. Our current test implementation do not disarm the interrupts they are testing, which leaves them interrupting other tests. I think we have mention somewhere that each test is responsible to free the resources they use, once the test is finished.
> 
> This is a very good point 🤔 should we add some `tests_clear_state()` called by the harness before every tets, that brings the vm to a predefined state (ie, disables all interrupts, and we can think of other state it might require resetting...)

While that might be a good solution for the generic tests, how do you propose to implement a clear state for platform or architectural tests?
I mean, let's say we test a new function or peripheral in a certain platform. Do we have to add a new function or edit the clear state function to clear that specific peripheral? DMAs come to mind, multiple peripherals/tests might want to use the same channels. Same for IRQs. Are we clearing all the IRQ nodes? 

Probably a better solution is to have the test developer be responsible for this and we as reviewers also be responsible for checking that before accepting any test.

PR #365, opened 2026-05-21
fix(riscv): use ~0ULL when programming stimecmp to "infinity"

josecm:
~0U is a 32-bit unsigned literal; when widened to the 64-bit parameter of csrs_stimecmp_write() it is zero-extended, leaving stimecmp with the upper half cleared instead of all ones. Affects both RV32 and RV64.


danielRep (approved):

malejo97 (approved):
This is correct.

PR #367, opened 2026-06-07
fix(riscv/sbi): do not truncate the HSM hart_start opaque (priv)

josecm:
Fix truncation of the opaque (priv) argument passed to the vcpu on SBI HSM hart_start.


sandro2pinto (approved):
All good to me.

ninolomata (approved):

PR #368, opened 2026-06-07
feat(riscv/vaplic): allow level-sensitive interrupt sources

josecm:
Allow guests to configure level-sensitive interrupt source modes in the vAPLIC.


sandro2pinto (dismissed):
All good to me.

sandro2pinto (approved):
All good to me.

ninolomata (approved):

PR #372, opened 2026-06-19
misc rh850 fixes

DavidMCerdeira:
ALIGN rounds the memory address up causing the emulation handler to try to configure the wrong interrupt.

The fault address of a bitwise operation is the final address, already reflecting base + imm from the instruction opcode.

### DIFF
--- a/src/arch/rh850/aborts.c
+++ b/src/arch/rh850/aborts.c
@@ -126,18 +126,6 @@ static void decode_access(struct emul_access* acc, unsigned long addr)
     if (opcode == F8_OPCODE) {
         bit = (inst & BITIDX_MASK) >> BITIDX_SHIFT;
         bwop = ((inst & SUB8_MASK) >> SUB8_SHIFT) + 1;
-        int16_t imm = (int16_t)(inst >> 16);
-        if (imm < 0) {
-            if ((unsigned long)(-imm) > addr) {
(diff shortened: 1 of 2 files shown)


malejo97 (approved):
Seems good to me.

josecm (commented):
1. The `ALIGN_FLOOR` macro should be able to replace the `ALIGN` macros with the intended effect.
2. Please clarify: the second edit removes a redudant calculation of the faulting address as this faulting address is autmatically given by the cpu in the `mea` register. Correct?

DavidMCerdeira:
> The ALIGN_FLOOR macro should be able to replace the ALIGN macros with the intended effect.

Any alignment operation before the division may result in the wrong reg_idx calculation.

edit:
After giving a bit more thought, although ALIGN_FLOOR would work, it is unnecessary in this case, since the end result is the same.

> Please clarify: the second edit removes a redudant calculation of the faulting address as this faulting address is autmatically given by the cpu in the mea register. Correct?

Not redundant, actually incorrect. The instruction performs an access based on base_addr + imm, in case of a fault the fault address is base_addr + imm. If we add the imm again (read from instruction op code) we are calculating the wrong fault address resulting in a final value of base_addr + imm + imm.

josecm (approved):

PR #378, opened 2026-07-15
fix(mem): invalidate TLB after mem_map to make new PTEs visible

kdamaski:
The RISC-V privileged spec permits implementations to cache invalid PTEs: making a stored valid PTE visible to the page-table walker then requires an sfence.vma. Only the Svvptc extension waives this — it was ratified precisely to guarantee invalid->valid updates become visible without a fence. mem_map() only issued a data fence (fence_sync()) after writing PTEs, which orders the stores but does not invalidate cached translations; the pre-existing "// TODO tlb shootdown?" comment suggests this was a known gap.

On cores that exercise this permission, the first access through a fresh mapping faults. Observed on the SiFive P550 (ESWIN EIC7700X, Milk-V Megrez): the memset of a just-mapped page table in vm_mem_prot_init() store-page-faults during VM setup, before any guest runs. QEMU/TCG makes PTE stores visible immediately, which is why CI does not reproduce it.

tlb_inv_all(as) is used instead of per-page tlb_inv_va(): mem_map() writes entries at mixed page/superpage granularity, and its common callers map whole VM images and memory regions (thousands of pages), where a single address-space-scoped invalidation is cheaper than one fence per page. mem_map() is not on any hot path.

### DIFF
--- a/src/arch/riscv/mem.c
+++ b/src/arch/riscv/mem.c
@@ -7,6 +7,7 @@
 
 #include <platform.h>
 #include <cpu.h>
+#include <tlb.h>
 
 static inline void as_map_physical_identity(struct addr_space* as)
 {
@@ -68,4 +69,26 @@ bool mem_translate(struct addr_space* as, vaddr_t va, paddr_t* pa)
     } else {
         return false;
     }
+}
+
+/**
+ * RISC-V permits implementations to cache invalid PTEs unless Svvptc is
+ * implemented (the SiFive P550 does; QEMU does not, which masks the bug), so
+ * after an invalid->valid transition a hart may keep faulting on the new
+ * mapping until the cached negative entry is evicted. Invalidate so the
+ * walker re-reads the new PTEs. tlb_inv_all already scopes the fence to the
+ * address space: hfence.gvma restricted to the VM's VMID for guest address
+ * spaces, sfence.vma for the hypervisor's.
+ *
+ * TODO: for small num_pages a ranged invalidation over [va, va + num_pages)
+ * would preserve the rest of the VMID's cached translations, but ranged
+ * remote fences run as per-page loops in the SBI implementation, so bulk
+ * mappings (VM RAM regions) must keep the full flush; picking the crossover
+ * threshold needs measurement. The va/num_pages parameters exist for this.
+ */
+void mem_arch_map_sync_tlbs(struct addr_space* as, vaddr_t va, size_t num_pages)
+{
+    UNUSED_ARG(va);
+    UNUSED_ARG(num_pages);
+    tlb_inv_all(as);
 }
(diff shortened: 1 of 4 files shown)


josecm:
@kdamaski thanks for this PR! Very nice people starting to run Bao on real RISC-V hardware! (Finally!)

Would it be possible to the tlb invalidation to be finer grained and refer only to the TLBs that actually matter. For example, only for G-stage TLBs in case we are mapping something on a guest address space? Also thinking if we could it even refer only to the mapped VAs. A last point would be to maybe make this arch specific, that is, calling some mem_arch_map_sync_tlbs(), which could have a default/weak empty implementation and only implemented for risc-v.

Let me know what you think about these ideas.

PS: Also please rebase instead of merging when syncing with main.

kdamaski:
> @kdamaski thanks for this PR! Very nice people starting to run Bao on real RISC-V hardware! (Finally!)
> 
> Would it be possible to the tlb invalidation to be finer grained and refer only to the TLBs that actually matter. For example, only for G-stage TLBs in case we are mapping something on a guest address space? Also thinking if we could it even refer only to the mapped VAs. A last point would be to maybe make this arch specific, that is, calling some mem_arch_map_sync_tlbs(), which could have a default/weak empty implementation and only implemented for risc-v.
> 
> Let me know what you think about these ideas.

Thanks for your review @josecm!

Agree on all three. One clarification on (1): the invalidation is already stage and VMID-scoped — tlb_inv_all(as) dispatches on as->type (core/inc/tlb.h), so for a guest AS it issues hfence.gvma limited to that VM's VMID; HS-stage and other VMs are untouched. That's why the fix passes as instead of a bare global flush.

For the other suggestions: i can try and rework it as a weak mem_arch_map_sync_tlbs(as, va, num_pages) — empty default, implemented only for RISC-V, following the existing weak-hook pattern (such as config_mem_prot_init, mem_map_reclr, …) and since it's not arch independent issue.

The signature can carry va, num_pages so the suggested mapped VAs only can slot in, but i'd do that as a follow-up since it's not the bug fix but rather an optimization.

kdamaski:
Okay i did some changes. Ready for re-review

josecm:
This seems exactly what I had in mind! Thanks @kdamaski

josecm (dismissed):

ninolomata (dismissed):

ninolomata:
Hi @kdamaski, thank you for the PR. I have already encountered what appears to be the same issue on another RISC-V platform, and I implemented a similar fix, it would be nice to optimize this in the future.

DavidMCerdeira:
Hi @kdamaski,
Before we merge this, please re-write the commit history into a single commit and rebase the branch with main.

Thanks!

kdamaski:
Done! It looks clean now.

DavidMCerdeira (approved):

danielRep (approved):

josecm (approved):

PR #379, opened 2026-07-15
fix(riscv): only access henvcfg/senvcfg when a configured extension n…

kdamaski:
…eeds them

vmm_arch_init() unconditionally writes henvcfg (clean-slate write plus the STCE clear on the no-Sstc path) and vcpu init unconditionally writes senvcfg. Both are priv-1.12 CSRs: on cores that implement the H extension but predate priv 1.12 they are absent, and the access traps with illegal instruction. Since illegal instruction is not delegated, the trap bounces through M-mode back into the hypervisor's own exception path before any vcpu exists, wedging every hart right after the Bao banner.

Observed on the SiFive P550 (ESWIN EIC7700X / Milk-V Megrez, OpenSBI reports "Priv Version: v1.11", H present): boot wedges with either CSR access present and succeeds with both guarded. QEMU implements both CSRs and does not reproduce this.

Gate the accesses on the platform declaring at least one extension whose state lives in the respective CSR (Sstc, Svpbmt, Zicboz, Zicbom), following the existing CPU_HAS_EXTENSION() pattern. No behavior change on platforms that declare any of them.

### DIFF
--- a/src/arch/riscv/arch.mk
+++ b/src/arch/riscv/arch.mk
@@ -33,6 +33,10 @@ endif
 irqc_arch_dir=$(cpu_arch_dir)/irqc/$(IRQC_DIR)
 src_dirs+=$(irqc_arch_dir)
 
+# Privileged architecture version. Bao assumes >= 1.12; pre-1.12 platforms override.
+RISCV_PRIV_VERSION ?= RISCV_PRIV_VERSION_1_12
+arch-cppflags += -DRISCV_PRIV_VERSION=$(RISCV_PRIV_VERSION)
+
 ifeq ($(ARCH_SUB), riscv64)
 arch-cppflags+=-DRV_XLEN=64
 else ifeq ($(ARCH_SUB), riscv32)
--- a/src/arch/riscv/inc/arch/csrs.h
+++ b/src/arch/riscv/inc/arch/csrs.h
@@ -6,6 +6,11 @@
 #ifndef __ARCH_CSR_H__
 #define __ARCH_CSR_H__
 
+/* Privileged-arch version (set per-platform in platform.mk); selects CSR availability. */
+#define RISCV_PRIV_VERSION_1_11 (11)
+#define RISCV_PRIV_VERSION_1_12 (12)
+#define RISCV_PRIV_VERSION_1_13 (13)
+
 #include <bao.h>
 
 #define CSR_SISELECT      0x150
--- a/src/arch/riscv/vm.c
+++ b/src/arch/riscv/vm.c
@@ -62,7 +62,10 @@ void vcpu_arch_reset(struct vcpu* vcpu, vaddr_t entry)
         csrs_sstateen0_write(0);
     }
 
+    /* senvcfg is priv-1.12 like henvcfg (see vmm.c); gate on RISCV_PRIV_VERSION. */
+#if RISCV_PRIV_VERSION >= RISCV_PRIV_VERSION_1_12
     csrs_senvcfg_write(0);
+#endif
     csrs_hcounteren_write(HCOUNTEREN_TM);
     csrs_htimedelta_write(0);
     csrs_vsstatus_write(SSTATUS_SD | SSTATUS_FS_DIRTY | SSTATUS_XS_DIRTY);
--- a/src/arch/riscv/vmm.c
+++ b/src/arch/riscv/vmm.c
@@ -22,11 +22,14 @@ void vmm_arch_init()
     csrs_hideleg_write(HIDELEG_VSSI | HIDELEG_VSTI | HIDELEG_VSEI);
     csrs_hedeleg_write(HEDELEG_BKP | HEDELEG_ECU | HEDELEG_IPF | HEDELEG_LPF | HEDELEG_SPF);
 
-    /**
-     * Start from a clean slate for the entire HENVCFG CSR
-     * to avoid unintended side effects from any non-zero default bits
+    /*
+     * Clean-slate the whole HENVCFG CSR (avoid stale default bits). It is a
+     * priv-1.12 CSR: on pre-1.12 cores (e.g. SiFive P550) it is absent and the
+     * access traps, so gate on the platform's RISCV_PRIV_VERSION.
      */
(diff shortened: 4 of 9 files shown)


josecm:
@kdamaski I am a little on the fence about going through with this because we do are assuming at least 1.12, or maybe even 1.13. The issue is that this is implicit throughout the code. 

The version for the SoC on your target board has what I would consider an outdated version we isn't our goal to support, especially because I believe it also features and H extension version prior to 1.0.

@ninolomata any thoughts on this?

kdamaski:
Understood — I won't push on this. The core is strictly pre-ratification H (priv 1.11),
so I get the hesitation about carrying guards for it in the common path.
                                                                                        
Let me ask the underlying question instead: would a milkv-megrez platform port be 
welcome in the project? The EIC7700X / SiFive P550 boards (Milk-V Megrez, HiFive 
Premier P550) are among the few commercially available H-extension silicon right now, 
and I have Bao running a full 2-VM setup on it (device passthrough + remote I/O, both 
Linux guests). In practice these two CSR accesses are the only place the pre-1.12 
baseline surfaces — everything else already falls out of the existing 
CPU_HAS_EXTENSION() gating.

If a port is of interest, I'd fold this gating into the platform PR (e.g. behind a 
single explicit macro in the platform's cpu_ext.h, so conforming platforms keep the 
unconditional clean-slate writes and the ≥1.12 assumption stays the visible default). 
If the project would rather set priv ≥1.12 / ratified H as a hard floor, that's a fair 
call too — I'll carry this downstream, and it might be worth stating that baseline in 
the docs so it's explicit.
                                                                                        
Happy to close this PR either way pending your take.

josecm:
> Let me ask the underlying question instead: would a milkv-megrez platform port be welcome in the project? The EIC7700X / SiFive P550 boards (Milk-V Megrez, HiFive Premier P550) are among the few commercially available H-extension silicon right now, and I have Bao running a full 2-VM setup on it (device passthrough + remote I/O, both Linux guests). In practice these two CSR accesses are the only place the pre-1.12 baseline surfaces — everything else already falls out of the existing CPU_HAS_EXTENSION() gating.
> 
> If a port is of interest, I'd fold this gating into the platform PR (e.g. behind a single explicit macro in the platform's cpu_ext.h, so conforming platforms keep the unconditional clean-slate writes and the ≥1.12 assumption stays the visible default). If the project would rather set priv ≥1.12 / ratified H as a hard floor, that's a fair call too — I'll carry this downstream, and it might be worth stating that baseline in the docs so it's explicit.
> 
> Happy to close this PR either way pending your take.

We did consider adding support for these boards, since we have a HiFive Premier P550, but ultimately put the effort aside because of this exact issue.

That said, since you already have the platform working and the required compatibility changes appear to be limited to a couple of well-contained guards, I think it would make sense to fold the platform support into this PR. I’d be very grateful for the contribution!

ninolomata:
Hi @kdamaski, thank you for the PR. I also think it would be reasonable to introduce a minimal guard if that allows us to support these boards.

A clean solution would be to introduce an explicit privileged architecture version instead of inferring the availability of henvcfg and senvcfg from the extensions that currently use them. Bao already follows a similar pattern for the GIC: each platform defines GIC_VERSION in its platform.mk, and the architecture code uses that value to select version-specific code. We could follow the same approach for RISC-V, for example:

RISCV_PRIV_VERSION := RISCV_PRIV_VERSION_1_11

and then:

#if RISCV_PRIV_VERSION >= RISCV_PRIV_VERSION_1_12
    csrs_henvcfg_write(0);
#endif

With an explicit version, RISCV_PRIV_VERSION would determine whether a CSR exists, while CPU_HAS_EXTENSION() would continue to determine which optional features and individual CSR fields should be enabled.

The default could be RISCV_PRIV_VERSION_1_12, which is the minimum privileged architecture version currently assumed by Bao, given the unconditional accesses to henvcfg and senvcfg. This would preserve the existing behaviour for all current platforms, while allowing pre-1.12 platforms such as the P550 to override the version explicitly.

This would also avoid having to extend the current SSTC || SVPBMT || ZICBOZ || ZICBOM condition whenever another extension starts using an envcfg field.

kdamaski:
Thank you both for your replies and guidance! I've reworked the PR along the lines @ninolomata suggested and folded in the platform per @josecm. It's now two commits:

1. feat(riscv): add explicit RISCV_PRIV_VERSION …  — replaces the extension-inference guard with an explicit privileged-arch version, mirroring the per-platform GIC_VERSION pattern. Platforms set it in platform.mk; it defaults to RISCV_PRIV_VERSION_1_12 (Bao's current baseline), so every existing platform is unchanged. The unconditional henvcfg/senvcfg writes are guarded on >= 1.12 .
2. feat(plat): add Milk-V Megrez (ESWIN EIC7700X) support ... — 4× P550, PLIC, SBI IPIs/timer, DW-APB console; sets RISCV_PRIV_VERSION_1_11. Deliberately minimal: identity-mapped passthrough (no IOMMU) since Megrez board embeds ARM-v3.1 SMMU and and the VM's RISC-V stage-2 table can't be reused for an ARM-VMSA walker -- enabling it is out of scope here. The managed DRAM regions respect the board's reserved-memory no-map carveouts.

src/arch/riscv/arch.mk
```diff
@@ -33,6 +33,10 @@ endif
 irqc_arch_dir=$(cpu_arch_dir)/irqc/$(IRQC_DIR)
 src_dirs+=$(irqc_arch_dir)
 
+# Privileged architecture version. Bao assumes >= 1.12; pre-1.12 platforms override.
+RISCV_PRIV_VERSION ?= RISCV_PRIV_VERSION_1_12
```
josecm, line 37: @kdamaski  I'd say we only change this default to 13. As this if what we'll expect for all future platforms featuring RVA23 or above. Do you agree? T

Everything else else seems fine.

josecm: Just noticed @ninolomata was the one that suggested 1.12 as the baseline. I'd still be inclined for 1.13 for the reasons stated above. @ninolomata is this ok with you? Also would like to hear your opinon on this @kdamaski

ninolomata: I am ok with it :).

PR #382, opened 2026-07-17
feat(armv8/iommu): add SMMUv3 stage-2 driver

rabara:
Add a minimal ARM SMMUv3 (IHI 0070) driver for Armv8-A, providing stage-2-only DMA translation. The driver takes over the non-secure SMMU left enabled by firmware: it installs a linear stream table (all streams BYPASS by default so masters keep working with their incoming attributes), a command queue and an event queue, and promotes individual streams to stage-2 as VMs bind them. Faults are drained and logged via the event queue (wired SPI plus a cheap poll hook on the GIC IRQ path).

Driver selection is done at build time via SMMU_VERSION (default 2): arch/smmu.h resolves to smmuv2.h or smmuv3.h, and objects.mk builds the matching driver. Platforms opt into v3 with SMMU_VERSION:=3 in their platform.mk.

Verified working on Agilex5 SoCFPGA platform.

### DIFF
--- a/src/arch/armv8/arch.mk
+++ b/src/arch/armv8/arch.mk
@@ -17,6 +17,11 @@ arch_profile_sub_dir:=$(arch_profile_dir)/$(ARCH_SUB)
 src_dirs+=$(arch_profile_sub_dir)
 
 arch-cppflags+=-DGIC_VERSION=$(GIC_VERSION)
+
+# Default to SMMUv2 so platforms that don't care about the IOMMU (or predate
+# the v2/v3 split) keep compiling. Platforms select v3 in their platform.mk.
+SMMU_VERSION?=2
+arch-cppflags+=-DSMMU_VERSION=$(SMMU_VERSION)
 ifeq ($(CC_IS_GCC),y)
 arch-cflags+=-mgeneral-regs-only
 endif
--- a/src/arch/armv8/armv8-a/inc/arch/iommu.h
+++ b/src/arch/armv8/armv8-a/inc/arch/iommu.h
@@ -7,11 +7,23 @@
 #define __IOMMU_ARCH_H__
 
 #include <bao.h>
-#include <arch/smmuv2.h>
+#include <arch/smmu.h>
 
+#if (SMMU_VERSION == 3)
+/*
+ * SMMUv3 is stage-2-only here: each bound stream gets a Stream Table Entry that
+ * points directly at the VM's stage-2 page table (vm->as.pt.root) tagged with
+ * the VM's id as S2VMID. No per-VM context banks, so the only per-VM state we
+ * track is whether the STE(s) have been installed.
+ */
+struct iommu_vm_arch {
+    bool initialized;
+};
+#else
 struct iommu_vm_arch {
     streamid_t global_mask;
     ssize_t ctx_id;
 };
+#endif
 
 #endif
--- a/src/arch/armv8/armv8-a/inc/arch/smmuv3.h
+++ b/src/arch/armv8/armv8-a/inc/arch/smmuv3.h
@@ -1,0 +1,210 @@
+/**
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (c) Bao Project and Contributors. All rights reserved.
+ *
+ * Minimal ARM SMMUv3 (IHI 0070) driver interface for Bao.
+ *
+ * Scope: stage-2-only translation. Bao owns stage 2 (the VM's IPA->PA page
+ * table); stage 1 is disabled (VMs see no SMMU). The driver installs a linear
+ * stream table, a command queue and an event queue, leaves every stream in
+ * BYPASS by default (so masters the firmware set up keep working unchanged),
+ * and promotes individual streams to stage-2 translation when a VM binds them.
+ *
+ * Register layout and field encodings follow the ARM SMMUv3 architecture
+ * specification (IHI 0070).
+ */
+
+#ifndef __ARCH_SMMUV3_H__
+#define __ARCH_SMMUV3_H__
+
+#include <bao.h>
+#include <bit.h>
+
+typedef deviceid_t streamid_t;
+
+/* ----------------------------------------------------------------------- */
+/* Non-secure register page (offsets from platform.arch.smmu.base)          */
+/* ----------------------------------------------------------------------- */
+
+struct smmuv3_hw {
+    uint32_t IDR0;            /* +0x0000 */
+    uint32_t IDR1;            /* +0x0004 */
+    uint32_t IDR2;            /* +0x0008 */
+    uint32_t IDR3;            /* +0x000C */
+    uint32_t IDR4;            /* +0x0010 */
+    uint32_t IDR5;            /* +0x0014 */
+    uint32_t IIDR;            /* +0x0018 */
+    uint32_t AIDR;            /* +0x001C */
+    uint32_t CR0;             /* +0x0020 */
+    uint32_t CR0ACK;          /* +0x0024 */
+    uint32_t CR1;             /* +0x0028 */
+    uint32_t CR2;             /* +0x002C */
+    uint8_t pad0[0x40 - 0x30];
+    uint32_t STATUSR;         /* +0x0040 */
+    uint32_t GBPA;            /* +0x0044 */
+    uint32_t AGBPA;           /* +0x0048 */
+    uint8_t pad1[0x50 - 0x4C];
+    uint32_t IRQ_CTRL;        /* +0x0050 */
+    uint32_t IRQ_CTRLACK;     /* +0x0054 */
+    uint8_t pad2[0x60 - 0x58];
+    uint32_t GERROR;          /* +0x0060 */
+    uint32_t GERRORN;         /* +0x0064 */
+    uint64_t GERROR_IRQ_CFG0; /* +0x0068 */
+    uint32_t GERROR_IRQ_CFG1; /* +0x0070 */
+    uint32_t GERROR_IRQ_CFG2; /* +0x0074 */
+    uint8_t pad3[0x80 - 0x78];
+    uint64_t STRTAB_BASE;     /* +0x0080 */
+    uint32_t STRTAB_BASE_CFG; /* +0x0088 */
+    uint8_t pad4[0x90 - 0x8C];
+    uint64_t CMDQ_BASE;       /* +0x0090 */
+    uint32_t CMDQ_PROD;       /* +0x0098 */
+    uint32_t CMDQ_CONS;       /* +0x009C */
+    uint64_t EVENTQ_BASE;     /* +0x00A0 */
+    uint32_t EVENTQ_PROD;     /* +0x00A8 (alias of page-1 PROD) */
+    uint32_t EVENTQ_CONS;     /* +0x00AC (alias of page-1 CONS) */
+};
+
+/* Every register is at its natural alignment, so the struct is NOT packed
+ * (packing would force -Waddress-of-packed-member on the &reg poll helpers).
+ * Guard the hand-computed offsets so a layout slip is caught at compile time. */
+_Static_assert(__builtin_offsetof(struct smmuv3_hw, CR0) == 0x20, "smmuv3 CR0 offset");
+_Static_assert(__builtin_offsetof(struct smmuv3_hw, GBPA) == 0x44, "smmuv3 GBPA offset");
+_Static_assert(__builtin_offsetof(struct smmuv3_hw, STRTAB_BASE) == 0x80,
+    "smmuv3 STRTAB_BASE offset");
+_Static_assert(__builtin_offsetof(struct smmuv3_hw, CMDQ_BASE) == 0x90, "smmuv3 CMDQ_BASE offset");
+_Static_assert(__builtin_offsetof(struct smmuv3_hw, EVENTQ_PROD) == 0xA8,
+    "smmuv3 EVENTQ_PROD offset");
+
+/* SMMU_IDR0 */
+#define SMMUV3_IDR0_ST_LEVEL_OFF (27)
+#define SMMUV3_IDR0_ST_LEVEL_LEN (2)
+#define SMMUV3_IDR0_COHACC_BIT   (1U << 4)
+#define SMMUV3_IDR0_VMID16_BIT   (1U << 18)
+#define SMMUV3_IDR0_TTF_OFF      (2)
+#define SMMUV3_IDR0_TTF_LEN      (2)
+#define SMMUV3_IDR0_S2P_BIT      (1U << 0)
+
+/* SMMU_IDR1 */
+#define SMMUV3_IDR1_SIDSIZE_OFF  (0)
+#define SMMUV3_IDR1_SIDSIZE_LEN  (6)
+
+/* SMMU_IDR5 */
+#define SMMUV3_IDR5_OAS_OFF      (0)
+#define SMMUV3_IDR5_OAS_LEN      (3)
+#define SMMUV3_IDR5_GRAN4K_BIT   (1U << 4)
+
+/* SMMU_CR0 / CR0ACK */
+#define SMMUV3_CR0_SMMUEN        (1U << 0)
+#define SMMUV3_CR0_EVENTQEN      (1U << 2)
+#define SMMUV3_CR0_CMDQEN        (1U << 3)
+
+/* SMMU_CR1 cacheability/shareability for table & queue accesses */
+#define SMMUV3_CR1_INSH          (0x3) /* inner shareable */
+#define SMMUV3_CR1_WBCACHE       (0x1) /* normal WB cacheable */
+#define SMMUV3_CR1_DEFAULT                                   \
+    ((SMMUV3_CR1_INSH << 10) | (SMMUV3_CR1_WBCACHE << 8) |   \
+        (SMMUV3_CR1_WBCACHE << 6) |    /* table: SH/OC/IC */ \
+        (SMMUV3_CR1_INSH << 4) | (SMMUV3_CR1_WBCACHE << 2) | (SMMUV3_CR1_WBCACHE)) /* queue */
+
+/* SMMU_GBPA */
+#define SMMUV3_GBPA_UPDATE                  (1U << 31)
+#define SMMUV3_GBPA_ABORT                   (1U << 20)
+
+/* SMMU_IRQ_CTRL / _CTRLACK (wired-SPI delivery of fault interrupts) */
+#define SMMUV3_IRQ_CTRL_GERROR              (1U << 0)
+#define SMMUV3_IRQ_CTRL_PRIQ                (1U << 1)
+#define SMMUV3_IRQ_CTRL_EVENTQ              (1U << 2)
+
+/* SMMU_STRTAB_BASE / _CFG */
+#define SMMUV3_STRTAB_BASE_RA               (1ULL << 62)
+#define SMMUV3_STRTAB_BASE_ADDR_MASK        BIT64_MASK(6, 46)
+#define SMMUV3_STRTAB_BASE_CFG_FMT_LINEAR   (0x0U << 16)
+#define SMMUV3_STRTAB_BASE_CFG_LOG2SIZE_OFF (0)
+
+/* SMMU_CMDQ_BASE / EVENTQ_BASE */
+#define SMMUV3_Q_BASE_RA                    (1ULL << 62)
+#define SMMUV3_Q_BASE_ADDR_MASK             BIT64_MASK(5, 47)
+#define SMMUV3_Q_BASE_LOG2SIZE_MASK         (0x1FULL)
+/* PROD/CONS wrap bit sits just above the index field */
+#define SMMUV3_Q_WRAP(log2size)             (1U << (log2size))
+#define SMMUV3_Q_IDX_MASK(log2size)         (SMMUV3_Q_WRAP(log2size) - 1U)
+
+/* ----------------------------------------------------------------------- */
+/* Stream Table Entry (8 x 64-bit). Stage-2-only / bypass fields.           */
+/* Bit positions per ARM IHI 0070.                                          */
+/* ----------------------------------------------------------------------- */
+
+#define SMMUV3_STE_DWORDS                   (8)
+
+struct smmuv3_ste {
+    uint64_t data[SMMUV3_STE_DWORDS];
+} __attribute__((__aligned__(64)));
+
+/* word 0 */
+#define STE_W0_V                  (1ULL << 0)
+#define STE_W0_CONFIG_OFF         (1)
+#define STE_W0_CONFIG_BYPASS      (0x4ULL) /* S1 bypass, S2 bypass */
+#define STE_W0_CONFIG_S2          (0x6ULL) /* S1 bypass, S2 translate */
+
+/* word 1 (attribute overrides used in bypass STEs) */
+#define STE_W1_SHCFG_OFF          (44)
+#define STE_W1_SHCFG_USE_INCOMING (0x1ULL) /* pass master's shareability */
+
+/* word 2 (stage-2 config) */
+#define STE_W2_S2VMID_OFF         (0)
+#define STE_W2_S2T0SZ_OFF         (32)
+#define STE_W2_S2SL0_OFF          (38)
+#define STE_W2_S2IR0_OFF          (40)
+#define STE_W2_S2OR0_OFF          (42)
+#define STE_W2_S2SH0_OFF          (44)
+#define STE_W2_S2TG_OFF           (46)
+#define STE_W2_S2PS_OFF           (48)
+#define STE_W2_S2AA64             (1ULL << 51)
+#define STE_W2_S2AFFD             (1ULL << 53)
+#define STE_W2_S2R                (1ULL << 58) /* record stage-2 faults to evtq */
+#define STE_W2_WB                 (0x1ULL)     /* IR0/OR0: WB cacheable RA/WA */
+#define STE_W2_ISH                (0x3ULL)     /* SH0: inner shareable */
+#define STE_W2_TG_4K              (0x0ULL)
+
+/* word 3 stage-2 translation table base */
+#define STE_W3_S2TTB_MASK         BIT64_MASK(4, 48)
+
+/* ----------------------------------------------------------------------- */
+/* Command queue: 128-bit commands (2 x 64-bit).                            */
+/* ----------------------------------------------------------------------- */
+
+#define SMMUV3_CMDQ_ENT_DWORDS    (2)
+
+#define CMD_OP_CFGI_STE           (0x03ULL)
+#define CMD_OP_CFGI_STE_RANGE     (0x04ULL)
+#define CMD_OP_TLBI_S12_VMALL     (0x28ULL)
+#define CMD_OP_TLBI_NSNH_ALL      (0x30ULL)
+#define CMD_OP_SYNC               (0x46ULL)
+
+/* CMD_CFGI_STE / _RANGE: StreamID in [63:32] of dword0 */
+#define CMD_CFGI_SID_OFF          (32)
+#define CMD_CFGI_STE_LEAF         (1ULL << 0)  /* dword1 */
+#define CMD_CFGI_STE_RANGE_ALL    (31ULL << 0) /* dword1: Range=31 -> all */
+
+/* CMD_TLBI_S12_VMALL: VMID in [47:32] of dword0 */
+#define CMD_TLBI_VMID_OFF         (32)
+
+/* CMD_SYNC: CS (completion signal) = none; we poll CONS instead */
+#define CMD_SYNC_CS_NONE          (0x0ULL << 12)
+
+/* Event queue record: 4 x 64-bit. dword0 [7:0]=fault type, [63:32]=StreamID;
+ * dword2 = input (IPA) address for translation/permission faults. */
+#define SMMUV3_EVTQ_ENT_DWORDS    (4)
+#define SMMUV3_EVT_TYPE_MASK      (0xFFULL)
+#define SMMUV3_EVT_SID_OFF        (32)
+
+/* ----------------------------------------------------------------------- */
+/* Driver API (consumed by iommu.c).                                        */
+/* ----------------------------------------------------------------------- */
+
+void smmuv3_init(void);
+bool smmuv3_write_ste_s2(streamid_t sid, paddr_t root_pt, uint16_t vmid);
+bool smmuv3_write_ste_bypass(streamid_t sid);
+void smmuv3_poll_events(void); /* drain+log SMMU faults; safe before init */
+
+#endif                         /* __ARCH_SMMUV3_H__ */
--- a/src/arch/armv8/armv8-a/iommu.c
+++ b/src/arch/armv8/armv8-a/iommu.c
@@ -9,6 +9,67 @@
 #include <mem.h>
 #include <config.h>
 
+#if (SMMU_VERSION == 3)
+
+bool iommu_arch_init(void)
+{
+    if (cpu_is_master() && platform.arch.smmu.base) {
+        smmuv3_init();
+        return true;
+    }
+
+    return false;
+}
+
+/* Install a stage-2-only STE for every stream id covered by {mask, id}. */
+static bool iommu_vm_arch_add(struct vm* vm, streamid_t mask, streamid_t id)
+{
+    paddr_t rootpt;
+    mem_translate(&cpu()->as, (vaddr_t)vm->as.pt.root, &rootpt);
+
+    /*
+     * A zero mask is the common case (a single device). A non-zero mask means
+     * "every stream id that matches id under this mask" -- iterate the masked
+     * bits and bind each concrete stream id to the same stage-2 config. SMMUv3
+     * has a flat stream table (no stream-match registers), so we materialise
+     * one STE per id rather than relying on a hardware mask.
+     */
+    streamid_t base = id & (streamid_t)~mask;
+    for (streamid_t sub = 0;; sub++) {
+        streamid_t sid = base | (sub & mask);
+        if (!smmuv3_write_ste_s2(sid, rootpt, (uint16_t)vm->id)) {
+            return false;
+        }
+        if (sub == mask) {
+            break;
+        }
+    }
+
+    vm->io.prot.mmu.initialized = true;
+    return true;
+}
+
+bool iommu_arch_vm_add_device(struct vm* vm, streamid_t id)
+{
+    return iommu_vm_arch_add(vm, 0, id);
+}
+
+bool iommu_arch_vm_init(struct vm* vm, const struct vm_config* vm_config)
+{
+    vm->io.prot.mmu.initialized = false;
+
+    for (size_t i = 0; i < vm_config->platform.arch.smmu.group_num; i++) {
+        const struct smmu_group* group = &vm_config->platform.arch.smmu.groups[i];
+        if (!iommu_vm_arch_add(vm, group->mask, group->id)) {
+            return false;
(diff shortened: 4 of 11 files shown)


src/arch/armv8/armv8-a/smmuv3.c
```diff
@@ -0,0 +1,393 @@
...
+void smmuv3_init(void)
+{
+    smmu.lock = SPINLOCK_INITVAL;
+
+    /* Map page 0 (the registers in struct smmuv3_hw) AND page 1 (offset 0x10000,
+     * which holds the architectural EVENTQ_PROD/CONS). */
+    vaddr_t regs = mem_alloc_map_dev(&cpu()->as, SEC_HYP_GLOBAL, INVALID_VA,
+        platform.arch.smmu.base, NUM_PAGES(SMMUV3_PAGE1_EVENTQ_CONS + 4));
+    smmu.hw = (volatile struct smmuv3_hw*)regs;
+    smmu.evtq_prod = (volatile uint32_t*)(regs + SMMUV3_PAGE1_EVENTQ_PROD);
```
joaopeixoto13, line 260: `evtq_prod`/`evtq_cons` are published early in `smmuv3_init()`, but `smmu.evtq` is allocated later and `EVENTQ_BASE` programmed later still. `smmuv3_poll_events()` (called from `gic_handle()` on every CPU, every IRQ) gates only on `evtq_prod == NULL`, so an IPI taken by another core during init can read a stale PROD/CONS pair, and dereference `smmu.evtq` while it's still NULL.

Suggest a dedicated `ready` flag published last. Sketch (spans three spots, so not a one-click suggestion):

```c
/* in struct smmuv3_priv, alongside `coherent`: */
bool coherent;
bool ready;          /* set once init completes; gates the poll hook */
spinlock_t lock;

/* in smmuv3_poll_events(), replace the NULL gate: */
if (!smmu.ready) {
    return;
}

/* at the very end of smmuv3_init(), after SMMUEN is acked: */
smmu.ready = true;
```

src/arch/armv8/gic.c
```diff
@@ -107,6 +110,12 @@ void gic_handle()
             gicc_dir(ack);
         }
     }
+
+#if (SMMU_VERSION == 3)
+    /* Surface any SMMU translation faults (a refused DMA usually coincides with
+     * a device error IRQ). Cheap no-op when the event queue is empty. */
+    smmuv3_poll_events();
```
joaopeixoto13: This runs on every interrupt on every CPU and does two page-1 MMIO reads even in the common empty case: a measurable per-IRQ cost on interrupt-heavy workloads. It's a genuinely useful backstop where the wired eventq IRQ isn't reliably delivered, so I wouldn't drop it, but just consider scoping it. Cheapest option, restrict the poll to the master CPU (the wired IRQ handler already covers the general case):

```c
#if (SMMU_VERSION == 3)
    if (cpu_is_master()) {
        smmuv3_poll_events();
    }
#endif
```

src/arch/armv8/armv8-a/smmuv3.c
```diff
@@ -0,0 +1,393 @@
...
+#include <cpu.h>
+#include <mem.h>
+#include <cache.h>
+#include <string.h>
+#include <bit.h>
+#include <interrupts.h>
+
+/* Linear stream table cap: 2^8 = 256 entries (16 KiB). This comfortably covers
+ * typical single-level stream-ID spaces without needing a 2-level table. */
+#define SMMUV3_ST_LOG2_CAP   (8)
```
joaopeixoto13: The 256-entry cap assumes a small, dense stream-ID space. Stream IDs are integrator-defined and can be sparse or start high (PCIe Requester IDs especially). On a platform whose IDs exceed 255, every `smmuv3_write_ste_s2()` fails its bounds check and VM init aborts, with the only fix being to edit this constant. The table is linear, so letting a platform override it costs only memory. Suggestion makes it overridable from `platform.mk` (e.g. `-DSMMUV3_ST_LOG2_CAP=11`):

```suggestion
/* Linear stream table cap: 2^SMMUV3_ST_LOG2_CAP entries (default 2^8 = 256,
 * 16 KiB). A platform whose stream-ID space is sparse or starts high can
 * override this from its platform.mk (e.g. -DSMMUV3_ST_LOG2_CAP=11). */
#ifndef SMMUV3_ST_LOG2_CAP
#define SMMUV3_ST_LOG2_CAP   (8)
#endif
```

src/arch/armv8/armv8-a/smmuv3.c
```diff
@@ -0,0 +1,393 @@
...
+    smmuv3_drain_events();
+    spin_unlock(&smmu.lock);
+}
+
+/* Cheap poll hook called from the GIC IRQ path: if the SMMU event queue is
+ * non-empty, drain+log it. This makes faults visible even if the dedicated
+ * wired event IRQ isn't delivered. No-op until the driver is initialised. */
+void smmuv3_poll_events(void)
+{
+    if (smmu.evtq_prod == NULL) {
```
joaopeixoto13: Part of the init-race fix above. This is the gate that goes live too early. Once a `ready` flag exists, this line becomes `if (!smmu.ready) return;`. Flagging it here so the two spots are reviewed together.

src/arch/armv8/gic.c
```diff
@@ -157,9 +166,7 @@ void gicd_set_pend(irqid_t int_id, bool pend)
 }
 
 bool gicd_get_pend(irqid_t int_id)
-{
-    return (gicd->ISPENDR[GIC_INT_REG(int_id)] & GIC_INT_MASK(int_id)) != 0;
-}
+{ return (gicd->ISPENDR[GIC_INT_REG(int_id)] & GIC_INT_MASK(int_id)) != 0; }
```
joaopeixoto13: This looks like unrelated reformatting (the one-lined bodies don't match the surrounding style and aren't part of the SMMUv3 change). Worth reverting to keep the diff focused:

```suggestion
bool gicd_get_pend(irqid_t int_id)
{
    return (gicd->ISPENDR[GIC_INT_REG(int_id)] & GIC_INT_MASK(int_id)) != 0;
}
```

src/arch/armv8/gic.c
```diff
@@ -173,9 +180,7 @@ void gicd_set_act(irqid_t int_id, bool act)
 }
 
 bool gicd_get_act(irqid_t int_id)
```
joaopeixoto13, line 188: Same as `gicd_get_pend`

src/arch/armv8/armv8-a/smmuv3.c
```diff
@@ -0,0 +1,393 @@
...
+static void smmuv3_cmdq_sync(void)
+{
+    smmuv3_cmdq_push(CMD_OP_SYNC | CMD_SYNC_CS_NONE, 0);
+
+    for (size_t i = 0; i < SMMUV3_POLL_LIMIT; i++) {
+        if ((smmu.hw->CMDQ_CONS & SMMUV3_Q_IDX_MASK(SMMUV3_CMDQ_LOG2SIZE)) ==
+            (smmu.cmdq_prod & SMMUV3_Q_IDX_MASK(SMMUV3_CMDQ_LOG2SIZE))) {
+            return;
+        }
+    }
```
joaopeixoto13, line 143: The sync-completion check compares only the index bits and drops the wrap bit, so a completely full queue (PROD = CONS + size) is indistinguishable from empty. Not reachable today since every push sequence syncs within a few entries of a 128-deep queue, but latent if commands are ever batched before syncing. `cmdq_prod` already carries the full `{wrap, index}` (it's kept modulo 2×size), and hardware maintains the wrap bit in `CMDQ_CONS`, so comparing the full field is robust:

```suggestion
    uint32_t fullmask = (SMMUV3_Q_WRAP(SMMUV3_CMDQ_LOG2SIZE) << 1) - 1;
    for (size_t i = 0; i < SMMUV3_POLL_LIMIT; i++) {
        if ((smmu.hw->CMDQ_CONS & fullmask) == (smmu.cmdq_prod & fullmask)) {
            return;
        }
    }
```

src/arch/armv8/armv8-a/inc/arch/smmuv3.h
```diff
@@ -0,0 +1,216 @@
...
+
+/* SMMU_IDR1 */
+#define SMMUV3_IDR1_SIDSIZE_OFF   (0)
+#define SMMUV3_IDR1_SIDSIZE_LEN   (6)
+#define SMMUV3_IDR1_TABLES_PRESET (1U << 30)
+#define SMMUV3_IDR1_QUEUES_PRESET (1U << 29)
+#define SMMUV3_IDR1_CMDQS_OFF     (21)
+#define SMMUV3_IDR1_CMDQS_LEN     (5)
+#define SMMUV3_IDR1_EVTQS_OFF     (16)
+#define SMMUV3_IDR1_EVTQS_LEN     (5)
```
joaopeixoto13: These capability macros are defined but never used

joaopeixoto13 (changes requested):
Thanks a lot for this. An SMMUv3 driver has been sitting in our "someday" pile for some time, so it's great to see a clean, focused driver actually land!

Did a close pass and two things I'd want sorted before moving on, and a couple of design questions that aren't blockers.

### Functional issues

1. **Init-order race on the poll hook.** `smmu.evtq_prod` gets published near the top of `smmuv3_init()`, right after the `mem_alloc_map_dev()`, but `smmu.evtq` isn't allocated and `EVENTQ_BASE` isn't programmed until much further down. `smmuv3_poll_events()` only checks `evtq_prod != NULL` and fires from `gic_handle()` on every core for every IRQ, so any IPI landing on a secondary CPU while the boot CPU is still mid-init will latch a stale PROD/CONS delta and walk `smmu.evtq` while it's still NULL. Cleanest fix is a `ready` flag flipped last, once `SMMUEN` is acked, with the poll hook gated on that instead of on the pointer. Left specifics inline.

2. **No SMMU TLBI on stage-2 remap.** You issue `TLBI_S12_VMALL` at bind time, but nothing hangs off the stage-2 invalidation path (`core/inc/tlb.h`, the `// TODO: inval iommu tlbs` sites), so `TLBI VMALLS12E1IS` on the CPU side never reaches the SMMU's TLBs. Fine as long as every device mapping is static, which is almost certainly why it comes up clean, but the moment a bound VM's stage-2 gets touched the SMMU keeps hitting stale walk-cache/TLB entries and translating against the old tables. This PR is what finally makes those TODOs actionable, so it's the natural spot to hook SMMU invalidation into that path, or at least to spell out the static-mapping assumption in the docs.

### Design questions (not blockers)

- **BYPASS-by-default vs abort-by-default.** Any streamID without a group stays parked on a BYPASS STE, i.e. free-for-all DMA into any VM's PA space. Reasonable for a transparent firmware handoff, but it means the SMMU is acting as an opt-in translator rather than an isolation boundary: a master nobody enumerated sails through instead of getting terminated. Once a platform's streamID map is fully known you'd usually want the unbound default to be abort (invalid STE, faults land in the event queue). Any appetite for a build/per-platform knob to flip the default?

- **Single instance.** One `static struct smmuv3_priv smmu` plus a single `platform.arch.smmu` bakes in one SMMU per SoC. Multi-SMMU parts are increasingly common, and a streamID is only unique within a given instance's stream table, so the config eventually needs a per-group "which SMMU" selector. Nothing to do here, just worth not boxing the interface in now.

Rest is smaller stuff and a few latent-but-not-yet-reachable bits, all inline. Thanks again for putting this up.

src/arch/armv8/gic.c
rabara: Agree, I will update the PR with this change.

src/arch/armv8/armv8-a/smmuv3.c
rabara: Good point - I will update the PR code , so it can be overridable from platform.mk.

src/arch/armv8/armv8-a/inc/arch/smmuv3.h
rabara: I will remove these unused macros.

rabara:
Thanks for the review @joaopeixoto13 

> Thanks a lot for this. An SMMUv3 driver has been sitting in our "someday" pile for some time, so it's great to see a clean, focused driver actually land!
> 
> Did a close pass and two things I'd want sorted before moving on, and a couple of design questions that aren't blockers.
> 
> ### Functional issues
> 1. **Init-order race on the poll hook.** 

I agree with your suggestion and will update the code in this PR. 
Additionally I am also thinking of adding a release barrier (fence_sync_write() / DSB) before publishing ready so the store can't be reordered ahead of the queue setup as seen by other observers — the flag alone isn't enough on a weakly-ordered core.

> 2. **No SMMU TLBI on stage-2 remap.** 

Agreed this is a real gap. It's not reachable in the current static-partitioning model. 
My preference is to document the static-mapping assumption explicitly in this PR and track the full invalidation hook in future PR. What is your opinion? 

I foresee the testing would be a major challenge as well. 
 
> ### Design questions (not blockers)
> * **BYPASS-by-default vs abort-by-default.** 

The knob is cheap, low-risk, and strictly increases capability while keeping the safe default — worth adding. 
How about per-platform knob (e.g. -DSMMUV3_DEFAULT_ABORT from platform.mk) that flips the unbound-stream default from BYPASS to an abort STE, while keeping BYPASS as the default so the transparent firmware handoff (and bring-up on platforms without a fully-enumerated streamID map) is unaffected. The cost is just which STE template the unbound entries get. I'll keep GBPA in BYPASS during the reconfiguration window regardless, so in-flight transactions during handoff don't fault.

> * **Single instance.** 

Fair point. I'll leave single-instance as-is for now, but I agree the interface shouldn't be boxed in — a per-group "which SMMU" selector is the right eventual shape given stream IDs are only unique within an instance's stream table. 
Shall we add a note in code about this as a known limitation so multi-SMMU support can be layered on without an interface break? Or you have some other thought?

src/arch/armv8/armv8-a/smmuv3.c
rabara: Agree - I will update the PR.

src/arch/armv8/gic.c
rabara: This must be due to my local clang-format check, will fix in PR.

src/arch/armv8/armv8-a/smmuv3.c
rabara: Agee - I will add ready flag along with release barrier before publishing _ready_

```
/* smmuv3.c — struct */
struct smmuv3_priv {
    ...
    bool coherent;
    volatile bool ready;   /* published last; gates the poll hook */
};

```

```
/* smmuv3_poll_events() */
void smmuv3_poll_events(void)
{
    if (!smmu.ready) {
        return;
    }
    /* Acquire: order the flag load before the queue-state loads below.
     * A control dependency alone doesn't order load→load on ARMv8. */
    fence_ord_read();           /* DMB ISHLD */
    uint32_t prod = *smmu.evtq_prod;
}
```

```
/* end of smmuv3_init(), after the CR0ACK poll loop */

 /* Release: make every prior store (evtq pointer, zeroed queue memory,
 * PROD/CONS shadow init) visible to other observers before the flag.
 * Without this, another core can observe ready==true while smmu.evtq
 * is still NULL from its point of view.
 */
	fence_ord_write();          /* DMB ISHST — store→store ordering */
	smmu.ready = true;
	
```

joaopeixoto13:
> Thanks for the review @joaopeixoto13
> 
> > Thanks a lot for this. An SMMUv3 driver has been sitting in our "someday" pile for some time, so it's great to see a clean, focused driver actually land!
> > Did a close pass and two things I'd want sorted before moving on, and a couple of design questions that aren't blockers.
> > ### Functional issues
> > 
> > 1. **Init-order race on the poll hook.**
> 
> I agree with your suggestion and will update the code in this PR. Additionally I am also thinking of adding a release barrier (fence_sync_write() / DSB) before publishing ready so the store can't be reordered ahead of the queue setup as seen by other observers — the flag alone isn't enough on a weakly-ordered core.

Yeah, agreed. Just note it needs a barrier on both sides, not just the writer. The `if (!ready) return;` check only orders later stores, not later loads on ARM, so another core can still read `smmu.evtq` / `*evtq_prod` before it sees `ready == true`. Easiest fix is to make `ready` a load-acquire (`LDAR`) on the reader, or put a `DMB LD` right after the gate. Then it's fully closed.

> > 2. **No SMMU TLBI on stage-2 remap.**
> 
> Agreed this is a real gap. It's not reachable in the current static-partitioning model. My preference is to document the static-mapping assumption explicitly in this PR and track the full invalidation hook in future PR. What is your opinion?
> 
> I foresee the testing would be a major challenge as well.

Fine to split it. Nothing hits that path in the static model today, so documenting the assumption now and doing the real hook later works for me. Two small asks: drop a note at the two `// TODO: inval iommu tlbs` lines in `core/inc/tlb.h` (not just the driver) so anyone adding dynamic remap later runs into it, and open an issue to track the actual hook.

And yeah, testing is the hard bit. You can't really trigger the stale-TLB case without remapping an already-bound VM's stage-2, which the static model never does, so the test only makes sense once the invalidation code exists. Fine to defer.

> 
> > ### Design questions (not blockers)
> > 
> > * **BYPASS-by-default vs abort-by-default.**
> 
> The knob is cheap, low-risk, and strictly increases capability while keeping the safe default — worth adding. How about per-platform knob (e.g. -DSMMUV3_DEFAULT_ABORT from platform.mk) that flips the unbound-stream default from BYPASS to an abort STE, while keeping BYPASS as the default so the transparent firmware handoff (and bring-up on platforms without a fully-enumerated streamID map) is unaffected. The cost is just which STE template the unbound entries get. I'll keep GBPA in BYPASS during the reconfiguration window regardless, so in-flight transactions during handoff don't fault.

Sounds good, that's exactly what I'd want: `-DSMMUV3_DEFAULT_ABORT` per platform, BYPASS as the default, GBPA staying in BYPASS during the reconfig window. All good.

> 
> > * **Single instance.**
> 
> Fair point. I'll leave single-instance as-is for now, but I agree the interface shouldn't be boxed in — a per-group "which SMMU" selector is the right eventual shape given stream IDs are only unique within an instance's stream table. Shall we add a note in code about this as a known limitation so multi-SMMU support can be layered on without an interface break? Or you have some other thought?

Yeah, a code note is enough. I'd put it in two spots: at `static struct smmuv3_priv smmu` (deliberately single-instance) and at the `smmu_group` struct (where a per-group instance selector would go later). That way someone can add multi-SMMU support down the line without breaking the interface. No need to do any of it now.

src/arch/armv8/armv8-a/smmuv3.c
rabara: Done

src/arch/armv8/gic.c
rabara: Done

src/arch/armv8/armv8-a/smmuv3.c
rabara: Done

src/arch/armv8/armv8-a/smmuv3.c
rabara: Done

src/arch/armv8/gic.c
rabara: Done

src/arch/armv8/gic.c
rabara: Done

src/arch/armv8/armv8-a/smmuv3.c
rabara: Done

src/arch/armv8/armv8-a/inc/arch/smmuv3.h
rabara: Done

src/arch/armv8/armv8-a/smmuv3.c
```diff
@@ -0,0 +1,458 @@
...
+#include <cache.h>
+#include <string.h>
+#include <bit.h>
+#include <interrupts.h>
+
+/* Linear stream table cap: 2^SMMUV3_ST_LOG2_CAP entries (default 2^8 = 256,
+ * 16 KiB). A platform whose stream-ID space is sparse or starts high can
+ * override this from its platform.mk (e.g. -DSMMUV3_ST_LOG2_CAP=11). */
+#ifndef SMMUV3_ST_LOG2_CAP
+#define SMMUV3_ST_LOG2_CAP (8)
```
josecm, line 31: I know the platform.mk override came out of the earlier discussion, but I'd rather not have a compile time cap at all. There is no reason 2^8 covers whatever platforms we end up running on, and any PCIe derived stream IDs blow way past what a linear table can reasonably hold. Since the full config is available when the driver initializes, can we size the table from it instead? Scan every VM's smmu groups and device ids, take the largest (for groups, `id | mask`), round up to a power of two and clamp to SIDSIZE. We could eventually do it completely statically using the config scritps to generate macros. Anything above the used ID space then fail-closes with C_BAD_STREAMID without spending STEs on it. Also please add a TODO noting the proper fix for large or sparse StreamID spaces is a 2-level stream table when SMMU_IDR0.ST_LEVEL allows it.

@joaopeixoto13 wdyt, does the config derived sizing work for you?

src/arch/armv8/armv8-a/smmuv3.c
```diff
@@ -0,0 +1,458 @@
...
+        if ((*reg & mask) == val) {
+            return 0;
+        }
+    }
+    return -1;
+}
+
+static void smmuv3_check_features(void)
+{
+    uint32_t aidr = smmu.hw->AIDR & 0xF;
```
josecm, line 93: Maybe we could: 
```suggestion
    uint32_t aidr = smmu.hw->AIDR & SMMUV3_AIDR_ArchMinorRev_MASK;
```

src/arch/armv8/armv8-a/smmuv3.c
```diff
@@ -0,0 +1,458 @@
...
+#endif
+
+/* Default STE for streams no VM has bound. BYPASS by default so the firmware
+ * handoff stays transparent (masters keep working with incoming attributes). A
+ * platform with a fully-known streamID map can flip the default to ABORT --
+ * turning the SMMU into an isolation boundary -- via -DSMMUV3_DEFAULT_ABORT in
+ * its platform.mk. Note this only changes the unbound default; GBPA is kept in
+ * BYPASS during the reconfiguration window regardless (see smmuv3_init), so
+ * in-flight transactions during handoff never fault. */
+static void smmuv3_ste_fill_default(struct smmuv3_ste* e)
```
josecm, line 186: Picking up the bypass vs abort question from the earlier discussion: I'd go all the way and make abort the only behavior, dropping `SMMUV3_DEFAULT_ABORT` and the GBPA bypass during the init window with it. The smmuv2 driver already faults unmatched streams (USFCFG), and with bypass the SMMU stops being an isolation boundary for anything the configs didn't bind. It's also inconsistent: stream IDs above the table size already abort with C_BAD_STREAMID, so whether an unbound master is contained depends on where its ID happens to fall. @rabara  Is there a concrete use case you have or can think of that requires this bypass? If something ever requires it, I'd rather have an explicit per-stream bypass in the config than a global default.

@joaopeixoto13 you raised the original question, agree with going abort only?

src/arch/armv8/armv8-a/smmuv3.c
```diff
@@ -0,0 +1,458 @@
...
+
+    /* Map page 0 (the registers in struct smmuv3_hw) AND page 1 (offset 0x10000,
+     * which holds the architectural EVENTQ_PROD/CONS). */
+    vaddr_t regs = mem_alloc_map_dev(&cpu()->as, SEC_HYP_GLOBAL, INVALID_VA,
+        platform.arch.smmu.base, NUM_PAGES(SMMUV3_PAGE1_EVENTQ_CONS + 4));
+    smmu.hw = (volatile struct smmuv3_hw*)regs;
+    smmu.evtq_prod = (volatile uint32_t*)(regs + SMMUV3_PAGE1_EVENTQ_PROD);
+    smmu.evtq_cons = (volatile uint32_t*)(regs + SMMUV3_PAGE1_EVENTQ_CONS);
+
+    smmu.coherent = (smmu.hw->IDR0 & SMMUV3_IDR0_COHACC_BIT) != 0;
```
josecm, line 263: I see `coherent` is only used to decide whether to flush the driver's own structures (stream table and queues). But COHACC covers every SMMU-originated access, including the stage 2 walks, and those read the VM page tables, which Bao writes cacheable and never cleans since CPU walkers are coherent. So on a COHACC=0 SMMU the queues would work but the walks would read stale PTEs. I'd make this a hard requirement: ERROR in `smmuv3_check_features` when COHACC is 0, and drop the non-coherent path altogether, keeping only the barriers. At some point we can add the infrastructure to propagate the core's TLB maintenance to the IOMMU and lift this, but until that exists coherency is a must. Btw, does your Agilex5 integration tie COHACC high? On MMU-600 this depends on the `sup_cohacc` tie-off.

src/arch/armv8/armv8-a/smmuv3.c
```diff
@@ -0,0 +1,458 @@
...
+    uint64_t ps = (uint64_t)parange;
+
+    uint64_t w2 = ((uint64_t)vmid << STE_W2_S2VMID_OFF) | ((uint64_t)t0sz << STE_W2_S2T0SZ_OFF) |
+        (sl0 << STE_W2_S2SL0_OFF) | (STE_W2_WB << STE_W2_S2IR0_OFF) |
+        (STE_W2_WB << STE_W2_S2OR0_OFF) | (STE_W2_ISH << STE_W2_S2SH0_OFF) |
+        (STE_W2_TG_4K << STE_W2_S2TG_OFF) | (ps << STE_W2_S2PS_OFF) | STE_W2_S2AA64 |
+        STE_W2_S2AFFD | STE_W2_S2R;
+
+    spin_lock(&smmu.lock);
+    struct smmuv3_ste* e = &smmu.st[sid];
```
josecm, line 423: Nothing prevents two VMs from binding the same stream id: the second write just replaces the STE and the device silently moves to the last VM that bound it. The v2 driver ERRORs on this (`smmu_compatible_sme_exists`). Since the old STE is right here, checking for a valid entry with a different S2VMID before writing and ERRORing would be enough. We could also return success if the ste we are writing for that sid matches what is already there. This would be useful in platforms where sids are programable and we decide to set all sid of devices assigned to the same VM with the same sid

src/arch/armv8/armv8-a/iommu.c
```diff
@@ -9,6 +9,67 @@
...
+    vm->io.prot.mmu.initialized = true;
+    return true;
+}
+
+bool iommu_arch_vm_add_device(struct vm* vm, streamid_t id)
+{
+    return iommu_vm_arch_add(vm, 0, id);
+}
+
+bool iommu_arch_vm_init(struct vm* vm, const struct vm_config* vm_config)
```
josecm, line 57: The v2 path merges the platform and VM config `global_mask` into every match, but here both are silently ignored (the fields still exist in the generic structs). I'm actually not sure the concept translates: SMMUv3 has no stream match registers, the StreamID indexes the table directly, so there is no hardware notion of don't care bits, and the v2 use case (TBU or implementation bits folded into the stream id, like tx2 and imx8qm) doesn't carry over to v3 integrations. Emulating it would mean materializing an STE for every masked combination, which seems like waste for a concept with no v3 rationale. I'd remove the `global_mask` fields from the platform and VM smmu structs under SMMU_VERSION conditionals, so a config that sets it on a v3 platform fails to build instead of silently changing meaning. Am I missing a v3 scenario where this still makes sense?

src/arch/armv8/armv8-a/smmuv3.c
```diff
@@ -0,0 +1,458 @@
...
+{
+    for (size_t i = 0; i < SMMUV3_POLL_LIMIT; i++) {
+        if ((*reg & mask) == val) {
+            return 0;
+        }
+    }
+    return -1;
+}
+
+static void smmuv3_check_features(void)
```
josecm, line 91: Following up on the stage 2 remap point from the earlier discussion: I'd rather solve it with broadcast TLB maintenance than with software hooks or a documented assumption. The STEs already tag entries with the CPU VMID, so if the SMMU participates in DVM, the TLBIs bao already issues when mapping or unmapping a VM's stage 2 reach the SMMU too, and runtime changes just work with zero driver involvement.

Two things are needed. Check IDR0.BTM here and ERROR without it, same as the v2 driver does for its broadcast bit. And write CR2 = 0 during init, since the reset value of CR2.PTM is not guaranteed and PTM=1 makes the SMMU ignore broadcasts even on BTM hardware (zeroing also gives VMW=0, exact VMID matching). CFGI_STE at bind stays, config caches are not covered by broadcast, but the bind time TLBI_S12_VMALL becomes redundant, and the long comments this PR adds to tlb.h can shrink to a note that runtime invalidation rides on DVM. Does Agilex5 tie `sup_btm` high btw?

@joaopeixoto13 does this address your remap concern?

src/arch/armv8/arch.mk
```diff
@@ -17,6 +17,11 @@ arch_profile_sub_dir:=$(arch_profile_dir)/$(ARCH_SUB)
 src_dirs+=$(arch_profile_sub_dir)
 
 arch-cppflags+=-DGIC_VERSION=$(GIC_VERSION)
+
+# Default to SMMUv2 so platforms that don't care about the IOMMU (or predate
+# the v2/v3 split) keep compiling. Platforms select v3 in their platform.mk.
+SMMU_VERSION?=2
```
josecm, line 23: As it stands nothing in tree builds this driver: no platform sets SMMU_VERSION:=3 and the Agilex5 description has no smmu base or interrupt. Now that the Agilex5 platform is merged, can you add the enablement in this same PR? SMMU_VERSION:=3 in its platform.mk plus whatever is needed in the platform description struct. Without an in-tree user CI never compiles smmuv3.c and it will rot with the first refactor that touches anything around it.

src/arch/armv8/armv8-a/smmuv3.c
```diff
@@ -0,0 +1,458 @@
...
+    /* Publish "ready" last, once SMMUEN is acked and everything the poll hook
+     * touches (evtq buffer, EVENTQ_BASE, the PROD/CONS pointers and their reset
+     * values) is live. Release barrier first so those prior stores are visible
+     * to any other core before it observes ready==true; the poll hook does the
+     * matching acquire. Without this a secondary CPU taking an IRQ mid-init
+     * could see ready==true while smmu.evtq is still NULL from its viewpoint. */
+    fence_ord_write();
+    smmu.ready = true;
+
+    INFO("smmuv3: enabled, %d stream-table entries, %s table walks", (int)smmu.st_entries,
```
josecm, line 375: Missing the trailing `\n`, here and in every other log message in the driver (console_printk doesn't add one).

src/arch/armv8/armv8-a/inc/arch/iommu.h
```diff
@@ -7,11 +7,23 @@
...
 
+#if (SMMU_VERSION == 3)
+/*
+ * SMMUv3 is stage-2-only here: each bound stream gets a Stream Table Entry that
+ * points directly at the VM's stage-2 page table (vm->as.pt.root) tagged with
+ * the VM's id as S2VMID. No per-VM context banks, so the only per-VM state we
+ * track is whether the STE(s) have been installed.
+ */
+struct iommu_vm_arch {
+    bool initialized;
```
josecm, line 20: This is written but never read anywhere. I'd drop it.

src/arch/armv8/armv8-a/smmuv3.c
```diff
@@ -0,0 +1,458 @@
...
+        smmu.coherent ? "coherent" : "non-coherent");
+}
+
+bool smmuv3_write_ste_bypass(streamid_t sid)
```
josecm, line 379: Nothing calls this. If we go abort only for unbound streams it becomes the building block for an explicit per-stream bypass, but until that exists I'd drop it.

src/arch/armv8/armv8-a/smmuv3.c
```diff
@@ -0,0 +1,458 @@
...
+    }
+
+    /* Route the event-queue fault interrupt (wired SPI) to a handler that drains
+     * + logs faults, so a refused DMA is visible instead of silent. */
+    if (platform.arch.smmu.interrupt_id != 0) {
+        irqid_t eid =
+            interrupts_reserve(platform.arch.smmu.interrupt_id, smmuv3_eventq_irq_handler);
+        if (eid != INVALID_IRQID) {
+            interrupts_cpu_enable(eid, true);
+            smmu.hw->IRQ_CTRL = SMMUV3_IRQ_CTRL_EVENTQ | SMMUV3_IRQ_CTRL_GERROR;
```
josecm, line 343: GERROR is enabled here but there's no handler or ack path for it, so a command queue error only surfaces as the CMD_SYNC timeout WARNING. Either handle it minimally, log plus ack, or don't enable it.

src/arch/armv8/armv8-a/smmuv3.c
```diff
@@ -0,0 +1,458 @@
...
+
+/* Drain + log the event queue. Each record is a translation/permission/etc.
+ * fault: a DMA the SMMU refused. On a correctly-bound, well-behaved system this
+ * stays empty; any line here is a master accessing memory it isn't allowed to
+ * (the IOMMU doing its job) -- so we surface it loudly. Caller holds smmu.lock. */
+static void smmuv3_drain_events(void)
+{
+    uint32_t wrapmask = (SMMUV3_Q_WRAP(SMMUV3_EVTQ_LOG2SIZE) << 1) - 1;
+    uint32_t idxmask = SMMUV3_Q_IDX_MASK(SMMUV3_EVTQ_LOG2SIZE);
+    uint32_t prod = *smmu.evtq_prod & wrapmask;
```
josecm, line 203: The overflow flag in EVENTQ_PROD (bit 31) is masked out and ignored, so an overflowing queue silently loses fault records. Worth at least a WARNING when it's set.

src/arch/armv8/armv8-a/smmuv3.c
```diff
@@ -0,0 +1,458 @@
...
+    smmu.hw->CR0 = SMMUV3_CR0_CMDQEN | SMMUV3_CR0_EVENTQEN;
+    if (smmuv3_poll(&smmu.hw->CR0ACK, SMMUV3_CR0_EVENTQEN, SMMUV3_CR0_EVENTQEN) != 0) {
+        ERROR("smmuv3: EVENTQEN not acknowledged");
+    }
+
+    /* Route the event-queue fault interrupt (wired SPI) to a handler that drains
+     * + logs faults, so a refused DMA is visible instead of silent. */
+    if (platform.arch.smmu.interrupt_id != 0) {
+        irqid_t eid =
+            interrupts_reserve(platform.arch.smmu.interrupt_id, smmuv3_eventq_irq_handler);
```
josecm, line 340: On the fault path as a whole: since we run the terminate model, an aborted DMA already surfaces inside the owning VM through the device itself, bus error status, the device's own error interrupt, a timeout, and the guest driver has to handle those on bare metal anyway. The eventq SPI on the other hand has a single target, so whoever runs on the master cpu absorbs fault handling for every VM's devices, a cross VM interference channel. Hypervisor side fault observation is still valuable to detect a misbehaving VM, but smmuv2 already reports nothing and I'd rather we approach that in a structured way for all iommus in a separate PR, where we can also design around the interference. So for this one I'd drop the eventq interrupt and the gic.c poll entirely and keep the driver to what the isolation needs. @joaopeixoto13 this reverses the earlier conclusion on scoping the poll to the master cpu, wdyt?

josecm: Maybe the comment on EVENTQ_PROD comment loses the effect if we remove fault handling...

josecm (changes requested):

PR #384, opened 2026-07-24
fix(ci): free runner disk space before image pull

josecm:
## Problem

The `baoproject/bao` container image (5.5 GB compressed, ~9 GB extracted) no longer fits in the ~14 GB of free disk on GitHub-hosted runners, so jobs fail during image pull:

```
failed to register layer: write /opt/v850-toolchain/v850-elf/lib/soft-float/libstdc++exp.a: no space left on device
Error: Docker pull failed with exit code 1
```

With the `container:` job directive the image is pulled during job initialization, before any step runs, so no cleanup step can help while that directive is used.

## Change

Adds a composite action (`.github/actions/bao-container-run`) that:

1. Frees ~25 GB by removing preinstalled runner software the CI never uses (dotnet, Android SDK, GHC, CodeQL)
2. Runs the job's command inside `baoproject/bao:latest` via `docker run` with the workspace mounted

All jobs drop the `container:` directive and run their (unchanged) make invocations through this action. The `safe.directory` git config moved into the action since the checkout now happens on the host as a different user than the container's root; this covers the gitlint job's previous explicit step and the git usage of the other targets.

The disk cleanup adds roughly 1 to 2 minutes per job. If that proves annoying it can be overlapped with the image pull in a follow-up.

This is a stopgap until the image is split into per-architecture images with a common base layer, see bao-project/bao-ci#94.

## Testing

- actionlint clean on all modified workflows
- Local runs of the exact container command: `qemu-riscv64-virt` build, `license-check` and `gitlint` all pass inside `docker run` with a host-owned checkout
- The workflows in this PR run with these changes, so green checks here validate the fix on real runners

### DIFF
--- a/.github/actions/bao-container-run/action.yml
+++ b/.github/actions/bao-container-run/action.yml
@@ -1,0 +1,26 @@
+name: Run in Bao CI container
+description: >
+  Frees runner disk space and runs a command inside the baoproject/bao
+  container. Stopgap while the CI image is too large for the free disk on
+  GitHub-hosted runners (see bao-project/bao-ci#94): with the `container:`
+  directive the image is pulled before any step can clean up, so jobs mount
+  the workspace and run the container explicitly instead.
+inputs:
+  run:
+    description: Command to run inside the container
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: Free runner disk space
+      shell: bash
+      run: >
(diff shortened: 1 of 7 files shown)


sandro2pinto (approved):
All good from my side.

DavidMCerdeira (approved):

PR #391, opened 2026-08-07
fix(mem): fix hypervisor self-coloring boot hang

rabara:
## Summary
- Fix root page-pool bitmap relocation under hyp coloring (bitmap lives in the image via `pp_bitmap_alloc()`, not a standalone SEC_HYP_GLOBAL region).
- Maintain I/D cache coherency when relocating the colored image (clean to PoC before switch; `ic iallu` in `switch_space()`).

## Test
- [x] bao-demos `PLATFORM=qemu-aarch64-virt DEMO=baremetal` with `config.hyp.colors` set
- [x] Unpatched main hangs after U-Boot `go 0x50000000`; fixed build prints Bao banner and boots guest

### DIFF
--- a/src/arch/armv8/armv8-a/aarch64/relocate.S
+++ b/src/arch/armv8/armv8-a/aarch64/relocate.S
@@ -91,6 +91,18 @@ switch_space:
     isb
 
     /**
+     * Invalidate the instruction cache.
+     *
+     * This routine returns into a copy of the image the caller relocated with a data-side copy and
+     * cleaned to the PoC. The image virtual addresses are about to be repointed at that copy, so
+     * drop any instruction cache lines this PE holds for the new physical pages, otherwise the
+     * fetches after the switch may hit stale instructions.
+     */
+    ic iallu
+    dsb ish
+    isb
+
+    /**
      * Update TTBR
      */
 
--- a/src/core/mmu/mem.c
+++ b/src/core/mmu/mem.c
@@ -747,7 +747,7 @@ void mem_color_hypervisor(const paddr_t load_addr, struct mem_region* root_regio
     struct cpu* cpu_new;
     struct ppages p_cpu;
     struct ppages p_image;
-    struct ppages p_bitmap;
+    void* image_cpy = NULL;
 
     size_t image_load_size = (size_t)(&_image_load_end - &_image_start);
     size_t image_noload_size = (size_t)(&_image_end - &_image_load_end);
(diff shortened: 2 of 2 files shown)


josecm (approved):

josecm:
Thanks @rabara! This feature has not been used that much lately, so the issue slipped through when we modified the memory layout and allocations to accommodate MCU architectures. Just about to add a test infrastructure finally #363, so we should probably add tests which include setting up hypervisor colors.

Both fixes look good to me. I went through the allocation flow and the invariant the recopy leans on holds: every allocation happens before the barrier, so the refreshed image picks up the final bitmap along with the rest of the pool bookkeeping. The removed cleanup was also zeroing pages that by now belong to the free pool, so that was a corruption waiting to happen. The check that the bitmap lives inside the image is a nice tripwire for future changes to how it is allocated. On the cache side, the sequence is the right one: clean through the temporary mapping while the image virtual addresses still translate to the old pages, and the per PE icache invalidate in switch_space before the TTBR write.

src/core/mmu/mem.c
```diff
@@ -831,16 +837,18 @@ void mem_color_hypervisor(const paddr_t load_addr, struct mem_region* root_regio
...
+         *
+         * The bitmap lives inside the image, so the copy above already relocated it, but that
+         * snapshot is stale: it predates the page tables allocated to map the image in the new
+         * address space. Every CPU is past the barrier and no allocation is left to be done, so
+         * re-copy the image over the mapping copy_space() left behind to pick up the final bitmap,
+         * along with the rest of the page pool bookkeeping. Copying the image again takes no
+         * allocation, which a separate bitmap copy would, and that allocation would no longer be
+         * tracked by the bitmap being copied.
+         */
+        memcpy(image_cpy, &_image_start, image_size);
```
DavidMCerdeira, line 851: With this, we memcpy the whole image twice. Once in copy_space(), then again after the barrier to pick up the final bitmap. I think we can skip the first copy altogether (without skipping the allocation and mapping).

So I'm proposing we split the operations of  copy_space() (alloc+map and copy) to save an additional image memcpy on boot.

src/arch/armv8/armv8-a/aarch64/relocate.S
```diff
@@ -90,6 +90,18 @@ switch_space:
...
+     *
+     * This routine returns into a copy of the image the caller relocated with a data-side copy and
+     * cleaned to the PoC. The image virtual addresses are about to be repointed at that copy, so
+     * drop any instruction cache lines this PE holds for the new physical pages, otherwise the
+     * fetches after the switch may hit stale instructions.
+     */
+    ic iallu
+    dsb ish
+    isb
+
```
DavidMCerdeira, line 104: Could the `ic iallu` just go after the `tlbi alle2` in the existing block? They don't depend on each other and both do a dsb+isb

PR #392, opened 2026-08-07
fix: latent cache coherency issues in relocation and cache flush

josecm:
`cache_flush_range()` on armv8 ended with a `dmb`, which does not guarantee completion of the cache maintenance operations, while its callers rely on it: the relocation flow before other cores fetch the relocated image, the guest image flush, and the psci off state flush before a core powers down. Replace it with a `dsb`.

The riscv `switch_space()` never executed `fence.i`, so instruction fetch is not synchronized with the data-side copy of the relocated image, the same problem #391 fixes for aarch64. Execute it just before the `satp` write, while still fetching from the original pages. This requires enabling `zifencei` in the march string.

Both qemu-aarch64-virt and qemu-riscv64-virt builds pass.

### DIFF
--- a/src/arch/armv8/cache.c
+++ b/src/arch/armv8/cache.c
@@ -114,5 +114,6 @@ void cache_flush_range(vaddr_t base, size_t size)
         cache_addr += min_line_size;
     }
 
-    DMB(ish);
+    /* Ensure the cache maintenance operations complete before returning */
+    DSB(ish);
 }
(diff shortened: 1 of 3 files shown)


src/arch/armv8/cache.c
```diff
@@ -114,5 +114,10 @@ void cache_flush_range(vaddr_t base, size_t size)
         cache_addr += min_line_size;
     }
 
-    DMB(ish);
+    /**
+     * A dmb does not guarantee completion of the cache operations. Callers rely on the flush being
```
DavidMCerdeira: I don't see the need to keep this comment, at least the `dmb` vs `dsb` part

josecm: Addressed!

danielRep (approved):

DavidMCerdeira (approved):

ninolomata (approved):

malejo97 (approved):

PR #393, opened 2026-08-13
fix(riscv): fix multi-hart boot synchronization

ninolomata:
Fix a race where harts losing the boot lock could read `CPU_MASTER` before it was updated by the winning hart.

Losing harts now wait for the boot hart to complete bootstrap initialization instead of reading `CPU_MASTER`.


src/arch/riscv/boot.S
```diff
@@ -153,20 +153,23 @@ _boot_lock:
...
 1:
     lr.w    t2, (t0)
-    bnez    t2, 2f
+    bnez    t2, wait_for_bsp /* Another hart already won */
     sc.w    t2, t1, (t0)   
     bnez    t2, 1b 
     la      t0, CPU_MASTER
     STORE      a0, 0(t0)
-2:
+    j       init_bsp
```
josecm:
```suggestion
```

This is not needed as this only executes if !CPU_MASTER_FIXED, and the code before `init_bsp` only exists if CPU_MASTER_FIXED

josecm:
@ninolomata From what I understand the race condition is:
1. On the lottery mechanism (so !CPU_MASTER_FIXED), the cpu master already won the lottery but didn't yet set CPU_MASTER (it is somehow slowed down in lines (159 and 160).
2. A secondary cpu gets to 167 and read a unitilized CPU_MASTER which might be its own id and proceed to initialize global pages tables.

Is this analysis correct? If so, the fix makes sense. I'd just put 

```
    LD_SYM  t0, CPU_MASTER
    bne     a0, t0, wait_for_bsp  
```

directly inside the first `#if defined(CPU_MASTER_FIXED)`. No need for a separate one.

josecm:
@ninolomata sorry! Applied my own suggestion by mistake. Needed to force push to revert.

src/arch/riscv/boot.S
ninolomata: I applied a fix now to remove that.

ninolomata:
I move this code to the top and also removed LD_SYM because the CPU_MASTER is fixed and we have it on t1 register.

    LD_SYM  t0, CPU_MASTER
    bne     a0, t0, wait_for_bsp

josecm (approved):

malejo97 (approved):
LGTM.

PR #394, opened 2026-08-21
fix(armv8/vgic): correct addressing of forwarded register writes

josecm:
Four bugs in the vgic register emulation, plus a preparatory refactor of the message encoding.

- ref(armv8/vgic): pack vgic ipi messages in a bitfield union. Replaces the VGIC_MSG_* shift/mask macros with a union type. No functional change.
- fix(armv8/vgic): carry the redistributor id in forwarded register writes. The message hardcoded vgicr id 0, so a write to a private interrupt register of another vcpu's redistributor was applied to vcpu 0's interrupt instead of the intended one.
- fix(armv8/vgic): update the route at the emulation site. The router emulation merged partial word writes against a route value that could be behind an in-flight forward to a remote interrupt owner, so the second half of a split 64-bit guest write (all aarch32 guests write IROUTER in two 32-bit halves) could resurrect the old route and the update was silently lost. It also truncated the forwarded value to the 8-bit message val field, losing the IRM bit and any affinity above Aff0. The route and the physical routing are now updated at the access site under the interrupt lock, where merges always see the current value, and the message to the owner becomes a data-less request to re-evaluate the routing.
- fix(armv8/vgic): return the full 64-bit route on irouter reads. A doubleword read of IROUTER returned only the low word, truncating Aff3.
- fix(armv8/vgic): pin private interrupt ownership to its vcpu. On GICv3 private interrupts were initialized without an owner and the owner could later be yielded back to none, so a core emulating a write to another vcpu's redistributor could take ownership and handle the access in place, manipulating another vcpu's banked interrupt state from the wrong core. The owner is now set at init and never yielded, as already on GICv2, so remote redistributor writes are always forwarded to the owning core, where the vgicr id fix above makes them land on the right interrupt.

The forwarding fixes trigger whenever the written interrupt is owned by a vcpu on another cpu, which with the ownership pinning now includes every write to another vcpu's private interrupt registers. Build-tested for GICv3 and GICv2 on qemu-aarch64-virt, and GICv3 on mps3-an536 (aarch32). The route update path is exercised at runtime by a guest test that cycles an SPI's route across all cpus and verifies each delivery lands on the routed cpu.

### DIFF
--- a/src/arch/armv8/inc/arch/vgic.h
+++ b/src/arch/armv8/inc/arch/vgic.h
@@ -120,7 +120,7 @@ void vgic_send_sgi_msg(struct vcpu* vcpu, cpumap_t pcpu_mask, irqid_t int_id);
 size_t vgic_get_itln(const struct vgic_dscrp* vgic_dscrp);
 struct vgic_int* vgic_get_int(struct vcpu* vcpu, irqid_t int_id, vcpuid_t vgicr_id);
 void vgic_int_set_field(struct vgic_reg_handler_info* handlers, struct vcpu* vcpu,
-    struct vgic_int* interrupt, unsigned long data);
+    struct vgic_int* interrupt, unsigned long data, vcpuid_t vgicr_id);
 void vgic_emul_razwi(struct emul_access* acc, struct vgic_reg_handler_info* handlers,
     bool gicr_access, cpuid_t vgicr_id);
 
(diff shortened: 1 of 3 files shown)


josecm:
Pushed one more fix in 54a6eb44: on GICv3, private interrupts were initialized without an owner, so a core emulating a write to another vcpu's redistributor could take ownership and handle the access in place instead of forwarding it to the owning core. Private interrupt ownership is now pinned to the vcpu, as already on GICv2, so those writes always go through the owning core, where the vgicr id fix makes them land on the right interrupt.

danielRep (approved):

DavidMCerdeira (approved):

PR #398, opened 2026-08-31
fix(vgic): break the no-pending maintenance storm when all LRs are active

josecm:
Based on top of #334; only the last two commits are new, will rebase once it merges.

The NP maintenance condition is level sensitive: it stays asserted while NPIE is set and no list register holds a pending interrupt. If every LR is occupied by an active interrupt (guest acknowledged them all, e.g. nested handlers) while a pending interrupt sits on the spilled list, vgic_refill_lrs finds no free LR, so it neither injects the pending interrupt nor clears NPIE, and the maintenance interrupt re-asserts immediately on guest entry. The guest, the only agent that can free an LR by EOIing an active, never runs again: the VM livelocks. Reachable with guest interrupt nesting depth equal to the number of LRs (4 on GICv2 implementations like GIC-400) plus one more, lower priority, pending interrupt.

The fix restructures the refill to pick the delivery candidate before requiring a free LR, so the nothing-to-deliver path clears NPIE even with all LRs in use, and, when there is a pending candidate but no free LR, spills the lowest priority active LR to make room. This keeps the spec's recommendation of at least one resident pending interrupt (IHI 0048B.b 5.2.1) unconditional. The evicted active's EOI is handled by the existing EOICount/LRENP path, and the guest's preemption level is untouched since it is tracked by GICH_APR, not by the LR. A first commit re-keys NPIE arming on a pending interrupt actually sitting on the spilled list, so active only spills no longer arm it and a spilled active that later becomes pending does.

### DIFF
--- a/src/arch/armv8/inc/arch/vgic.h
+++ b/src/arch/armv8/inc/arch/vgic.h
@@ -47,6 +47,7 @@ struct vgic_int {
 #endif
     bool hw;
     bool in_lr;
+    bool in_spilled;
     bool enabled;
 };
 
--- a/src/arch/armv8/vgic.c
+++ b/src/arch/armv8/vgic.c
@@ -180,7 +180,12 @@ static void vgic_route(struct vcpu* vcpu, struct vgic_int* interrupt)
         vgic_add_lr(vcpu, interrupt);
     }
 
-    if (!interrupt->in_lr && vgic_int_has_other_target(vcpu, interrupt)) {
+    /**
+     * An active interrupt cannot be forwarded to another CPU — ownership
+     * cannot be yielded while the interrupt is active, so any IPI would
+     * be silently ignored by the recipient.
+     */
+    if (!interrupt->in_lr && !(interrupt->state & ACT) && vgic_int_has_other_target(vcpu, interrupt)) {
         struct cpu_msg msg = {
             (uint32_t)VGIC_IPI_ID,
             VGIC_ROUTE,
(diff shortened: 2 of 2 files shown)


DavidMCerdeira (commented):
In vgic_highest_prio_spilled the owner is ignored. An SPI on the VM-wide list can stay owned by another vcpu. If that entry is the highest-priority candidate for another vcpu's refill, eligible fails, the rescan returns the same entry, and the core spins until the owner moves on. This is not a regression, but skiping entries owned by another vcpu would fix the existing bug.


A second point is that, the NPIE storm can still happen when every LR is in the pending-and-active state.

`vgic_spill_lowest_prio_active_lr()` only accepts LRs with `lr_state == GICH_LR_STATE_ACT`, so a pending-and-active LR (State 0b11) is never a victim. But NP does not consider such an LR as pending either. IHI 0069G, `ICH_HCR_EL2.NPIE`:

> Enables the signaling of a maintenance interrupt when there are no List registers with the state field set to 0b01 (pending)

and `ICH_MISR_EL2.NP`:

> This maintenance interrupt is asserted when ICH_HCR_EL2.NPIE==1 and no List register is in pending state.

The GICv2 text (IHI 0048B.b, `GICH_HCR.NPIE` / `GICH_MISR.NP`) uses "no List register is in pending state" without spelling out the encoding, but "pending state" is the name of 0b01 there too, and 5.2.1 notes that only hypervisor-generated interrupts can be active and pending, which is exactly the case we produce.

So with `NUM_LRS` nested software interrupts that were each re-raised while active
(`vgic_write_lr()` writes PENDACT unchanged for non-HW entries), plus one more pending interrupt on the spilled list: `vgic_add_spilled()` arms NPIE, NP asserts because no LR is 0b01, the refill finds a candidate and no free LR, the victim search returns -1, and we exit with NPIE still set. Same livelock as before, just needing PENDACT instead of ACT in every LR.

This is also inconsistent with the existing victim selection in `vgic_add_lr()`, which uses
`lr_state & GICH_LR_STATE_ACT` and therefore already evicts pending-and-active LRs.

josecm:
@DavidMCerdeira I'll wait merging #334 before addressing your comments here.

Issue #399, opened 2026-08-31
Possible optimization of the allocation loop in function pp_alloc_clr


nkosmatov:
While working on formal verification of cache coloring and page coloring mechanisms in Bao, we discovered two possible optimizations in function `pp_alloc_clr`. Here is the first optimization. 

The outer loop (currently line 123 in commit 7448c66 in file src/core/mmu/mem.c) performs two search iterations:

-  the first one starting from `pool->last` until the end of the pool, and 
-  the second one starting from the beginning of the pool until the end of the pool. 

A proposed optimization would be to stop the second search iteration if the required set of pages does not start before `pool->last`. Indeed, since a required set of pages starting at or after `pool->last` was not found during the first iteration, it cannot be found in that part of the pool during the second iteration. 

We will propose a pull request with this optimization. 

We were able to prove a (slightly simplified) corrected version of cache coloring and page coloring mechanisms in Bao with our optimizations. Many thanks to Axel Amestoy who performed the verification of the optimized version. 

FYI @aamstoy

josecm:
The PR was merged. Thank you both @nkosmatov and @aamstoy for your contribution!

PR #400, opened 2026-09-01
opt(mem): avoid rescanning already visited pages in pp_alloc_clr

aamstoy:
In function `pp_alloc_clr` in file src/core/mmu/mem.c, the second search pass previously rescanned the whole pool, including a region already scanned by the first pass. This change restricts the second pass to the range actually left unsearched, removing redundant work. We introduce variable `first_index_limit` which contains the maximum possible offset of the first page of an acceptable set of pages. This variable is updated before the second `for` iteration to avoid rescanning already visited pages.

We were able to formally verify a slightly simplified version of the cache coloring and page coloring mechanisms which includes this optimization as well as another (mainly unrelated) one.

Issue: #399

### DIFF
--- a/src/core/mmu/mem.c
+++ b/src/core/mmu/mem.c
@@ -111,22 +111,26 @@ bool pp_alloc_clr(struct page_pool* pool, size_t n, colormap_t colors, struct pp
 
     /**
      * Lets start the search at the first available color after the last known free position to the
-     * top of the pool.
+     * top of the pool. During this first part of the search, we'll go until the end of the pool.
      */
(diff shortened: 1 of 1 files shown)


josecm:
@aamstoy thank you for this contribution! It looks pretty straightforward to me, so no comments on the code itself. The only thing I'd ask is to fix up the formatting. If you pull the ci submodule simply running `make format` and ammend the commit should be enough.

aamstoy:
Thank you, it should be better now.

josecm (approved):

danielRep (approved):

Issue #389, opened 2026-08-02
[BUG] GICv3: vgic_int_ptarget_mask wrong shift operand + uint8_t truncates cpumap


nks0005:
## Describe the bug

`vgic_int_ptarget_mask()` (`src/arch/armv8/vgicv3.c`) returns the pCPU mask that `vgic_route()` uses for `VGIC_ROUTE` IPIs. Two bugs:

https://github.com/bao-project/bao-hypervisor/blob/9d9321128ecd12286dd67c48b3d436913fbec403/src/arch/armv8/vgicv3.c#L34-L41

### 1) Non-broadcast shifts MPIDR instead of `phys_id`

```c
return (uint8_t)(1U << interrupt->phys.route);
```

`phys.route` is the physical MPIDR stored by `vgic_int_set_route()`, not a CPU bitmap index. On QEMU `virt` + GICv3, Aff0 wraps every 16 CPUs, so pCPU16 has MPIDR `0x100`. `1U << 0x100` is wrong / UB.

`set_route` itself is fine. Broadcast already clears with `phys_id`; only this branch uses the wrong value.

### 2) Return type is `uint8_t` while the caller wants a `cpumap_t`

Both branches cast to `uint8_t`:

```c
return (uint8_t)(cpu()->vcpu->vm->cpus & ~(1U << cpu()->vcpu->phys_id)); /* broadcast */
return (uint8_t)(1U << interrupt->phys.route);                           /* non-broadcast */
```

Caller in `vgic.c`:

```c
cpumap_t trgtlist = vgic_int_ptarget_mask(vcpu, interrupt) & ~(1UL << vcpu->phys_id);
```

Bits for `phys_id >= 8` are dropped. Even after fixing (1), `(uint8_t)(1U << 16)` is still `0`. Broadcast hits the same wall: `vm->cpus = 0x1ffff` becomes `0xff`, so CPU8–16 never get the IPI.

---

## Expected behavior

1. Non-broadcast mask bit = owning `phys_id` (reverse-map MPIDR → CPU, then `1 << phys_id`). Leave `phys.route` as MPIDR for `gicd_set_route()` / affinity compares.
2. Return a full-width mask (`cpumap_t`) on both paths.

---

### 1) Wrong operand

| Case | phys_id | phys.route | result |
|------|---------|------------|--------|
| `-smp 4` | 2 | 2 | works by accident |
| `-smp 17`, target pCPU16 | 16 | `0x100` | wrong mask / UB |

[...]

PR #403, opened 2026-09-05
fix(armv8): use phys_id for GICv3 vgic_int_ptarget_mask

nks0005:
vgic_int_ptarget_mask() builds the pCPU bitmap for VGIC_ROUTE IPIs. On GICv3, interrupt->phys.route is the physical MPIDR stored by vgic_int_set_route(), not a linear CPU index. Shifting it (e.g. 1U << 0x100 on pCPU16) is wrong and undefined when the shift exceeds the width of unsigned. The uint8_t return also truncates bits for phys_id >= 8, so broadcast masks drop CPUs 8 and above.

Map phys.route back to phys_id via cpu_id_to_mpidr() and return cpumap_t on both GICv3 and GICv2. Leave phys.route as MPIDR for gicd_set_route() and affinity compares.

Verified on QEMU virt (aarch64, GICv3, -smp 17): IROUTER of an SPI targeted at pCPU16 yields mask 1 << 16 instead of a truncated or shifted-MPIDR mask.

Issue: #389

### DIFF
--- a/src/arch/armv8/inc/arch/vgic.h
+++ b/src/arch/armv8/inc/arch/vgic.h
@@ -25,9 +25,9 @@ struct vgic_int {
     struct vcpu* owner;
 #if (GIC_VERSION != GICV2)
     unsigned long route;
-    union {
-        vcpuid_t redist;
+    struct {
         unsigned long route;
+        cpuid_t cpu;
     } phys;
 #endif
     spinlock_t lock;
--- a/src/arch/armv8/inc/arch/vgic.h
+++ b/src/arch/armv8/inc/arch/vgic.h
@@ -26,8 +26,11 @@ struct vgic_int {
 #if (GIC_VERSION != GICV2)
     unsigned long route;
     struct {
+        union {
+            cpuid_t redist;
+            cpuid_t cpu;
+        };
         unsigned long route;
-        cpuid_t cpu;
     } phys;
 #endif
     spinlock_t lock;
--- a/src/arch/armv8/inc/arch/vgic.h
+++ b/src/arch/armv8/inc/arch/vgic.h
@@ -126,7 +126,7 @@ void vgic_emul_razwi(struct emul_access* acc, struct vgic_reg_handler_info* hand
 
 /* interface for version specific vgic */
 bool vgic_int_has_other_target(struct vcpu* vcpu, struct vgic_int* interrupt);
-uint8_t vgic_int_ptarget_mask(struct vcpu* vcpu, struct vgic_int* interrupt);
+cpumap_t vgic_int_ptarget_mask(struct vcpu* vcpu, struct vgic_int* interrupt);
 void vgic_inject_sgi(struct vcpu* vcpu, struct vgic_int* interrupt, vcpuid_t source);
 
 #endif /* __VGIC_H__ */
--- a/src/arch/armv8/vgicv3.c
+++ b/src/arch/armv8/vgicv3.c
@@ -34,13 +34,9 @@ cpumap_t vgic_int_ptarget_mask(struct vcpu* vcpu, struct vgic_int* interrupt)
 {
     if (vgic_broadcast(vcpu, interrupt)) {
         return cpu()->vcpu->vm->cpus & ~(1UL << cpu()->vcpu->phys_id);
+    } else if (interrupt->phys.cpu != INVALID_CPUID) {
+        return (cpumap_t)(1UL << interrupt->phys.cpu);
     } else {
-        unsigned long route = interrupt->phys.route & MPIDR_AFF_MSK;
-        for (cpuid_t i = 0; i < platform.cpu_num; i++) {
-            if ((cpu_id_to_mpidr(i) & MPIDR_AFF_MSK) == route) {
-                return (cpumap_t)(1UL << i);
-            }
-        }
         return 0;
     }
 }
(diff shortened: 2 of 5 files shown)


src/arch/armv8/vgicv3.c
```diff
@@ -30,12 +30,18 @@ bool vgic_int_has_other_target(struct vcpu* vcpu, struct vgic_int* interrupt)
...
-uint8_t vgic_int_ptarget_mask(struct vcpu* vcpu, struct vgic_int* interrupt)
+cpumap_t vgic_int_ptarget_mask(struct vcpu* vcpu, struct vgic_int* interrupt)
 {
     if (vgic_broadcast(vcpu, interrupt)) {
-        return (uint8_t)(cpu()->vcpu->vm->cpus & ~(1U << cpu()->vcpu->phys_id));
+        return cpu()->vcpu->vm->cpus & ~(1UL << cpu()->vcpu->phys_id);
     } else {
-        return (uint8_t)(1U << interrupt->phys.route);
+        unsigned long route = interrupt->phys.route & MPIDR_AFF_MSK;
+        for (cpuid_t i = 0; i < platform.cpu_num; i++) {
```
josecm: Do we want the vgic doing the reverse lookup at all? `phys.route` is only ever derived from a `phys_id` in `vgic_int_set_route`, so we could store the `phys_id` itself and only turn it into an MPIDR in `vgic_int_set_route_hw`. This would then be just `1UL << interrupt->phys.route`, the `routed_here` compares in `vgicv3.h` and `vgic_int_has_other_target` become a plain compare against `phys_id` instead of reading MPIDR_EL1, and `route` and `redist` end up holding the same kind of value. The one wrinkle is the invalid route, which would map to `INVALID_CPUID` here and back to `GICD_IROUTER_INV` when writing the hardware.

If you'd rather keep the MPIDR in `phys.route`, I'd at least move this loop to a `cpu_mpidr_to_id()` in `cpu.c` next to `cpu_id_to_mpidr`. The vgic shouldn't need to know how the platform lays out MPIDRs, and we'll want that helper elsewhere anyway.

src/arch/armv8/vgicv3.c
```diff
@@ -30,12 +30,18 @@ bool vgic_int_has_other_target(struct vcpu* vcpu, struct vgic_int* interrupt)
...
+            if ((cpu_id_to_mpidr(i) & MPIDR_AFF_MSK) == route) {
+                return (cpumap_t)(1UL << i);
```
josecm: nit: we try to keep to a single return per function where feasible. If the loop stays, can you use a result local and `break` instead?

src/arch/armv8/vgicv3.c
josecm: This runs on the routing path, so I'd rather not do a table walk here at all. `vgic_int_set_route` already has the target `phys_id` in hand when it computes `phys.route`, so can we cache it there and make this a plain `1UL << interrupt->phys.cpu`? `phys.route` stays the MPIDR we write to GICD_IROUTER, we just keep the cpu id next to it (`INVALID_CPUID` for the invalid route). `phys.redist` already holds the same kind of value for private interrupts, so I think it can be that same field, turning the union into a struct with `route` and `cpu`.

josecm (changes requested):

src/arch/armv8/vgicv3.c
nks0005: Thanks — updated in f277bad.

`phys` is now a struct with `route` (MPIDR for GICD_IROUTER) and `cpu` (phys_id; INVALID_CPUID when the route is invalid).
As you suggested, `vgic_int_set_route()` stores both the MPIDR and the phys_id index, and `vgic_int_ptarget_mask()` just reads the cached index (`1UL << interrupt->phys.cpu`) with no table walk.

`phys.cpu` also replaces `phys.redist` for private interrupts — same phys_id value, just one field for both cases.

src/arch/armv8/inc/arch/vgic.h
```diff
@@ -25,9 +25,9 @@ struct vgic_int {
     struct vcpu* owner;
 #if (GIC_VERSION != GICV2)
     unsigned long route;
-    union {
-        vcpuid_t redist;
+    struct {
         unsigned long route;
+        cpuid_t cpu;
```
josecm: To minimize changes I'd further propose this: 


```suggestion
    struct {
        union {
            cpuid_t redist;
            cpuid_t cpu;
        }
        unsigned long route;
```

nks0005: Updated in a530c22. 

Agreed keeping phys.redist via a union and leaving the private IRQ call sites alone is cleaner. SPI path still uses phys.cpu for the cached phys_id.

josecm (commented):
Looks good to me now, thanks for going through the two rounds. Two remaining things, one in the SGI path that this PR doesn't touch, and a history request.

**Same shift in `vgic_icc_sgir_handler`**

The IRM branch there still does `cpu()->vcpu->vm->cpus & ~(1U << cpu()->vcpu->phys_id)`, which is the same narrow shift this PR fixes in `vgic_int_ptarget_mask` (a broadcast SGI from a pCPU at 32 or above is undefined). Since we're here, can we make it `1UL` too? The `vgicv2.c` twin doesn't matter, GICv2 caps at 8 cpus.

**History**

Before merging, can you squash the three commits into one? The second and third are review fixups and don't stand on their own.

I've rebased the branch on main, so please pull it before making these changes.

PR #405, opened 2026-09-14
fix(tc4): move hypercall argument to correct register

miguelafsilva5:
## PR Description

The hypercall instruction, `hvcall`, allows for a 8-bit argument that is received in the handler through the d15 register. Moving this value to d4 is necessary to follow the ABI rules and for it to be used in the following C functions.


src/arch/tricore/exceptions.S
```diff
@@ -203,6 +203,7 @@ hyp_sys_bus_errors_handler:
 .global hyp_call_handler
 hyp_call_handler:
     VM_EXIT
+    mov %d4, %d15
```
DavidMCerdeira: This will use the hvcall id encoded in the hvcall instruction. For alignment with other architecture implementations we should use the id received from function-like arguments.

miguelafsilva5: Fixed in the new commit. Changed to behave similar to other architectures, where the id is passed in the 1st register argument.

DavidMCerdeira (approved):

danielRep (approved):

PR #408, opened 2026-09-14
fix(armv8): use MPIDR affinity mask

DavidMCerdeira:
PSCI expects the affinity fields of the target cpu's MPIDR in the CPU_ON target argument, with the register's other bits clear. Bao passed the whole MPIDR, including the RES1 bit. Mask the target with MPIDR_AFF_MSK, the affinity levels the port supports.


josecm:
@DavidMCerdeira did you check if any other PSCI call argument needs this bit cleared?

DavidMCerdeira:
> @DavidMCerdeira did you check if any other PSCI call argument needs this bit cleared?

No, only this. I'll check for other instances.

josecm:
> No, only this. I'll check for other instances.

I'll aprove anyway. Please force push if you find other instances which will render the approval obsolete.

josecm (approved):

joaopeixoto13 (approved):

Issue #347, opened 2026-04-06
feat(platform): support per-CPU local/tightly-coupled memory for cpu and vcpu struct placement


josecm:
## Summary

On MCU-class targets each CPU typically has its own tightly-coupled or local memory (TCM, DSPR, local RAM) with deterministic, zero-wait-state access latency. An important optimization for real-time workloads would be to place each CPU's `struct cpu` and the `struct vcpu`s it runs in that CPU's local memory, eliminating cache variability from the hot path of every context switch, interrupt, and scheduler invocation.

## Motivation

ARM explicitly recommends placing "call stack, RTOS control structures, interrupt data structures, and interrupt handler code" in TCM for deterministic latency [[ARM TCM documentation](https://docs.kernel.org/arch/arm/tcm.html)]. The Cortex-R52 TCM offers zero-wait-state access [[Cortex-R52 TRM](https://developer.arm.com/documentation/100026/latest/)], and Infineon describes DSPR as "high-speed, single-cycle SRAM core-private memory intended for stacks, local variables, and non-shared data to achieve maximum performance with deterministic access timing". Placing `struct cpu` and the associated `struct vcpu` in the local memory of the CPU that owns them would:

- Eliminate cache miss variability on every `cpu()` dereference and vcpu context switch
- Improve VM exit and entry latencies, since guest state is saved to and restored from these low-latency memories on every world switch
- Improve worst-case interrupt latency bounds — critical for mixed-criticality systems
- Reduce cross-core memory traffic on shared buses

## Current state

Several platforms already describe their per-CPU local memories in the platform descriptor as generic `mem_region` entries, but **no mechanism exists to direct allocations there**. `struct cpu` is always allocated linearly from `_dmem_beg` in global memory, regardless of available local memory.

- **TriCore TC4DX** — DSPR per core (240 KB each) described in `src/platform/tc4dx/tc4dx_desc.c`, but `_dmem_beg` follows `_image_end` in global LMU memory; DSPR is unused by the hypervisor
- **RH850 U2A16** — Local RAM per core (64 KB each) described in `src/platform/rh850-u2a16/u2a16_desc.c`
- **NXP S32Z270** — ATCM/BTCM/CTCM per core described in `src/platform/s32z270/s32z270_desc.c` [[S32Z2 datasheet](https://www.nxp.com/docs/en/data-sheet/S32Z27_DS.pdf)]

[...]

PR #410, opened 2026-09-14
feat/percpu block

DavidMCerdeira:
This PR lays the groundwork for optimizations on MCU targets: the hypervisor's per-cpu state becomes one linker-defined block per cpu that the platform can place in cpu-coupled memory, and initialization stops allocating its vm and vcpu structures from the page allocator.

1. **Per-cpu private block.** A new `.percpu` linker section is the template of the block:    `struct cpu` (with its stack) first, then every object declared with `DEFINE_PERCPU`. Each    cpu owns one instance placed by the boot code; objects are reached with `percpu_get(name, cpu)`    through an address valid on every cpu. On MPU targets that is the block's physical address; on    MMU targets `percpu_init()` maps all blocks in an alias window at the top of the private    section. The boot code of every port sizes and places the block from the linked size    (`_percpu_size`) instead of `sizeof(struct cpu)`.
2. **Static vm and vcpu storage.** `vms[CONFIG_VM_NUM]` replaces the page-allocated vm block;    each cpu holds a pool of vcpus in its private block (`struct vcpu_pool`: counter plus `vcpus[CONFIG_VCPU_PER_CPU_NUM]`, 1 for now) handed out by `vcpu_alloc()`, so vcpu scheduling can later grow the pool without changing the layout. `vm_mem_prot_cpu_init()` replaces the vm install step: MMU shares the master's vm-section page-table entry, MPU maps the sibling cpus' blocks.
3. **Cpu-affine memory regions.** `struct mem_region.cpu_affinity` (`CORE(n)`) marks memory coupled to a cpu. Affine regions are never page pools, vm memory placed in them is accepted after a check against the private blocks, and a region affine to exactly one cpu hosts that cpu's block: the platform defines generator emits the per-cpu base table, the linker asserts the block fits, `percpu_base()` resolves through the table. Ports opt in with `arch_cpu_private_placement()` once their boot code reads the table.
4. **rh850 and tricore placement.** rh850-u2a16 couples its four local RAMs, tc4dx its six DSPRs, so `struct cpu`, the stack and the vcpus live in the core's local memory; tricore additionally moves its 32 context save areas per cpu into the block.

Trade-offs: the vm structures are no longer isolated per vm in the hypervisor address space; a cpu private block is reachable from the other cpus of the vm (needed for cross-vcpu access, e.g. vGIC routing and PSCI). Other architectures keep the image-adjacent layout until their boot code reads the table; the generator rejects affine regions there.

Marking this as a draft to discuss implementation, tradeoffs or other ideas.


src/core/inc/percpu.h
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+#ifndef __PERCPU_H__
+#define __PERCPU_H__
+
+#include <bao.h>
+#include <cpu.h>
+
+/**
+ * Per-cpu objects live in the per-cpu private block described by the .percpu linker section: the
```
josecm: I don't think we need a second mechanism to access per-cpu data, we already have one: `struct cpu` through `cpu()`, one register read on MPU and the constant `BAO_CPU_BASE` on MMU. Everything here (the vcpus, the tricore CSAs) can just be a member of it, arch stuff in `cpu_arch`, and the compiler computes the offsets. With that the whole section, `_percpu_size` and the boot code changes go away, and the boot code keeps using `CPU_SIZE`.

This is what I had in mind in #347: a per-cpu local memory that the boot code allocates `struct cpu` (and the vcpus pinned to that cpu) from, without changing how per-cpu data is accessed.

src/core/inc/percpu.h
```diff
@@ -0,0 +1,51 @@
...
+#define DEFINE_PERCPU(type, name) type name __attribute__((section(".percpu")))
+
+#define PERCPU_OFFSET(name)       ((uintptr_t) & (name) - (uintptr_t) & _percpu_start)
+
+/**
+ * The instance of a per-cpu object belonging to the given cpu, through an address that is valid
+ * on every cpu (see percpu_base()), so the resulting pointer may be shared with other cpus. The
+ * calling cpu's own instance is percpu_get(name, cpu()->id).
+ */
+#define percpu_get(name, cpuid)   ((__typeof__(name)*)(percpu_base(cpuid) + PERCPU_OFFSET(name)))
```
josecm: Compared to `cpu()->field` (register plus immediate offset) this is a load of `id`, a call to `percpu_base`, a multiply, two absolute address loads for the offset and an add, and on MMU it lands on a different virtual address for the same memory. For a change meant to speed things up on MCUs, the access path got slower on every architecture.

Also, macros are upper case, and function-like macros should be static inline functions.

src/core/mmu/percpu.c
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Every cpu sees its own private block at BAO_CPU_BASE, so that address cannot name another cpu's
+ * block. Instead each cpu maps the blocks of all cpus, its own included, in an alias window at the
+ * top of its private section. The alias of a given cpu is the same address on every cpu, which
+ * makes it the canonical pointer for per-cpu objects shared with other cpus.
+ */
+
+static paddr_t percpu_pa[PLAT_CPU_NUM];
+
+vaddr_t percpu_base(cpuid_t cpuid)
```
josecm: On MMU targets we already access the cpu struct directly at `BAO_CPU_BASE` through aliasing, which is as fast as it gets. Why do we need this window at all? The only thing it buys is reaching another cpu's block, and I don't think we should be doing that (see the comment on the sibling mappings in the mpu vm code).

src/core/cpu.c
```diff
@@ -8,6 +8,20 @@
...
+ * instance (see percpu.h): the boot code provides each cpu its own copy, reachable through cpu().
+ */
+struct cpu cpu_private_block __attribute__((section(".percpu.cpu"), used));
+
+/**
+ * Where the boot code places each cpu's private block when the platform couples a memory region
+ * to that cpu (see struct mem_region.cpu_affinity); 0 means the image-adjacent fallback layout.
+ * Read by the boot code, hence a plain constant table.
+ */
+const paddr_t cpu_private_base_tbl[PLAT_CPU_NUM] = PLAT_CPU_PRIVATE_BASES;
```
josecm: Agree with generating this table, but it should only be read by the boot code, which then stores the pointer in the `cpu()` register. C should never need it. And I'd put the whole thing behind a build flag, allowed only on MPU architectures since that is where these MCU-like cores with local memories are, keeping the current layout as the default.

src/core/vm.c
```diff
@@ -11,6 +11,37 @@
...
 #include <list.h>
+#include <percpu.h>
+
+struct vcpu_pool {
+    /* next free slot, only ever advanced by the owning cpu */
+    size_t count;
+    struct vcpu vcpus[CONFIG_VCPU_PER_CPU_NUM];
+};
+
+static DEFINE_PERCPU(struct vcpu_pool, vcpu_pool);
```
josecm: This is my other main issue. `CONFIG_VCPU_PER_CPU_NUM` is always 1 here, and once cpu-sharing lands (soon) it cannot be computed at build time at all, the assignment of vcpus to cpus may or may not be static. Sizing this for the maximum possible vcpus per cpu wastes memory we do not have in a TCM.

What I'd do instead: each cpu allocates an array of vcpus from its private memory right after its cpu struct, sized from the config at init, and hands vcpus out from there. More generally, an allocator over the cpu's private memory that can allocate contiguously with the previous block, so the cpu, its vcpus and (for single-vcpu vms) the vm struct end up as one MPU entry through the vmpu merging. Any other cpu, vcpu or vm-specific allocation could go through the same path.

scripts/config_defs_gen.c
```diff
@@ -21,6 +21,16 @@ static size_t remio_dev_num(void)
     return dev_num;
 }
 
+/**
+ * Number of vcpu structures each physical cpu holds in its private block. Without vcpu scheduling
+ * a physical cpu runs at most one vcpu; once the configuration can pin several vcpus to a cpu this
+ * becomes the maximum over all cpus.
+ */
+static size_t vcpu_per_cpu_num(void)
```
josecm: With cpu-sharing this is not computable at build time, see my comment on `vcpu_pool`.

src/core/mpu/vm.c
```diff
@@ -4,6 +4,30 @@
  */
 
 #include <vm.h>
+#include <percpu.h>
+#include <platform.h>
+
+void vm_mem_prot_cpu_init(struct vm* vm)
```
josecm: A vcpu should only be mapped on the cpu hosting it. This costs one locked entry per sibling, and with the blocks scattered across the local RAMs they can't be merged, so a 4 cpu vm on the u2a16 spends 3 entries per cpu just to reach the other vcpus. Keeping the MPU entry count down is a requirement for this feature. Whatever other cpus need from a vcpu (`phys_id`, `started` on rh850, the vgic state on Arm) should live in the vm struct instead.

src/core/vmm.c
```diff
@@ -8,19 +8,17 @@
 #include <config.h>
 #include <cpu.h>
 #include <spinlock.h>
-#include <fences.h>
-#include <string.h>
 #include <shmem.h>
 
+/* The vm structures are statically allocated, one per vm in the configuration */
+static struct vm vms[CONFIG_VM_NUM];
```
josecm: An objpool of vms is fine, the size is known at build time. What I'd like to keep possible is putting the vm struct itself in the cpu's private memory for single-vcpu vms, which a global array precludes. With a per-cpu private allocator that falls out naturally.

src/core/mem.c
```diff
@@ -445,7 +497,8 @@ static void mem_reserve_physical_memory(struct page_pool* pool)
 static bool mem_create_ppools(struct mem_region* root_mem_region)
 {
     for (size_t i = 0; i < platform.region_num; i++) {
-        if (&platform.regions[i] != root_mem_region) {
+        /* Cpu-affine regions (TCM, local RAM) are never page pools */
```
josecm: I'd go the other way: make the local memory the cpu's private page pool, with the cpu struct at its base, and let each cpu reserve and allocate its own stuff there. As it is, nothing can be allocated in the local RAM at runtime and everything past the block is dead space unless a vm config places memory there by hand.

scripts/platform_defs_gen.c
```diff
@@ -36,13 +45,49 @@ int main() {
...
+
         reg_size = platform.regions[i].size;
 
         bitmap_array_size += BITMAP_SIZE_IN_BYTES(NUM_PAGES(reg_size));
     }
 
     printf("#define PLAT_BITMAP_POOL_SIZE (0x%lx)\n", bitmap_array_size);
 
+    /*
+     * Cpu private block placement: the first region affine to exactly one cpu hosts that cpu's
```
josecm: Placement being a side effect of `cpu_affinity` is too implicit for me. I'd rather have the platform (or the config) say explicitly where each cpu's private block goes, which also lets someone pick a specific TCM when there are several, or an offset inside it.

src/arch/rh850/boot.S
```diff
@@ -107,12 +107,22 @@ _reset_handler:
...
+    LOAD_ADDR cpu_private_base_tbl, r20
     mov r10, r21
-    mulh r12, r21   /* cpu struct offset */
-    add r21, r20    /* end of .bss + cpu struct offset = r20 points to cpu */
+    shl 2, r21          /* table index * sizeof(paddr_t) */
+    add r21, r20
+    ld.w 0[r20], r20
+    cmp r0, r20
+    bne _cpu_block_found
+    /* fallback: blocks laid out back to back after the image */
```
josecm: For the cpus that fall back to the image-adjacent layout, instead of `cpuid * size` we could take the slot from a lock-protected counter at boot, so only the cpus without a local memory consume slots, and mem init reads the same counter to know how much to mark as allocated.

src/arch/tricore/exceptions.S
```diff
@@ -250,9 +250,12 @@ vcpu_arch_entry:
     mfcr %d0, $pcxi
     mfcr %d1, $fcx
     mtcr $fcx, %d0
-    load_32bit_val_d %d0, csa_array
-    mov %d2, CSA_ARRAY_SIZE
-    madd %d0, %d0, %d3, %d2
+    /* this cpu's CSA pool: cpu private block (a8) + offset of the template */
```
josecm: This went from 4 to 7 instructions on the guest entry path. With the CSAs as a `cpu_arch` member this is `a8` plus a constant offset from asm_defs, 2 instructions.

src/arch/tricore/csa.c
```diff
@@ -4,6 +4,8 @@
  */
 
 #include <platform.h>
+#include <percpu.h>
 #include <arch/csa.h>
 
-union csa csa_array[PLAT_CPU_NUM][CSA_ENTRIES] __attribute__((aligned(64)));
+/* Each cpu's context save areas live in its private block (DSPR when the platform couples it) */
+DEFINE_PERCPU(union csa, csa_pool[CSA_ENTRIES]);
```
josecm: This belongs in `struct cpu_arch`.

src/linker.ld
```diff
@@ -139,7 +139,28 @@ SECTIONS
...
 	_cpu_private_beg = ABSOLUTE(.);
+#endif
 
+	/**
+	 * Per-cpu private block template: struct cpu (always first) followed by every object declared
+	 * with DEFINE_PERCPU. Each cpu owns one instance of the whole block. On MPU targets the
+	 * instances are laid out back to back starting at _image_end (or in a cpu-affine platform
+	 * region), on MMU targets every cpu maps its own instance at BAO_CPU_BASE.
+	 */
+	.percpu (NOLOAD) : ALIGN(PAGE_SIZE) {
```
josecm: One more problem with the template: on MPU with every cpu in its local RAM none of the fallback slots get reserved, so the link-time addresses of `cpu_private_block`, `vcpu_pool` and `csa_pool` point into live page pool memory. Any direct reference silently corrupts whatever got allocated there. A struct member has no address to misuse.

src/core/inc/vm.h
```diff
@@ -74,7 +75,12 @@ struct vm {
     struct cpu_synctoken sync;
     cpuid_t master;
 
-    struct vcpu* vcpus;
+    /**
+     * The vm's vcpus, indexed by vcpu id. Each vcpu lives in the private block of the physical
+     * cpu running it, the pointer is the one valid on every cpu (see vcpu_alloc()). One vcpu per
+     * physical cpu for now.
+     */
+    struct vcpu* vcpus[PLAT_CPU_NUM];
```
josecm: Keep this as an array of pointers, we want random access to the i-th vcpu. But `PLAT_CPU_NUM` per vm won't hold once several vcpus of a vm can share a cpu; the total is known at build time (`CONFIG_VCPU_NUM`).

src/core/inc/percpu.h
josecm: I don't think we need a second mechanism to access per-cpu data, we already have one: `struct cpu` through `cpu()`, one register read on MPU and the constant `BAO_CPU_BASE` on MMU. Everything here (the vcpus, the tricore CSAs) can just be a member of it, arch stuff in `cpu_arch`, and the compiler computes the offsets. With that the whole section, `_percpu_size` and the boot code changes go away, and the boot code keeps using `CPU_SIZE`.

This is what I had in mind in #347: a per-cpu local memory that the boot code allocates `struct cpu` (and the vcpus pinned to that cpu) from, without changing how per-cpu data is accessed.

src/core/inc/percpu.h
josecm: Compared to `cpu()->field` (register plus immediate offset) this is a load of `id`, a call to `percpu_base`, a multiply, two absolute address loads for the offset and an add, and on MMU it lands on a different virtual address for the same memory. For a change meant to speed things up on MCUs, the access path got slower on every architecture.

Also, macros are upper case, and function-like macros should be static inline functions.

src/core/mmu/percpu.c
josecm: On MMU targets we already access the cpu struct directly at `BAO_CPU_BASE` through aliasing, which is as fast as it gets. Why do we need this window at all? The only thing it buys is reaching another cpu's block, and I don't think we should be doing that (see the comment on the sibling mappings in the mpu vm code).

src/core/cpu.c
josecm: Agree with generating this table, but it should only be read by the boot code, which then stores the pointer in the `cpu()` register. C should never need it. And I'd put the whole thing behind a build flag, allowed only on MPU architectures since that is where these MCU-like cores with local memories are, keeping the current layout as the default.

Also, the address for each cpu should be overridable in the config. By default it would be the first private region of that cpu in the platform description, but the user might want a different TCM when a cpu has several, or an offset other than 0, for example to give the first address of the TCM to a vm and minimize the changes to the guest.

src/core/vm.c
josecm: This is my other main issue. `CONFIG_VCPU_PER_CPU_NUM` is always 1 here, and once cpu-sharing lands (soon) it cannot be computed at build time at all, the assignment of vcpus to cpus may or may not be static. Sizing this for the maximum possible vcpus per cpu wastes memory we do not have in a TCM.

What I'd do instead: each cpu allocates every vcpu it hosts from its private memory, one at a time as the vcpu is assigned to it, through something like a `mem_alloc_private()` that only hands out memory from the cpu's TCM, with a flag asking it to allocate contiguously with a previous allocation (the cpu struct here), so the cpu, its vcpus and (for single-vcpu vms) the vm struct end up as one MPU entry through the vmpu merging. Any other cpu, vcpu or vm-specific allocation could go through the same path.

scripts/config_defs_gen.c
josecm: With cpu-sharing this is not computable at build time, see my comment on `vcpu_pool`.

src/core/mpu/vm.c
josecm: A vcpu should only be mapped on the cpu hosting it. This costs one locked entry per sibling, and with the blocks scattered across the local RAMs they can't be merged, so a 4 cpu vm on the u2a16 spends 3 entries per cpu just to reach the other vcpus. Keeping the MPU entry count down is a requirement for this feature. Whatever other cpus need from a vcpu (`phys_id`, `started` on rh850, the vgic state on Arm) should live in the vm struct instead.

src/core/vmm.c
josecm: An objpool of vms is fine, the size is known at build time. What I'd like to keep possible is putting the vm struct itself in the cpu's private memory for single-vcpu vms, which a global array precludes. With a per-cpu private allocator that falls out naturally.

src/core/mem.c
josecm: I'd go the other way: keep the local memory allocatable, through a `mem_alloc_private()` that only hands out memory from the calling cpu's TCM. As it is, nothing can be allocated in the local RAM at runtime and everything past the block is dead space unless a vm config places memory there by hand.

scripts/platform_defs_gen.c
josecm: Placement being a side effect of `cpu_affinity` is too implicit for me. I'd rather have this say explicitly where each cpu's private block goes, with the config able to override it (see my comment on the table in cpu.c).

src/arch/rh850/boot.S
josecm: Instead of `cpuid * size`, we could take the slot from a lock-protected counter at boot. That lets each cpu opt in independently, even on MPU systems: some cpus keep their struct next to the image while others use their TCM, and mem init reads the same counter to know how much to mark as allocated.

src/arch/tricore/exceptions.S
josecm: This went from 4 to 7 instructions on the guest entry path. With the CSAs as a `cpu_arch` member this is `a8` plus a constant offset from asm_defs, 2 instructions.

src/arch/tricore/csa.c
josecm: This belongs in `struct cpu_arch`.

src/linker.ld
josecm: One more problem with the template: on MPU with every cpu in its local RAM none of the fallback slots get reserved, so the link-time addresses of `cpu_private_block`, `vcpu_pool` and `csa_pool` point into live page pool memory. Any direct reference silently corrupts whatever got allocated there. A struct member has no address to misuse.

src/core/inc/vm.h
josecm: Agree with this one, an array of pointers keeps random access to the i-th vcpu. But `PLAT_CPU_NUM` per vm won't hold once several vcpus of a vm can share a cpu; the total is known at build time (`CONFIG_VCPU_NUM`).

src/core/inc/percpu.h
josecm: I don't think we need a second mechanism to access per-cpu data, we already have one: `struct cpu` through `cpu()`, one register read on MPU and the constant `BAO_CPU_BASE` on MMU. Everything here (the vcpus, the tricore CSAs) can just be a member of it, arch stuff in `cpu_arch`, and the compiler computes the offsets. With that the whole section, `_percpu_size` and the boot code changes go away, and the boot code keeps using `CPU_SIZE`.

This is what I had in mind in #347: a per-cpu local memory that the boot code allocates `struct cpu` (and the vcpus pinned to that cpu) from, without changing how per-cpu data is accessed.

src/core/inc/percpu.h
josecm: Compared to `cpu()->field` (register plus immediate offset) this is a load of `id`, a call to `percpu_base`, a multiply, two absolute address loads for the offset and an add, and on MMU it lands on a different virtual address for the same memory. For a change meant to speed things up on MCUs, the access path got slower on every architecture.

Also, macros are upper case, and function-like macros should be static inline functions.

src/core/mmu/percpu.c
josecm: On MMU targets we already access the cpu struct directly at `BAO_CPU_BASE` through aliasing, which is as fast as it gets. Why do we need this window at all? The only thing it buys is reaching another cpu's block, and I don't think we should be doing that (see the comment on the sibling mappings in the mpu vm code). Ideally the MMU flow stays as close as possible to what we have today; none of this buys anything there, since there are no local memories on those platforms.

src/core/cpu.c
josecm: Agree with generating this table, but it should only be read by the boot code, which then stores the pointer in the `cpu()` register. C should never need it. And I'd put the whole thing behind a build flag, allowed only on MPU architectures since that is where these MCU-like cores with local memories are, keeping the current layout as the default.

Also, the address for each cpu should be overridable in the config. By default it would be the first private region of that cpu in the platform description, but the user might want a different TCM when a cpu has several, or an offset other than 0, for example to give the first address of the TCM to a vm and minimize the changes to the guest.

src/core/vm.c
josecm: This is my other main issue. `CONFIG_VCPU_PER_CPU_NUM` is always 1 here, and once cpu-sharing lands (soon) it cannot be computed at build time at all, the assignment of vcpus to cpus may or may not be static. Sizing this for the maximum possible vcpus per cpu wastes memory we do not have in a TCM.

What I'd do instead: each cpu allocates every vcpu it hosts from its private memory, one at a time as the vcpu is assigned to it, through something like a `mem_alloc_private()` that only hands out memory from the cpu's TCM, with a flag asking it to allocate contiguously with a previous allocation (the cpu struct here), so the cpu, its vcpus and (for single-vcpu vms) the vm struct end up as one MPU entry through the vmpu merging. Any other cpu, vcpu or vm-specific allocation could go through the same path.

scripts/config_defs_gen.c
josecm: With cpu-sharing this is not computable at build time, see my comment on `vcpu_pool`.

src/core/mpu/vm.c
josecm: A vcpu should only be mapped on the cpu hosting it. This costs one locked entry per sibling, and with the blocks scattered across the local RAMs they can't be merged, so a 4 cpu vm on the u2a16 spends 3 entries per cpu just to reach the other vcpus. Keeping the MPU entry count down is a requirement for this feature, which is why each vcpu should be allocated contiguously with the previous ones and with the cpu struct, so all of it is a single entry on the hosting cpu. Whatever other cpus need from a vcpu (`phys_id`, `started` on rh850, the vgic state on Arm) should live in the vm struct instead.

src/core/vmm.c
josecm: An objpool of vms is fine, the size is known at build time. What I'd like to keep possible is putting the vm struct itself in the cpu's private memory for single-vcpu vms, which a global array precludes. With a per-cpu private allocator that falls out naturally.

src/core/mem.c
josecm: I'd go the other way: affine regions should be page pools like any other. The difference is that they don't take part in normal allocation (maybe based on a region type), or only as the last resort when no global region has enough memory, and they are used exclusively by `mem_alloc_private()` (the placeholder name from my comment on `vcpu_pool`) for the calling cpu's own TCMs. As it is, the blocks placed in a TCM are not tracked anywhere (only overlap checked against placed vm memory), nothing can be allocated in the local RAM at runtime, and everything past the block is dead space unless a vm config places memory there by hand.

scripts/platform_defs_gen.c
josecm: Placement being a side effect of `cpu_affinity` is too implicit for me. I'd rather have this say explicitly where each cpu's private block goes, with the config able to override it (see my comment on the table in cpu.c).

src/arch/rh850/boot.S
josecm: Instead of `cpuid * size`, we could take the slot from a lock-protected counter at boot. That lets each cpu opt in independently, even on MPU systems: some cpus keep their struct next to the image while others use their TCM, and mem init reads the same counter to know how much to mark as allocated.

src/arch/tricore/exceptions.S
josecm: This went from 4 to 7 instructions on the guest entry path. With the CSAs as a `cpu_arch` member this is `a8` plus a constant offset from asm_defs, 2 instructions.

src/arch/tricore/csa.c
josecm: This belongs in `struct cpu_arch`.

src/linker.ld
josecm: One more problem with the template: on MPU with every cpu in its local RAM none of the fallback slots get reserved, so the link-time addresses of `cpu_private_block`, `vcpu_pool` and `csa_pool` point into live page pool memory. Any direct reference silently corrupts whatever got allocated there. A struct member has no address to misuse.

src/core/inc/vm.h
josecm: Agree with this one, an array of pointers keeps random access to the i-th vcpu. But `PLAT_CPU_NUM` per vm won't hold once several vcpus of a vm can share a cpu; the total is known at build time (`CONFIG_VCPU_NUM`).

src/core/inc/percpu.h
josecm: I don't think we need a second mechanism to access per-cpu data, we already have one: `struct cpu` through `cpu()`, one register read on MPU and the constant `BAO_CPU_BASE` on MMU. Everything here (the vcpus, the tricore CSAs) can just be a member of it, arch stuff in `cpu_arch`, and the compiler computes the offsets. With that the whole section, `_percpu_size` and the boot code changes go away, and the boot code keeps using `CPU_SIZE`.

This is what I had in mind in #347: a per-cpu local memory that the boot code allocates `struct cpu` (and the vcpus pinned to that cpu) from, without changing how per-cpu data is accessed.

src/core/inc/percpu.h
josecm: Compared to `cpu()->field` (register plus immediate offset) this is a load of `id`, a call to `percpu_base`, a multiply, two absolute address loads for the offset and an add, and on MMU it lands on a different virtual address for the same memory. For a change meant to speed things up on MCUs, the access path got slower on every architecture.

Also, macros are upper case, and function-like macros should be static inline functions.

src/core/mmu/percpu.c
josecm: On MMU targets we already access the cpu struct directly at `BAO_CPU_BASE` through aliasing, which is as fast as it gets. Why do we need this window at all? The only thing it buys is reaching another cpu's block, and I don't think we should be doing that (see the comment on the sibling mappings in the mpu vm code). Ideally the MMU flow stays as close as possible to what we have today; none of this buys anything there, since there are no local memories on those platforms.

src/core/cpu.c
josecm: Agree with generating this table, but it should only be read by the boot code, which then stores the pointer in the `cpu()` register. C should never need it. And I'd put the whole thing behind a build flag, allowed only on MPU architectures since that is where these MCU-like cores with local memories are, keeping the current layout as the default.

Also, the address for each cpu should be overridable in the config. By default it would be the first private region of that cpu in the platform description, but the user might want a different TCM when a cpu has several, or an offset other than 0, for example to give the first address of the TCM to a vm and minimize the changes to the guest.

src/core/vm.c
josecm: This is my other main issue. `CONFIG_VCPU_PER_CPU_NUM` is always 1 here, and once cpu-sharing lands (soon) it cannot be computed at build time at all, the assignment of vcpus to cpus may or may not be static. Sizing this for the maximum possible vcpus per cpu wastes memory we do not have in a TCM.

What I'd do instead: each cpu allocates every vcpu it hosts from its private memory, one at a time as the vcpu is assigned to it, through something like a `mem_alloc_private()` that only hands out memory from the cpu's TCM, with a flag asking it to allocate contiguously with a previous allocation (the cpu struct here), so the cpu, its vcpus and (for single-vcpu vms) the vm struct end up as one MPU entry through the vmpu merging. Any other cpu, vcpu or vm-specific allocation could go through the same path.

scripts/config_defs_gen.c
josecm: With cpu-sharing this is not computable at build time, see my comment on `vcpu_pool`.

src/core/mpu/vm.c
josecm: A vcpu should only be mapped on the cpu hosting it. This costs one locked entry per sibling, and with the blocks scattered across the local RAMs they can't be merged, so a 4 cpu vm on the u2a16 spends 3 entries per cpu just to reach the other vcpus. Keeping the MPU entry count down is a requirement for this feature, which is why each vcpu should be allocated contiguously with the previous ones and with the cpu struct, so all of it is a single entry on the hosting cpu. Whatever other cpus need from a vcpu (`phys_id`, `started` on rh850, the vgic state on Arm) should live in the vm struct instead.

src/core/vmm.c
josecm: An objpool of vms is fine, the size is known at build time. What I'd like to keep possible is putting the vm struct itself in the cpu's private memory for single-vcpu vms, which a global array precludes. With a per-cpu private allocator that falls out naturally.

src/core/mem.c
josecm: I'd go the other way: affine regions should be page pools like any other. The difference is that they don't take part in normal allocation (maybe based on a region type), or only as the last resort when no global region has enough memory, and they are used exclusively by `mem_alloc_private()` (the placeholder name from my comment on `vcpu_pool`) for the calling cpu's own TCMs. As it is, the blocks placed in a TCM are not tracked anywhere (only overlap checked against placed vm memory), nothing can be allocated in the local RAM at runtime, and everything past the block is dead space unless a vm config places memory there by hand.

scripts/platform_defs_gen.c
josecm: Placement being a side effect of `cpu_affinity` is too implicit for me. I'd rather have this say explicitly where each cpu's private block goes, with the config able to override it (see my comment on the table in cpu.c).

src/arch/rh850/boot.S
josecm: Instead of `cpuid * size`, we could take the slot from a lock-protected counter at boot. That lets each cpu opt in independently, even on MPU systems: some cpus keep their struct next to the image while others use their TCM, and mem init reads the same counter to know how much to mark as allocated.

src/arch/tricore/exceptions.S
josecm: This went from 4 to 7 instructions on the guest entry path. With the CSAs as a `cpu_arch` member this is `a8` plus a constant offset from asm_defs, 2 instructions.

src/arch/tricore/csa.c
josecm: This belongs in `struct cpu_arch`.

src/linker.ld
josecm: One more problem with the template: on MPU with every cpu in its local RAM none of the fallback slots get reserved, so the link-time addresses of `cpu_private_block`, `vcpu_pool` and `csa_pool` point into live page pool memory. Any direct reference silently corrupts whatever got allocated there. A struct member has no address to misuse.

src/core/inc/vm.h
josecm: Agree with this one, an array of pointers keeps random access to the i-th vcpu. But `PLAT_CPU_NUM` per vm won't hold once several vcpus of a vm can share a cpu; the total is known at build time (`CONFIG_VCPU_NUM`).

src/core/inc/percpu.h
josecm: I don't think we need a second mechanism to access per-cpu data, we already have one: `struct cpu` through `cpu()`, one register read on MPU and the constant `BAO_CPU_BASE` on MMU. Everything here (the vcpus, the tricore CSAs) can just be a member of it, arch stuff in `cpu_arch`, and the compiler computes the offsets. With that the whole section, `_percpu_size` and the boot code changes go away, and the boot code keeps using `CPU_SIZE`.

This is what I had in mind in #347: a per-cpu local memory that the boot code allocates `struct cpu` (and the vcpus pinned to that cpu) from, without changing how per-cpu data is accessed.

src/core/inc/percpu.h
josecm: Compared to `cpu()->field` (register plus immediate offset) this is a load of `id`, a call to `percpu_base`, a multiply, two absolute address loads for the offset and an add, and on MMU it lands on a different virtual address for the same memory. For a change meant to speed things up on MCUs, the access path got slower on every architecture.

Also, macros are upper case, and function-like macros should be static inline functions.

src/core/mmu/percpu.c
josecm: On MMU targets we already access the cpu struct directly at `BAO_CPU_BASE` through aliasing, which is as fast as it gets. Why do we need this window at all? The only thing it buys is reaching another cpu's block, and I don't think we should be doing that (see the comment on the sibling mappings in the mpu vm code). Ideally the MMU flow stays as close as possible to what we have today; none of this buys anything there, since there are no local memories on those platforms.

src/core/cpu.c
josecm: Agree with generating this table, but it should only be read by the boot code, which then stores the pointer in the `cpu()` register. C should never need it. And I'd put the whole thing behind a build flag, allowed only on MPU architectures since that is where these MCU-like cores with local memories are, keeping the current layout as the default.

Also, the address for each cpu should be overridable in the config. By default it would be the first private region of that cpu in the platform description, but the user might want a different TCM when a cpu has several, or an offset other than 0, for example to give the first address of the TCM to a vm and minimize the changes to the guest.

src/core/vm.c
josecm: This is my other main issue. `CONFIG_VCPU_PER_CPU_NUM` is always 1 here, and once cpu-sharing lands (soon) it cannot be computed at build time at all, the assignment of vcpus to cpus may or may not be static. Sizing this for the maximum possible vcpus per cpu wastes memory we do not have in a TCM.

What I'd do instead: each cpu allocates every vcpu it hosts from its private memory, one at a time as the vcpu is assigned to it, through something like a `mem_alloc_private()` that only hands out memory from the cpu's TCM, with a flag asking it to allocate contiguously with a previous allocation (the cpu struct here), so the cpu, its vcpus and (for single-vcpu vms) the vm struct end up as one MPU entry through the vmpu merging. Any other cpu, vcpu or vm-specific allocation could go through the same path.

scripts/config_defs_gen.c
josecm: With cpu-sharing this is not computable at build time, see my comment on `vcpu_pool`.

src/core/mpu/vm.c
josecm: A vcpu should only be mapped on the cpu hosting it. Each sibling block is mapped here as a separate locked region, and `mem_map` never coalesces locked regions, so whatever the layout a 4 cpu vm on the u2a16 spends 3 extra entries per cpu just to reach the other vcpus. And it is not just the vcpus: this maps the whole block, so every cpu of the vm can also write the other cpus' `struct cpu` and stacks. Keeping the MPU entry count down is a requirement for this feature, which is why each vcpu should be allocated contiguously with the previous ones and with the cpu struct, so all of it is a single entry on the hosting cpu. Whatever other cpus need from a vcpu (`phys_id`, `started` on rh850, the vgic state on Arm) should live in the vm struct instead.

src/core/vmm.c
josecm: An objpool of vms is fine, the size is known at build time. What I'd like to keep possible is putting the vm struct itself in the cpu's private memory for single-vcpu vms, which a global array precludes. With a per-cpu private allocator that falls out naturally.

src/core/mem.c
josecm: I'd go the other way: affine regions should be page pools like any other. The difference is that they don't take part in normal allocation (maybe based on a region type), or only as the last resort when no global region has enough memory, and they are used exclusively by `mem_alloc_private()` (the placeholder name from my comment on `vcpu_pool`) for the calling cpu's own TCMs. As it is, the blocks placed in a TCM are not tracked anywhere (only overlap checked against placed vm memory), nothing can be allocated in the local RAM at runtime, and everything past the block is dead space unless a vm config places memory there by hand.

scripts/platform_defs_gen.c
josecm: Placement being a side effect of `cpu_affinity` is too implicit for me. I'd rather have this say explicitly where each cpu's private block goes, with the config able to override it (see my comment on the table in cpu.c).

src/arch/rh850/boot.S
josecm: Instead of `cpuid * size`, we could take the slot from a lock-protected counter at boot. That lets each cpu opt in independently, even on MPU systems: some cpus keep their struct next to the image while others use their TCM, and mem init reads the same counter to know how much to mark as allocated.

src/arch/tricore/exceptions.S
josecm: This went from 4 to 7 instructions on the guest entry path. With the CSAs as a `cpu_arch` member this is `a8` plus a constant offset from asm_defs, 2 instructions.

src/arch/tricore/csa.c
josecm: This belongs in `struct cpu_arch`.

src/linker.ld
josecm: One more problem with the template: on MPU with every cpu in its local RAM none of the fallback slots get reserved, so the link-time addresses of `cpu_private_block`, `vcpu_pool` and `csa_pool` point into live page pool memory. Any direct reference silently corrupts whatever got allocated there. A struct member has no address to misuse.

src/core/inc/vm.h
josecm: Agree with this one, an array of pointers keeps random access to the i-th vcpu. But `PLAT_CPU_NUM` per vm won't hold once several vcpus of a vm can share a cpu; the total is known at build time (`CONFIG_VCPU_NUM`).

src/core/inc/percpu.h
josecm: I don't think we need a second mechanism to access per-cpu data, we already have one: `struct cpu` through `cpu()`, one register read on MPU and the constant `BAO_CPU_BASE` on MMU. Everything here (the vcpus, the tricore CSAs) can just be a member of it, arch stuff in `cpu_arch`, and the compiler computes the offsets. With that the whole section, `_percpu_size` and the boot code changes go away, and the boot code keeps using `CPU_SIZE`.

This is what I had in mind in #347: a per-cpu local memory that the boot code allocates `struct cpu` (and the vcpus pinned to that cpu) from, without changing how per-cpu data is accessed.

src/core/inc/percpu.h
josecm: Compared to `cpu()->field` (register plus immediate offset) this is a load of `id`, a call to `percpu_base`, a multiply, two absolute address loads for the offset and an add, and on MMU it lands on a different virtual address for the same memory. For a change meant to speed things up on MCUs, the access path got slower on every architecture.

Also, macros are upper case, and function-like macros should be static inline functions.

src/core/mmu/percpu.c
josecm: On MMU targets we already access the cpu struct directly at `BAO_CPU_BASE` through aliasing, which is as fast as it gets. Why do we need this window at all? The only thing it buys is reaching another cpu's block, and I don't think we should be doing that (see the comment on the sibling mappings in the mpu vm code). Ideally the MMU flow stays as close as possible to what we have today; none of this buys anything there, since there are no local memories on those platforms.

src/core/cpu.c
josecm: Agree with generating this table, but it should only be read by the boot code, which then stores the pointer in the `cpu()` register. C should never need it. And I'd put the whole thing behind a build flag, allowed only on MPU architectures since that is where these MCU-like cores with local memories are, keeping the current layout as the default.

Also, the address for each cpu should be overridable in the config. By default it would be the first private region of that cpu in the platform description, but the user might want a different TCM when a cpu has several, or an offset other than 0, for example to give the first address of the TCM to a vm and minimize the changes to the guest.

src/core/vm.c
josecm: This is my other main issue. `CONFIG_VCPU_PER_CPU_NUM` is always 1 here, and once cpu-sharing lands (soon) it cannot be computed at build time at all, the assignment of vcpus to cpus may or may not be static. Sizing this for the maximum possible vcpus per cpu wastes memory we do not have in a TCM.

What I'd do instead: each cpu allocates every vcpu it hosts from its private memory, one at a time as the vcpu is assigned to it, through something like a `mem_alloc_private()` that only hands out memory from the cpu's TCM, with a flag asking it to allocate contiguously with a previous allocation (the cpu struct here), so the cpu, its vcpus and (for single-vcpu vms) the vm struct end up as one MPU entry through the vmpu merging. Any other cpu, vcpu or vm-specific allocation could go through the same path.

scripts/config_defs_gen.c
josecm: With cpu-sharing this is not computable at build time, see my comment on `vcpu_pool`.

src/core/mpu/vm.c
josecm: I get why this is needed, other vcpus do touch a vcpu (the redistributor state on Arm, psci and sbi hsm, `started` on rh850, `phys_id` everywhere). But each sibling block is mapped as a separate locked region, and `mem_map` never coalesces locked regions, so whatever the layout a 4 cpu vm on the u2a16 spends 3 extra entries per cpu. And it maps the whole block, so every cpu of the vm can also write the other cpus' `struct cpu` and stacks. Keeping the MPU entry count down is a requirement for this feature, so we need a different answer here. Maybe split the vcpu into a private context with what is context switched, in the host's TCM contiguous with its cpu struct, and a global part with the intra-vm management state, allocated contiguously per vm in global RAM and mapped as one region on all the vm's cpus? What do you think?

src/core/vmm.c
josecm, line 16: An objpool of vms is fine, the size is known at build time. What I'd like to keep possible is putting the vm struct itself in the cpu's private memory for single-vcpu vms, which a global array precludes. With a per-cpu private allocator that falls out naturally.

src/core/mem.c
josecm, line 508: I'd go the other way: affine regions should be page pools like any other. The difference is that they don't take part in normal allocation (maybe based on a region type), or only as the last resort when no global region has enough memory, and they are used exclusively by `mem_alloc_private()` (the placeholder name from my comment on `vcpu_pool`) for the calling cpu's own TCMs. As it is, the blocks placed in a TCM are not tracked anywhere (only overlap checked against placed vm memory), nothing can be allocated in the local RAM at runtime, and everything past the block is dead space unless a vm config places memory there by hand.

scripts/platform_defs_gen.c
josecm: Placement being a side effect of `cpu_affinity` is too implicit for me. I'd rather have this say explicitly where each cpu's private block goes, with the config able to override it (see my comment on the table in cpu.c).

src/arch/rh850/boot.S
josecm: Instead of `cpuid * size`, we could take the slot from a lock-protected counter at boot. That lets each cpu opt in independently, even on MPU systems: some cpus keep their struct next to the image while others use their TCM, and mem init reads the same counter to know how much to mark as allocated.

src/arch/tricore/exceptions.S
josecm: This went from 4 to 7 instructions on the guest entry path. With the CSAs as a `cpu_arch` member this is `a8` plus a constant offset from asm_defs, 2 instructions.

src/arch/tricore/csa.c
josecm: This belongs in `struct cpu_arch`.

src/linker.ld
josecm: One more problem with the template: on MPU with every cpu in its local RAM none of the fallback slots get reserved, so the link-time addresses of `cpu_private_block`, `vcpu_pool` and `csa_pool` point into live page pool memory. Any direct reference silently corrupts whatever got allocated there. A struct member has no address to misuse.

src/core/inc/vm.h
josecm: Agree with this one, an array of pointers keeps random access to the i-th vcpu. But `PLAT_CPU_NUM` per vm won't hold once several vcpus of a vm can share a cpu; the total is known at build time (`CONFIG_VCPU_NUM`).

josecm (changes requested):

josecm (commented):
Thanks for putting this together. Placing each cpu's private state in its local RAM / DSPR is something we need for the MCU targets, and marking the local memories in the platform description plus a generated table the boot code reads to find them is the right way to do it. Where I disagree is with most of what is built on top of that.

My main issue is that this adds a second, separate mechanism to access per-cpu data. We already have one: `struct cpu`, reached through `cpu()`, which is a single register read on MPU targets and the constant `BAO_CPU_BASE` on MMU targets. Anything that needs to be per-cpu can be a member of it (arch specifics in `cpu_arch`), the compiler computes the offsets, and the boot code keeps using `CPU_SIZE`. The `.percpu` section, `DEFINE_PERCPU`, `percpu_get`, `percpu_base` and the MMU alias window make the access path strictly slower than the one they replace on every architecture, which is a strange outcome for a change motivated by performance, and they bring the problems I point out inline (the table exported to C, the dangling template addresses on MPU, the extra instructions on the tricore entry path).

The other big point is the vcpu allocation. A fixed `vcpus[CONFIG_VCPU_PER_CPU_NUM]` per cpu does not scale to cpu-sharing, which we will land soon: the number of vcpus per cpu cannot be computed at build time, and sizing for the maximum wastes memory we do not have in a TCM. What I think we need instead, roughly:

1. Keep the generated table, read only by the boot code, which stores the block address in the `cpu()` register (MMU targets keep mapping it at `BAO_CPU_BASE`). Behind a build flag, allowed only on MPU architectures, with the current layout as the default.
2. Keep the image-adjacent layout, but take the slot at boot from a lock-protected counter instead of `cpuid * size`. That lets each cpu opt in independently, even on MPU systems: some cpus keep their struct next to the image while others use their TCM, and mem init reads the counter to know how much to mark as allocated.
3. Add something like a `mem_alloc_private()` that allocates only from the calling cpu's private memory (its TCM). Any cpu, vcpu or even vm-specific allocation can go through it, and on request (a flag) the algorithm strives to allocate contiguously with a previous allocation of that cpu, so the MPU entries get merged: the cpu struct, its vcpus and the vm struct for single-vcpu vms as a single entry.
4. Each cpu allocates every vcpu it hosts through that allocator, one at a time as the vcpu is assigned to it (with cpu-sharing we don't know upfront how many a cpu will hold), contiguous with its cpu struct and the previous vcpus. An alternative is for the cpu, once it knows exactly how many vcpus are assigned to it, to allocate an array of that many vcpus in its private memory in one go and then take each vcpu from that array as needed. `vm->vcpus[]` stays an array of pointers filled with their addresses, as done in this PR, which I agree with: we keep random access to the i-th vcpu.
5. Other vcpus of the vm do need to reach a vcpu (vgic redistributor state on Arm, psci and sbi hsm contexts, `started` on rh850, `phys_id` everywhere), so some shared mapping is unavoidable. The problem is the MPU pressure: here every sibling block is mapped as a separate locked region, locked regions are never merged, so each cpu of an N cpu vm spends N-1 extra entries (and gets write access to the other cpus' `struct cpu` and stacks along the way). Keeping the MPU entry count down is a requirement of this feature, so we need a solution for this. Off the top of my head: split the vcpu in two. A private context holding what is actually context switched, allocated in the host cpu's TCM contiguously with its cpu struct and the previous ones (so one entry on that cpu). And a global part with everything the hypervisor needs for intra-vm management, allocated contiguously per vm in global RAM and mapped as a single region on all the vm's cpus, as we do today.

On MMU targets I'd keep the flow as close as possible to what we have today.

Happy to go through any of this on a call if that is easier.

src/core/inc/percpu.h
DavidMCerdeira: >  For a change meant to speed things up on MCUs, the access path got slower on every architecture.

I don't appreciate the comment.

Just the fact that we are using TCM for the private structures is enough to offset the performance hit in the few instances where it occurs.

src/core/vm.c
DavidMCerdeira: How should we handle allocation when we disable the bitmaps in this case?
I'd like to avoid having two different paths based on that option if possible.

src/core/inc/percpu.h
josecm: You can have both. You don't need this percpu infrastrcuture. You cam simply use cpu()-> without going through all this indirection percpu has *and* have the TCM access.

src/core/vm.c
josecm: That would be ideal, but, is that even possible? If there aren't two different paths based on that option you will be allowed no allocation inside the kernel. Like, never. Aren't there any other places where such allocations exist?

DavidMCerdeira: > That would be ideal, but, is that even possible?

Since the hypervisor is built against a known configuration, at build time we know what VMs and VCPUs need to allocated. Why should the hypervisor decided in runtime what can be decided at compile time? 

> If there aren't two different paths based on that option you will be allowed no allocation inside the kernel.

The purpose of dynamic allocation in our case should be deal with allocations that are only knowable in runtime.

I do agree with your point regarding wasting TCM by allocating for the maximum number of vcpus in all cpus with schedulling. My solution to that is that vcpu to cpu allocation can be decided offline, and the necessary per-cpu data size can be computed accordingly. Do you see any advantage to do it in runtime?

josecm: > at build time we know what VMs and VCPUs need to allocated. 

> My solution to that is that vcpu to cpu allocation can be decided offline, and the necessary per-cpu data size can be computed accordingly. Do you see any advantage to do it in runtime?

This was my main point when I raised the cpu sharing issue. On this case, you cannot know how many vcpus for each cpu you will have. So that all cpu allocation are defined offline we must force the user to specify affinity. This is currently optional and I would prefer to keep it this way: Bao's config doesn't mandate the user defines static cpu affinities or memory allocations, it allows it, optionally. Removing that optionality would change the paradigm we've folowed for the config so far.

DavidMCerdeira: > So that all cpu allocation are defined offline we must force the user to specify affinity. 

The decision can be made offline/build time if the user does not specify affinity. I'm proposing removing this decision process from the Bao runtime into the build system. Again, what's the benefit of doing this in runtime?

josecm: OK, but that is another direction I wasn't taking into account. No precedent for that also at this stage. But worth debating. The benefit of doing this at runtime is exactly not having complex machinery at build time. The existing machinery is simply counting stuff in the config to generate some macros. 

Furthermore, I think this will create a rabbit whole on what we can do at build time or not. On that line of reasoning: why don't we do all allocations statically in the script all build time and fully remove that code from the codebase also? Why don't we precompute everything we need, page-tables, MPU entries, system register setups? Doing this would limit a lot of use cases I believe we both envision for bao, especially on application platforms.

josecm: BTW, I think we can get to a middle-ground without foregoing the little flexibility bao has. But this will require further debate and design which I don't think is in the scope of this PR.

DavidMCerdeira: > OK, but that is another direction I wasn't taking into account. No precedent for that also at this stage. But worth debating. The benefit of doing this at runtime is exactly not having complex machinery at build time. The existing machinery is simply counting stuff in the config to generate some macros. 

How do you justify having this "complex machinery" in runtime instead? My understanding is that bao should be as simple as possible.

> Furthermore, I think this will create a rabbit whole on what we can do at build time or not. On that line of reasoning: why don't we do all allocations statically in the script all build time and fully remove that code from the codebase also? Why don't we precompute everything we need, page-tables, MPU entries, system register setups? Doing this would limit a lot of use cases I believe we both envision for bao, especially on application platforms.

I don't like this black and white thinking, nor am I in favor of the slippery slope argument. One could make the argument in reverse, why do anything static at all? Just derive everything from the config.c in runtime. Obviously there's a middle ground. I think my proposal is reasonable and aligns with our current goals with MCU platforms. Also, with my proposed approach, MMU platforms will benefit from a simpler boot. 

I still see no arguments for allocating VCPUs and VMs in runtime other than "that's what we've always done".

josecm: > How do you justify having this "complex machinery" in runtime instead? My understanding is that bao should be as simple as possible.

I think we both understand that bao is as minimal as possible without limiting itself in simply becoming a commit engine for build-time decisions.

> why do anything static at all? Just derive everything from the config.c in runtime.

Agree. That's the  balance and trade-off which we should be trying to strike.

> I still see no arguments for allocating VCPUs and VMs in runtime other than "that's what we've always done".

I dont think the point is "that's what we've always done". I believe is more, that is how the design is, and what you are proposing is a significant change that might take the design significantly towards one end of the spectrum, and might create limitations further ahead.

DavidMCerdeira:
Here's a new draft following our discussion earlier in the day. Since this is now on top of exp/kconfg I took the oportunity to add additional optimizations related to the use of TCMs.

What changed in the design:

- Per-cpu block. The linker .percpu template, DEFINE_PERCPU, the percpu.h/percpu.c pair per memory protection and the size symbol are gone. struct cpu is the block, and it embeds a single struct vcpu as an object, no vcpu pool or allocator. Assembly reaches the vcpu registers through generated offsets.
- Public vcpu state. New in this series: the state other cpus touch, ids, vm pointer, armv8 vmpidr, PSCI context and vGIC private state, riscv SBI context, rh850 started flag, lives in a global vcpus_public[] array. The vm's vcpu array points to it, so no cpu ever maps another core's block, and the MPU sibling mapping is gone.
- Placement. Originally every cpu's block went to its affine region unconditionally, with a linker assert. Now it is opt-in per cpu through the hex CPU_TCM_MASK, gated by the platform fact PLAT_HAS_TCM. The generator emits a base table with compact global slots for the cpus left out, puts a placed structure at the end of its region so the guest keeps the base, validates the mask and the description, and rejects architectures without the boot hook. A cpu_home mark on the region picks the TCM when a core has several. The boot code of rh850, tricore and armv8-r reads the table. A static assertion replaces the linker assert.
- Platforms. rh850-u2a16 and tc4dx as before, plus s32z270, whose three TCMs per core are coupled and whose Kconfig choice selects the home TCM, visible only with a nonzero mask.
- Hypervisor MPU map. New: the HYP_MPU_MAP choice with full, simple hulls of flash, RAM, MMIO and the cpu's own region, or none on rh850 where the supervisor bypasses the MPU. MMIO regions were added to rh850-u2a16 and fvp-r for it.
- Vm state split. split vm struct mutable fields into struct vm_mutable accessible through vm->mut, a pointer into a static pool in vm.c bound by the master during vm initialization, with per-arch mutable parts. The fixed fields stay directly in struct vm.
- Local copies. Optionally, copy runtime fixed state into TCM. The option depends on PLAT_HAS_TCM, and the work is two commits, one for the vm description and one for the interrupt bitmap with its global barrier. Making struct cpu embed a vm forced the header split into vm_types.h, cpu_sync.h and vcpu.h, and moved the riscv interrupt line constants into plic.h and aplic.h. For TC4 and RH850 this allows skipping access to global state in the most critical paths.
- Tricore context save areas sit in cpu_arch.csa[]

I still need to further refine and test, but pushing it now for further feedback.

src/core/mem.c
```diff
@@ -212,6 +212,11 @@ static bool pp_root_reserve_cpus(struct page_pool* root_pool)
     size_t cpu_size = platform.cpu_num * mem_cpu_boot_alloc_size();
     paddr_t cpu_base_addr;
 
+    /* With coupled memory the cpu structures are static: in the image or in the coupled regions */
+    if (DEFINED(PLAT_HAS_TCM)) {
+        return true;
+    }
+
```
DavidMCerdeira: This must be refined to consider scenarios where the feature exists but it is not used

src/core/mpu/mem.c
```diff
@@ -196,24 +343,22 @@ static void mem_init_boot_regions(void)
...
 
 void mem_prot_init()
 {
     mpu_init();
     as_init(&cpu()->as, AS_HYP, 0);
+#if CONFIG_HYP_MPU_MAP == HYP_MPU_MAP_SIMPLE
+    mem_init_hyp_hulls();
+    mem_init_cpu_region();
+#elif CONFIG_HYP_MPU_MAP == HYP_MPU_MAP_NONE
+    /* The hypervisor's accesses bypass the MPU: no entry describes its address space */
```
DavidMCerdeira: we can possibly skip other initialization

src/platform/s32z270/Kconfig
```diff
@@ -16,3 +16,31 @@ config GIC_VERSION
...
+	  guests.
+
+config S32Z270_CPU_HOME_ATCM
+	bool "ATCM"
+
+config S32Z270_CPU_HOME_BTCM
+	bool "BTCM"
+
+config S32Z270_CPU_HOME_CTCM
+	bool "CTCM"
```
DavidMCerdeira: this can be extended so that we select TCM per core, but I don't see the need for this right now

